PR #21, opened 2026-04-18
feat(frontend): inline region expansion + BadgeStack cap + map canvas fill

julianken:
## Diagrams

```mermaid
sequenceDiagram
  participant U as User
  participant R as Region.tsx
  participant S as styles.css
  participant B as BadgeStack.tsx

  U->>R: click / Enter on region
  R->>R: computeExpandTransform(svgPath, viewBox)
  Note over R: parsePoints → bbox →<br/>scale = min(vw/w, vh/h) * 0.85<br/>translate centers on viewBox
  R->>R: set transform attr on <g>
  S-->>R: CSS transition: transform 350ms ease
  R->>B: expanded={true}
  B->>B: render ALL badges<br/>(no cap, no pip)
  Note over B: parent g scale makes them readable

  U->>R: click elsewhere / clear region param
  R->>R: transform attr removed
  S-->>R: CSS transitions back to identity (350ms)
  R->>B: expanded={false}
  B->>B: slice(0, 11) + "+N more" pip
```

## Summary

- **Region expansion** (CRITICAL defect #2): clicking a region now computes a `translate(tx ty) scale(s)` transform from the region's own SVG bbox and applies it as an inline `transform` attribute on the expanded `<g>`. CSS `transition: transform 350ms ease` animates the interpolation. On collapse the transform is removed and CSS transitions back to identity. Matches the spec's `InlineExpansion` contract ("grows the selected region to fill the canvas while dimming others") without a new runtime dep.
- **BadgeStack cap** (MAJOR defect #4): collapsed stacks render at most 12 slots — 11 real badges sorted by count-desc plus a neutral `+N more` pip in the 12th slot when there are more. Expanded regions render all (the parent `<g>` scale makes them readable). Eliminates the 80-badge spill that made Sky Islands — Santa Ritas unreadable.
- **Map canvas fill** (MAJOR defect #5, with a post-review fix): root `<svg>` sets `preserveAspectRatio="xMidYMid meet"` + `style={{ width: '100%', height: '100%' }}`. The inline style is required because `.bird-map { width: auto; height: auto }` in `styles.css:31` would override plain HTML `width`/`height` attributes via CSS precedence — bot review caught this (the first fix was a no-op, rendering 739×781 before and after). Empirical measurement on fix: 1408×816 at 1440×900 (97.8% canvas width). `.map-wrap`/`.bird-map` classes are still untouched — PR #16's overflow fix stays intact.
[...]

PR #22, opened 2026-04-18
fix(frontend): transform-origin: 0 0 so region expansion actually renders

julianken:
## Diagrams

```mermaid
graph TB
  subgraph Before[Before this PR — expansion silently broken]
    b1[Click region] --> b2[transform attribute<br/>translate tx ty scale s<br/>set on inner g]
    b2 --> b3[CSS .region has<br/>transform-origin: center]
    b3 --> b4[Chrome applies<br/>pre/post-translate by<br/>bbox center on top of<br/>the baked-in pivot]
    b4 --> b5[Expanded region lands<br/>thousands of px off-screen<br/>nothing visible to user]
  end

  subgraph After[After this PR]
    a1[Click region] --> a2[Same transform attribute<br/>translate tx ty scale s]
    a2 --> a3[CSS .region has<br/>transform-origin: 0 0<br/>SVG default]
    a3 --> a4[Chrome applies transform<br/>exactly as authored]
    a4 --> a5[Expanded region<br/>lands at viewBox center]
  end

  b5 -.fixed.-> a5
```

## Summary

- 1-line CSS fix. Change `.region { transform-origin: center }` → `.region { transform-origin: 0 0 }` in `frontend/src/styles.css:9`.
- **Why the bug existed**: `computeExpandTransform` in `Region.tsx` bakes the pivot into its `translate(tx ty)` component, assuming SVG's default `transform-origin` of `0 0`. The CSS rule from PR #21 overrode that to `center`, which pre/post-translates the transform by the element's bbox center — compounding with the baked-in translate and pushing the expanded region thousands of pixels off the SVG viewport.
- **Empirical evidence**: on `main` (HEAD after #21), `path.getCTM().matrixTransform(pathCenter)` for the expanded Santa Ritas returns `(-1844, -2282)` (in viewport-local coords) instead of the expected `(704, 408)` viewBox center. With this fix: returns `(704, 408)` → region visibly expands.
- The regression was not caught by `Region.test.tsx` or the E2E happy-path because both assert the **presence** of a transform attribute, not its rendered effect. Visual regression testing (Playwright MCP) surfaced it.

## Screenshots

**Before this PR (HEAD of `main` after #21) — clicking Santa Ritas dims all regions but nothing grows; the transformed region lands off-screen:**

Before — broken expansion

**After this PR — Santa Ritas visibly expands to fill the map center (blue stroke border is its `.region-shape` at scale 7.59; other regions dim to opacity 0.2 in the background):**

After — expansion renders

Note: follow-up polish work (not in this PR) — when the region scales 7.59×, the badges inside scale with it too, which makes them oversized. A separate PR should apply a counter-scale to BadgeStack in the expanded state, or switch badge rendering to screen-pixel units (`vector-effect: non-scaling-stroke`). Tracking that as a follow-up defect rather than expanding this PR's scope.

## Test plan

- [x] `npx tsc -b --noEmit` (frontend): clean
- [x] `npm run build --workspace @bird-watch/frontend`: clean (1.00 kB CSS + 154.39 kB JS)
- [x] Empirical browser test against the local stack: `path.getCTM().matrixTransform(pathCenter)` now returns `(704, 408)` (viewport-local client coords) — matching the viewBox center. Was `(-1844, -2282)` before the fix.
- [x] Visual: Santa Ritas visibly grows to fill the map canvas (screenshot attached). The other 8 regions dim to `opacity: 0.2` behind it.
- [x] Does NOT modify `computeExpandTransform` math; does NOT modify `Region.tsx`; does NOT modify `.bird-map` / `.map-wrap`; does NOT add deps.
- [ ] Unit test coverage for the *rendered* expansion result — DEFERRED. jsdom doesn't compute CTMs; would require a Playwright-level test. The happy-path spec already asserts the transform attribute; a rendered-bbox check would be a follow-up test improvement.

## Plan reference

Out of plan — `docs/plans/2026-04-16-plan-*.md`. This is a post-merge defect surfaced by Playwright-MCP visual audit on HEAD after PR #21. Follow-up to `/Users/j/.claude/plans/gather-information-from-the-nested-catmull.md` §"PR B — frontend polish".


julianken:
@Mergifyio queue

Issue #34, opened 2026-04-19
[E2E] Playwright test suite expansion


julianken:
## Overview

Tracking issue for the full Playwright E2E test suite expansion. All child issues were derived from a codebase health audit and reviewed by @julianken-bot before being opened.

**Status:** All 11 child issues landed on `main`. Suite grew from 1 spec (`happy-path.spec.ts`) to 10 specs + POM + shared API stub fixture across 2 Playwright projects.

## Progress

- [x] #23 — Production build smoke test (`vite preview`) — **P0, test.fail until baseUrl fix** — merged as #36
- [x] #24 — Back-button regression (`pushState` not `replaceState`) — **P0, test.fail until pushState fix** — merged as #37
- [x] #25 — Filter flows (time window, family, species blur) — P1 — merged as #38
- [x] #26 — Error screen via `page.route()` interception — P1 — merged as #39
- [x] #27 — Accessibility: keyboard nav, `aria-busy`, `aria-label` — P1 — merged as #40
- [x] #28 — axe-core scan on key app states — P1 — merged as #41
- [x] #29 — Deep-link restore: all URL params survive cold load — P1 — merged as #42
- [x] #30 — Region collapse: background click, re-activate, Escape (test.fail) — P1 — merged as #43
- [x] #31 — Page Object Model (AppPage + FiltersBar, 3 specs migrated) — P2 — merged as #44
- [x] #32 — Shared API stub fixture (migrated error-states + axe) — P2 — merged as #45
- [x] #33 — Worker concurrency + test isolation audit — P2 — merged as #46

## Execution order (as executed)

1. **P0 first** (#23, #24) — both ship with `test.fail()` annotations that will flip to CI failure when the underlying bugs get fixed, forcing annotation removal
2. **P1 in any order** (#25–#30) — all pass on main today; serve as regression guards
3. **P2 last** (#31–#33) — abstractions extracted from working specs, workers/isolation pinned

## Out-of-plan prerequisite

- #35 — `docs(e2e): track pre-existing manual-test.md` — untracked `manual-test.md` was committed to main first so downstream PRs could diff cleanly against a tracked file

## Context

- Source: codebase health audit (`docs/codebase-health-audit/phase-4/analysis-report.md`)
- All issues reviewed and approved by `@julianken-bot` before dispatch
- Manual test reference: `frontend/e2e/manual-test.md`
[...]

PR #39, opened 2026-04-19
test(e2e): error screen via page.route interception (#26)

### LINKED ISSUE #26: [E2E] Error screen renders when API is unreachable (page.route fail)
## Context

`App.tsx:47-53` shows `<div class=\"error-screen\"><h2>Couldn't load map data</h2>...</div>` when any `useBirdData` fetch rejects. `manual-test.md` Flow 10 describes running this manually by killing the read-api. We can exercise the same path deterministically in CI using `page.route()` to intercept and fail requests — no server shutdown needed, no flake window.

This test is also a good place to pin the user-visible error copy so a future refactor that changes the error heading breaks this test before it breaks users.

## Acceptance criteria

- [ ] New spec `frontend/e2e/error-states.spec.ts` uses `page.route('**/api/regions', route => route.abort())` to force a network failure before `page.goto('/')`.
- [ ] Asserts the `.error-screen` is visible and contains the heading text `Couldn't load map data`.
- [ ] Asserts the error paragraph is not empty (contains the caught `Error.message`).
- [ ] Asserts `aria-busy` does NOT stay `true` indefinitely — i.e. the page does not hang in a loading state. (Check that either the error screen rendered OR `aria-busy` went to `false` within a reasonable timeout.)
- [ ] A second test asserts a 500 response from `/api/regions` also triggers the error screen (use `route.fulfill({ status: 500, body: 'boom' })`).
- [ ] A third test asserts that when ONLY `/api/observations` fails but `/api/regions` and `/api/hotspots` succeed, the error screen still renders (this is the current behavior in `use-bird-data.ts:34` — `setError` is the final resolution). If this behavior is deemed undesirable, open a follow-up issue; for now we test-lock the current behavior.

### Spec skeleton

```ts
import { test, expect } from '@playwright/test';

test.describe('error screen', () => {
  test('renders when /api/regions aborts', async ({ page }) => {
    await page.route('**/api/regions', route => route.abort());
    await page.goto('/');
    await expect(page.locator('.error-screen h2'))
      .toHaveText(\"Couldn't load map data\", { timeout: 10_000 });
    await expect(page.locator('.error-screen p')).not.toBeEmpty();
  });

  test('renders on 500 from /api/regions', async ({ page }) => {
    await page.route('**/api/regions', route =>
      route.fulfill({ status: 500, contentType: 'text/plain', body: 'boom' })
    );
    await page.goto('/');
    await expect(page.locator('.error-screen h2'))
      .toHaveText(\"Couldn't load map data\", { timeout: 10_000 });
  });

  test('renders when /api/observations fails even if regions+hotspots succeed', async ({ page }) => {
    await page.route('**/api/observations**', route => route.abort());
    await page.goto('/');
    await expect(page.locator('.error-screen h2'))
      .toHaveText(\"Couldn't load map data\", { timeout: 10_000 });
  });

  test('does not hang with aria-busy=true when API aborts', async ({ page }) => {
    await page.route('**/api/regions', route => route.abort());
    await page.goto('/');
    // Either the error screen appeared, OR aria-busy went false. Both are acceptable;
    // an infinite-loading state would be the bug we guard against.
    await expect(async () => {
      const errorVisible = await page.locator('.error-screen').isVisible();
      const ariaBusy = await page.locator('.map-wrap').getAttribute('aria-busy');
      expect(errorVisible || ariaBusy === 'false' || ariaBusy === null).toBeTruthy();
    }).toPass({ timeout: 10_000 });
  });
});
```

### Route-pattern gotcha

Use `**/api/regions` (not `/api/regions`) so the glob matches both `http://localhost:5173/api/regions` and `http://localhost:8787/api/regions` regardless of whether Vite proxies or the preview server is in front. A too-narrow pattern will not match the proxied request.

### Cleanup

Each `test(...)` runs in a fresh Playwright context, so routes are torn down automatically. No explicit `unroute` needed.

### Why `page.route()` over shutting down the server

- Deterministic: no race between server-down and `page.goto`.
- Isolated: one failing route per test, rest of the suite still has a working API.
- Faster: no server restart overhead.

## Test file location

`frontend/e2e/error-states.spec.ts`

## Labels

e2e, testing

julianken:
## Diagrams

```mermaid
graph TD
    subgraph FAIL_MODES["Route fail modes"]
        R1["regions aborts"] --> E["&lt;.error-screen&gt;"]
        R2["regions 500"] --> E
        R3["observations aborts<br/>regions+hotspots succeed"] --> E
    end

    subgraph NO_HANG["No-hang guard (test 4)"]
        G0["regions aborts"] --> G1{"Promise.race"}
        G1 -->|"leg A"| GA[".error-screen toBeVisible"]
        G1 -->|"leg B"| GB[".map-wrap aria-busy=false"]
    end

    E --> H["h2 pinned: Couldn't load map data"]
    E --> P["p: error.message non-empty"]
```

## Summary

- Adds `frontend/e2e/error-states.spec.ts` with 4 Playwright tests that exercise the `.error-screen` render path via `page.route()` interception — deterministic, no server shutdown, no flake window.
- Pins the user-visible error copy (`Couldn't load map data`) so a future refactor of the heading fails this test before it fails users.
- Covers three trigger paths (abort on regions, 500 on regions, abort on observations-only) plus a no-hang guard that races `toBeVisible` on `.error-screen` with `toHaveAttribute('aria-busy', 'false')` on `.map-wrap`.
- Hardened per internal code review: all route handlers use `async/await` (so Playwright surfaces rejections as test failures), and the no-hang guard uses web-first assertions under `Promise.race` instead of the `isVisible`/`getAttribute` anti-pattern inside `toPass`.

## Screenshots

N/A — not UI. Test-only change.

## Test plan

- [x] `npx tsc -b --noEmit` — clean
- [x] `npm test --workspace @bird-watch/frontend -- --run` — 33/33 passing
- [x] Route pattern uses `**/api/regions` glob (matches both direct and proxied requests); observations test uses `**/api/observations**` trailing-wildcard
- [x] All four route handlers are `async route => { await route.X(); }` — no bare callbacks
- [x] Test 4 races `expect().toBeVisible` and `expect().toHaveAttribute` via `Promise.race` — no `isVisible`/`getAttribute` inside `toPass`
- [x] `App.tsx`, `use-bird-data.ts`, `playwright.config.ts`, `e2e.yml` all untouched — spec locks current error-resolution behavior without changing it
- [x] No `test.fail()` — tests expected to pass on current main
- [ ] Local Playwright E2E deferred — CI is authoritative

## Plan reference

Closes #26. Part of tracking issue #34 (E2E Playwright test suite expansion).
Execution plan: `/Users/j/.claude/plans/i-want-you-to-nifty-petal.md`.

---

### DIFF
--- a/frontend/e2e/error-states.spec.ts
+++ b/frontend/e2e/error-states.spec.ts
@@ -1,0 +1,39 @@
+import { test, expect } from '@playwright/test';
+
+test.describe('error screen', () => {
+  test('renders when /api/regions aborts', async ({ page }) => {
+    await page.route('**/api/regions', route => route.abort());
+    await page.goto('/');
+    await expect(page.locator('.error-screen h2'))
+      .toHaveText("Couldn't load map data", { timeout: 10_000 });
+    await expect(page.locator('.error-screen p')).not.toBeEmpty();
+  });
+
+  test('renders on 500 from /api/regions', async ({ page }) => {
+    await page.route('**/api/regions', route =>
+      route.fulfill({ status: 500, contentType: 'text/plain', body: 'boom' })
+    );
+    await page.goto('/');
+    await expect(page.locator('.error-screen h2'))
+      .toHaveText("Couldn't load map data", { timeout: 10_000 });
+  });
+
+  test('renders when /api/observations fails even if regions+hotspots succeed', async ({ page }) => {
+    await page.route('**/api/observations**', route => route.abort());
+    await page.goto('/');
+    await expect(page.locator('.error-screen h2'))
+      .toHaveText("Couldn't load map data", { timeout: 10_000 });
+  });
+
+  test('does not hang with aria-busy=true when API aborts', async ({ page }) => {
+    await page.route('**/api/regions', route => route.abort());
+    await page.goto('/');
+    // Either the error screen appeared, OR aria-busy went false. Both are acceptable;
+    // an infinite-loading state would be the bug we guard against.
+    await expect(async () => {
+      const errorVisible = await page.locator('.error-screen').isVisible();
+      const ariaBusy = await page.locator('.map-wrap').getAttribute('aria-busy');
+      expect(errorVisible || ariaBusy === 'false' || ariaBusy === null).toBeTruthy();
+    }).toPass({ timeout: 10_000 });
+  });
+});
--- a/frontend/e2e/error-states.spec.ts
+++ b/frontend/e2e/error-states.spec.ts
@@ -2,7 +2,7 @@ import { test, expect } from '@playwright/test';
 
 test.describe('error screen', () => {
   test('renders when /api/regions aborts', async ({ page }) => {
-    await page.route('**/api/regions', route => route.abort());
+    await page.route('**/api/regions', async route => { await route.abort(); });
     await page.goto('/');
     await expect(page.locator('.error-screen h2'))
       .toHaveText("Couldn't load map data", { timeout: 10_000 });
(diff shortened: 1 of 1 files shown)


julianken:
@Mergifyio queue

PR #43, opened 2026-04-19
test(e2e): region collapse via background, re-activate, ESC (#30)

### LINKED ISSUE #30: [E2E] Collapse expanded region — click outside, second click on same region, ESC
## Context

Expanding a region is tested by `happy-path.spec.ts`. Collapsing it is not. There are three current entry points for collapse, and one that SHOULD exist but may not:

1. Clicking the SVG background (`Map.tsx:56-58` — the `onClick` on the `<svg>` calls `onSelectRegion(null)` when `e.target === e.currentTarget`).
2. Clicking the already-expanded region (`Map.tsx:76` toggles based on `isExpanded ? null : r.id`).
3. Pressing Enter/Space on the already-expanded region (`Region.tsx:95-100` uses the same toggle via `onSelect`).
4. **Missing but expected by users:** pressing Escape to collapse. Not currently implemented in `App.tsx` or `Map.tsx` — this test should be written to document the expectation and will FAIL until Escape handling is added. See acceptance criteria for the `test.fail` guidance.

## Acceptance criteria

- [ ] New spec `frontend/e2e/region-collapse.spec.ts` contains:
  1. Expand Santa Ritas (keyboard) → click the SVG background outside the region → `region-expanded` class removed AND `region=` param dropped from URL.
  2. Expand Santa Ritas → press Enter on it again → class + URL param removed.
  3. Expand Santa Ritas → click the shape once more (mouse) → class + URL param removed.
  4. Escape-to-collapse: written as `test.fail(...)` with a comment pointing at the planned fix issue; CI turns red the moment Escape handling ships and the test unexpectedly passes, at which point the agent removes the annotation.
- [ ] All tests start from the same precondition (happy-path setup) — consider a shared `beforeEach` that expands the region.

### Spec skeleton

```ts
import { test, expect } from '@playwright/test';

async function expandSantaRitas(page: import('@playwright/test').Page) {
  await page.goto('/');
  await expect(page.locator('[data-region-id]')).toHaveCount(9, { timeout: 15_000 });
  const region = page.locator('.region-shape[aria-label="Sky Islands — Santa Ritas"]');
  await region.focus();
  await page.keyboard.press('Enter');
  await expect(page.locator('[data-region-id="sky-islands-santa-ritas"]'))
    .toHaveClass(/region-expanded/);
}

test.describe('region collapse', () => {
  test('clicking SVG background collapses', async ({ page }) => {
    await expandSantaRitas(page);
    // Dispatch a synthetic click directly on the <svg> element to bypass
    // Playwright's hit-testing. The AZ viewBox tiles all 9 regions densely
    // enough that no coordinate is guaranteed to land on bare SVG rather than
    // a child <path> — confirmed by inspecting svg_path bounds in
    // migrations/1700000008000_seed_regions.sql:
    //   colorado-plateau covers y=0 from x≈3 to x=360 (so top-left corner is inside a polygon),
    //   sonoran-tucson reaches the bottom-right corner (so {x:355,y:375} is also inside a polygon).
    // page.evaluate dispatches the event on the element itself, so
    // e.target === e.currentTarget is always true, which is exactly the
    // condition Map.tsx:57 checks before calling onSelectRegion(null).
    await page.evaluate(() => {
      const svg = document.querySelector('.bird-map') as SVGSVGElement;
      svg.dispatchEvent(new MouseEvent('click', { bubbles: true, cancelable: true }));
    });
    await expect(page.locator('[data-region-id="sky-islands-santa-ritas"]'))
      .not.toHaveClass(/region-expanded/);
    await expect.poll(() => page.url()).not.toContain('region=');
  });

  test('Enter on already-expanded region collapses', async ({ page }) => {
    await expandSantaRitas(page);
    const region = page.locator('.region-shape[aria-label="Sky Islands — Santa Ritas"]');
    await region.focus();
    await page.keyboard.press('Enter');
    await expect(page.locator('[data-region-id="sky-islands-santa-ritas"]'))
      .not.toHaveClass(/region-expanded/);
    await expect.poll(() => page.url()).not.toContain('region=');
  });

  test('clicking expanded region collapses', async ({ page }) => {
    await expandSantaRitas(page);
    await page.locator('.region-shape[aria-label="Sky Islands — Santa Ritas"]').click();
    await expect(page.locator('[data-region-id="sky-islands-santa-ritas"]'))
      .not.toHaveClass(/region-expanded/);
    await expect.poll(() => page.url()).not.toContain('region=');
  });

  test.fail('Escape collapses expanded region (not yet implemented)', async ({ page }) => {
    // test.fail tells Playwright to run this test and assert it FAILS. If it
    // unexpectedly passes (i.e. Escape handling has landed), CI turns red and
    // the agent removes the annotation to make the test live. This is
    // intentionally different from test.skip / test.fixme, which silently
    // suppress the test forever and provide no signal when the feature ships.
    //
    // The Escape handler should be a document-level keydown listener added in
    // App.tsx via useEffect. It should call the same state update as the
    // SVG-background-click path: set({ regionId: null, speciesCode: null })
    // (matching the onSelectRegion prop at App.tsx:74). This means pressing
    // Escape while a region is expanded must clear BOTH the region and the
    // speciesCode from the URL — assert that neither `region=` nor `species=`
    // appears in the URL after Escape. The handler should be a no-op when no
    // region is currently expanded.
    await expandSantaRitas(page);
    await page.keyboard.press('Escape');
    await expect(page.locator('[data-region-id="sky-islands-santa-ritas"]'))
      .not.toHaveClass(/region-expanded/);
    await expect.poll(() => page.url()).not.toContain('region=');
    await expect.poll(() => page.url()).not.toContain('species=');
  });
});
```

### Gotcha — SVG click targeting

There is no safe bare-SVG coordinate in this viewBox. Checking `migrations/1700000008000_seed_regions.sql`:

- `colorado-plateau` SVG path: `M 3.1 0.0 L 176.9 0.0 L 360.0 0.0 …` — the top edge runs along y=0 from x=3.1 to x=360. **Point (4, 4) is inside this polygon.**
- `sonoran-tucson` SVG path: `M 207.9 253.3 … L 360.0 380.0 L 263.8 380.0 …` — reaches the bottom-right corner. **Point (355, 375) is inside this polygon.**

The `page.evaluate` approach in the skeleton above is the correct solution. It dispatches the click event directly on the `<svg>` element, so the browser never performs hit-testing and `e.target === e.currentTarget` is unconditionally true — which is the exact condition `Map.tsx:57` guards. This approach is more deterministic than any coordinate-based click.

### What the `test.fail` annotation is for

`test.fail` marks a test as "expected to fail." Playwright runs the test body and asserts that it does fail. If the test unexpectedly passes (because Escape handling has been implemented), the suite exits non-zero and CI turns red. That is the signal for the agent to remove `test.fail` and let the test run live. This is the correct annotation for known-broken functionality that should auto-promote once fixed. `test.fixme` silently skips without running the body and never produces any signal — do not use it here.

### Escape handler behavior (implementation note for the follow-up PR)

The Escape test should assert that pressing Escape while a region is expanded calls `onSelectRegion(null)` and clears `speciesCode` (matching App.tsx's `onSelectRegion` handler which sets `{ regionId: id, speciesCode: null }`). Verify both `region` and `species` params are absent from the URL after Escape. The handler should live in `App.tsx` as a document-level `useEffect` keydown listener, fire globally while the app is focused (not only when a region element has focus), and be a no-op when no region is currently expanded.

## Test file location

`frontend/e2e/region-collapse.spec.ts`

## Labels

e2e, testing

julianken:
## Diagrams

```mermaid
graph TD
    subgraph "Collapse entry points"
        E1[expand Santa Ritas<br/>focus + Enter]
        E1 --> C1[dispatchEvent click on .bird-map<br/>Map.tsx guard: e.target === e.currentTarget]
        E1 --> C2[focus + Enter again<br/>Region.tsx toggles via Map onSelect]
        E1 --> C3[dispatchEvent click on .region-shape<br/>bypasses BadgeStack hit-test]
        E1 --> C4[press Escape<br/>test.fail — handler not implemented]
    end

    C1 --> A[region-expanded class removed<br/>+ region= absent from URL]
    C2 --> A
    C3 --> A
    C4 --> B[test.fail inverts:<br/>fails today → passes in CI<br/>flips to red when handler ships]
```

## Summary

- Adds `frontend/e2e/region-collapse.spec.ts` with 4 tests covering three existing collapse entry points plus a `test.fail()`-guarded Escape test for the planned-but-unimplemented document-level Escape handler.
- **Background-click test** uses `page.locator('.bird-map').dispatchEvent('click')` — the AZ viewBox tiles all 9 regions so densely that no bare-SVG coordinate exists for `page.mouse.click`. Playwright's built-in `dispatchEvent` fires directly on the `<svg>` element, guaranteeing `e.target === e.currentTarget` — the exact condition `Map.tsx` guards before calling `onSelectRegion(null)`.
- **Click-expanded-region test** uses the same `dispatchEvent('click')` idiom to be robust to BadgeStack overlays that may cover the region center as observation counts grow.
- **Escape test** uses `test.fail()` (not `test.fixme`) so CI turns red the moment Escape handling ships — the annotation is the signal to remove itself. The test asserts both `region=` AND `species=` are cleared, matching `App.tsx`'s `onSelectRegion(null)` handler semantics.
- Shared `expandSantaRitas(page)` helper at file top — saves ~21 lines of duplication across the 4 tests.

## Screenshots

N/A — not UI. Test-only change.

## Test plan

- [x] `npx tsc -b --noEmit` — clean
- [x] `npm test --workspace @bird-watch/frontend -- --run` — 33/33 passing
- [x] Preflight: `Map.tsx` has `className="bird-map"`; SVG `onClick` guards `e.target === e.currentTarget`; Region.tsx fires `onSelect` on Enter/Space; `App.tsx` `onSelectRegion` clears both `regionId` and `speciesCode`
- [x] `expect.poll` on URL assertions uses explicit `{ timeout: 5_000 }` (matches house style)
- [x] No `test.fail()` on tests 1-3 (they should pass on current main); test 4 has `test.fail()` as the first statement with an explanatory comment
- [x] No source changes, no `playwright.config.ts` changes
- [ ] Local Playwright E2E deferred — CI is authoritative

## Plan reference

Closes #30. Part of tracking issue #34 (E2E Playwright test suite expansion).
Execution plan: `/Users/j/.claude/plans/i-want-you-to-nifty-petal.md`.

---

### DIFF
--- a/frontend/e2e/region-collapse.spec.ts
+++ b/frontend/e2e/region-collapse.spec.ts
@@ -1,0 +1,64 @@
+import { test, expect, type Page } from '@playwright/test';
+
+async function expandSantaRitas(page: Page) {
+  await page.goto('/');
+  await expect(page.locator('[data-region-id]')).toHaveCount(9, { timeout: 15_000 });
+  const region = page.locator('.region-shape[aria-label="Sky Islands — Santa Ritas"]');
+  await region.focus();
+  await page.keyboard.press('Enter');
+  await expect(page.locator('[data-region-id="sky-islands-santa-ritas"]'))
+    .toHaveClass(/region-expanded/);
+}
+
+test.describe('region collapse', () => {
+  test('clicking SVG background collapses', async ({ page }) => {
+    await expandSantaRitas(page);
+    // Dispatch a synthetic click directly on the <svg> element. The AZ viewBox
+    // tiles all 9 regions densely enough that no coordinate is guaranteed to
+    // land on bare SVG, so page.evaluate bypasses hit-testing and guarantees
+    // e.target === e.currentTarget (the exact condition Map.tsx guards).
+    await page.evaluate(() => {
+      const svg = document.querySelector('.bird-map') as SVGSVGElement;
+      svg.dispatchEvent(new MouseEvent('click', { bubbles: true, cancelable: true }));
+    });
+    await expect(page.locator('[data-region-id="sky-islands-santa-ritas"]'))
+      .not.toHaveClass(/region-expanded/);
+    await expect.poll(() => page.url(), { timeout: 5_000 }).not.toContain('region=');
+  });
+
+  test('Enter on already-expanded region collapses', async ({ page }) => {
+    await expandSantaRitas(page);
+    const region = page.locator('.region-shape[aria-label="Sky Islands — Santa Ritas"]');
+    await region.focus();
+    await page.keyboard.press('Enter');
+    await expect(page.locator('[data-region-id="sky-islands-santa-ritas"]'))
+      .not.toHaveClass(/region-expanded/);
+    await expect.poll(() => page.url(), { timeout: 5_000 }).not.toContain('region=');
+  });
+
+  test('clicking expanded region collapses', async ({ page }) => {
+    await expandSantaRitas(page);
+    await page.locator('.region-shape[aria-label="Sky Islands — Santa Ritas"]').click();
+    await expect(page.locator('[data-region-id="sky-islands-santa-ritas"]'))
+      .not.toHaveClass(/region-expanded/);
+    await expect.poll(() => page.url(), { timeout: 5_000 }).not.toContain('region=');
+  });
+
+  test('Escape collapses expanded region (not yet implemented)', async ({ page }) => {
+    // test.fail asserts this test MUST fail. When Escape handling ships and
+    // this test unexpectedly passes, CI turns red and the annotation must be
+    // removed. test.fixme silently skips forever with no signal — do NOT use
+    // it here.
+    //
+    // The planned handler is a document-level keydown listener in App.tsx
+    // that calls onSelectRegion(null), clearing BOTH region and species from
+    // the URL (matching App.tsx's onSelectRegion prop).
+    test.fail();
+    await expandSantaRitas(page);
+    await page.keyboard.press('Escape');
+    await expect(page.locator('[data-region-id="sky-islands-santa-ritas"]'))
+      .not.toHaveClass(/region-expanded/);
+    await expect.poll(() => page.url(), { timeout: 5_000 }).not.toContain('region=');
+    await expect.poll(() => page.url(), { timeout: 5_000 }).not.toContain('species=');
+  });
+});
--- a/frontend/e2e/region-collapse.spec.ts
+++ b/frontend/e2e/region-collapse.spec.ts
@@ -15,12 +15,9 @@ test.describe('region collapse', () => {
     await expandSantaRitas(page);
     // Dispatch a synthetic click directly on the <svg> element. The AZ viewBox
     // tiles all 9 regions densely enough that no coordinate is guaranteed to
-    // land on bare SVG, so page.evaluate bypasses hit-testing and guarantees
+    // land on bare SVG, so dispatchEvent bypasses hit-testing and guarantees
     // e.target === e.currentTarget (the exact condition Map.tsx guards).
-    await page.evaluate(() => {
-      const svg = document.querySelector('.bird-map') as SVGSVGElement;
-      svg.dispatchEvent(new MouseEvent('click', { bubbles: true, cancelable: true }));
-    });
+    await page.locator('.bird-map').dispatchEvent('click');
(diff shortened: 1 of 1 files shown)


julianken:
@Mergifyio queue

PR #546, opened 2026-05-15
phase 2: adaptive-grid atomic cutover — wire + delete auto-spider + delete mosaic (#542)

julianken:
## Diagrams

```mermaid
flowchart TB
  subgraph "Before (deleted in this PR)"
    A1[MosaicMarker.tsx · 2×2 fixed]
    A2[ClusterPill.tsx · numeric pill]
    A3[StackedSilhouetteMarker.tsx + use-auto-spider + stack-fanout + fan-layout]
    A4[auto-spider-leader-lines-layer]
  end
  subgraph "After (one component, all zooms)"
    B1[AdaptiveGridMarker.tsx]
    B1 -- "≤16 fams AND ≤64 obs" --> B2[1×1 → 4×4 adaptive grid]
    B1 -- ">16 fams OR >64 obs" --> B3[ClusterPill · still used]
    B4["clusterMaxZoom 14 → 22<br/>(source maxzoom 24)"]
  end
  subgraph "Memoization · 3 concerns · spec §5.3"
    M1[Concern A · useMemo identity stability]
    M2[Concern B · module-scoped async cache · zoom-prefixed key]
    M3[Concern C · cacheGeneration + silhouettesVersion]
  end
  A1 -.-> B1
  A2 -.-> B3
  A3 -.->|delete| X[ ]
  A4 -.->|delete| X
  B1 --- M1
```

## Summary

- Phase 2 of epic #539 — atomic cutover. Wires `<AdaptiveGridMarker>` into `MapCanvas.tsx`, raises `clusterMaxZoom` 14→22 (with `<Source>` `maxzoom={24}` per Phase 0 finding F4), implements the three-concern memoization (Concerns A/B/C from spec §5.3), deletes 14 legacy files (9 auto-spider + 5 mosaic), flips `VITE_FF_ADAPTIVE_GRID` default to `true`.
- **Cluster-level `isNotable` propagation (F7)**: chose **option (a)** — strict 1×1 only. Multi-family grids never propagate the amber ring at marker level.
- **Mergify gating for perf-gate**: **Option A (soft-launch)** — `continue-on-error: true` on the new `.github/workflows/perf-gate.yml`. Mirrors `orphan-classname-check` soft-launch pattern.
- 4 commits. Full suite 734 tests pass (including 4 new memoization tests for Concerns A/B/C). Knip clean. Orphan-classname-check PASS.

## Screenshots

[...]

PR #553, opened 2026-05-15
fix(map): AdaptiveGridMarker badge hidden when cell.count===1 (#552)

### LINKED ISSUE #552: bug(map): AdaptiveGridMarker shows badge for count=1 cells in multi-cell clusters (spec §4.3)
## Bug

In multi-cell `<AdaptiveGridMarker>` clusters, cells whose individual `count === 1` still render a "1" badge. Per spec §4.3 the badge should be hidden when `cell.count === 1`, regardless of the cluster's total observation count.

## Repro

Visit https://bird-maps.com at z≥12. Find any cluster of mixed counts — most grids in production hit this case. Screenshot example below shows 7 "1" badges on cells that should be silhouette-only.

bug example

## Expected

Cells with `count === 1` render the silhouette only, identical visual to a single-observation marker. Spec §4.3 line 124:

> **Hidden when `cell.count === 1`.** A 1×1 grid with count=1 looks visually identical to today's individual observation marker. Default to silence; only show count when count > 1.

## Root cause

`frontend/src/components/map/AdaptiveGridMarker.tsx:128`:

```ts
const showBadgeFor = (cellCount: number): boolean => totalCount > 1 || cellCount > 1;
```

The `totalCount > 1 ||` clause was over-thought — the comment at lines 125-127 explicitly states the buggy intent ("hidden ONLY when both the cluster total is 1 AND the cell count is 1"). Spec actually says hide when `cell.count === 1`, period.

## Fix

```ts
// frontend/src/components/map/AdaptiveGridMarker.tsx:128
- const showBadgeFor = (cellCount: number): boolean => totalCount > 1 || cellCount > 1;
+ const showBadgeFor = (cellCount: number): boolean => cellCount > 1;
```

Update the comment at lines 125-127 to reflect spec §4.3 wording.

## Test gap that let this slip

`frontend/src/components/map/AdaptiveGridMarker.test.tsx` covers:
- "renders 1×1 with NO badge when totalCount === 1" ✓
- "renders 1×1 with badge 5 when totalCount === 5" ✓
- "renders 2×2 with 4 per-cell badges in descending count order" — fixture used all counts > 1, missing the count=1-in-multi-cell-cluster case

Add a regression test:

```ts
it('hides badge for cells with count === 1 even in multi-cell clusters', () => {
  // 2×2 with mixed counts: one cell at count=1 → that cell renders silhouette only
  render(
    <AdaptiveGridMarker
      shape={{ tag: 'grid', cols: 2, rows: 2 }}
      tiles={[rendered('a', 5), rendered('b', 3), rendered('c', 2), rendered('d', 1)]}
      totalCount={11}
      uniqueFamilies={4}
      ariaLabel="…"
      onClick={() => {}}
    />,
  );
  const badges = screen.queryAllByTestId('adaptive-grid-marker-badge');
  expect(badges).toHaveLength(3); // not 4 — the count=1 cell has no badge
  expect(badges.map(b => b.textContent)).toEqual(['5', '3', '2']);
});
```

## Severity

Visual-noise bug. Spec-non-compliance. Not blocking — markers still function — but the production map looks busier than designed.

## Related

- Spec: `docs/specs/2026-05-14-adaptive-cluster-grid-design.md §4.3`
- Component: `frontend/src/components/map/AdaptiveGridMarker.tsx:128`
- Tests: `frontend/src/components/map/AdaptiveGridMarker.test.tsx`
- Epic that introduced: #539
- Closing PR for the epic: #547

## Refinements from julianken-bot review (round 1)

Two SUGGESTIONs from the bot review at issue-comment 4460175616:

1. **Strengthen the regression fixture** — current proposal asserts 3 badges with the right text in a `[5,3,2,1]` 2×2. Stronger version: use ties at count=1 (`[3,1,1,1]`) and explicitly assert the 4th-cell-by-DOM-order has NO badge. This pins "count=1 → silhouette-only" rather than just "3 badges with correct text". Updated regression test:

   ```ts
   it('hides badge for cells with count === 1 even in multi-cell clusters', () => {
     render(
       <AdaptiveGridMarker
         shape={{ tag: 'grid', cols: 2, rows: 2 }}
         tiles={[rendered('a', 3), rendered('b', 1), rendered('c', 1), rendered('d', 1)]}
         totalCount={6}
         uniqueFamilies={4}
         ariaLabel="…"
         onClick={() => {}}
       />,
     );
     const badges = screen.queryAllByTestId('adaptive-grid-marker-badge');
     expect(badges).toHaveLength(1); // only the count=3 cell gets a badge
     expect(badges[0].textContent).toBe('3');
   });
   ```

   Augment, don't replace, the existing "renders 2×2 with 4 per-cell badges in descending count order" test — it pins a distinct invariant (descending-order rendering with all counts >1).

2. **Boundary-check `aggregateClusterFamilies` does not silently filter count=1 families** before they reach the marker. The function in `frontend/src/components/map/adaptive-grid.ts` should be inspected: if it drops families whose count is 1, the surface symptom would look similar to this bug but the fix would be different (the silhouette wouldn't render at all). Verify via test or grep — likely a no-op, but worth confirming during the fix.

julianken:
## Diagrams

```mermaid
flowchart TB
  A["showBadgeFor cellCount"] -- "before: totalCount>1 || cellCount>1" --> B["badge on every cell in multi-cell cluster"]
  A -- "after: cellCount > 1" --> C["badge hidden when cell.count === 1"]
  D["spec §4.3 line 124"] -- "Hidden when cell.count === 1" --> C
  E["regression test: 2x2 [3,1,1,1]"] -- "asserts exactly 1 badge text=3" --> C
```

## Summary

- Fixes #552: `AdaptiveGridMarker` rendered a "1" badge on every count=1 cell in multi-cell clusters, contradicting spec §4.3 line 124 ("Hidden when `cell.count === 1`").
- Root cause: `showBadgeFor` at `AdaptiveGridMarker.tsx:128` had `totalCount > 1 || cellCount > 1` — the OR clause short-circuited the per-cell rule. Fix: drop `totalCount > 1 ||`. In the 1×1 single-observation case the outcome is unchanged (cellCount === totalCount === 1 → no badge).
- New regression test uses the strengthened `[3,1,1,1]` 2×2 fixture from issue #552's round-1 bot-review refinement: asserts exactly 1 badge with text "3" rather than 4 badges with mixed text.

## Screenshots

Before (production, shipped via PR #546) — every count=1 cell carries a redundant "1" badge per the buggy OR clause:

bug example

After: the new regression test pins the corrected behavior. Marker visual: silhouette-only for count=1 cells, badge only for count>1.

N/A for the canonical 5-viewport × 2-theme capture — this is a one-line behavior fix to existing component; visual diff is the badges disappearing on count=1 cells.

## Test plan

- [x] New regression test in `AdaptiveGridMarker.test.tsx`: `"hides badge for cells with count === 1 even in multi-cell clusters"` — fixture `[3,1,1,1]`, asserts `badges.length === 1` and `badges[0].textContent === '3'`
- [x] Existing 15 tests in `AdaptiveGridMarker.test.tsx` still pass (16/16 total green)
- [x] `aggregateClusterFamilies` boundary check: confirmed via `grep -n "count" adaptive-grid.ts` — no count filter applied to families. Bug is purely in the marker layer.
- [x] `npm run build` green (tsc -b + vite build clean)

## Risk

Minimal. One-line change to a pure rendering predicate. No memoization, no state, no API surface change.

Small scope-add: dropped the now-unused `totalCount` from the destructured props (kept on the public interface for aria-label parity). Doc-comment updated to record that `totalCount` is no longer load-bearing for badge visibility.

Closes #552

### DIFF
--- a/frontend/src/components/map/AdaptiveGridMarker.test.tsx
+++ b/frontend/src/components/map/AdaptiveGridMarker.test.tsx
@@ -71,6 +71,27 @@ describe('AdaptiveGridMarker', () => {
 
   // --- 2×2 grid: 4 per-cell badges in descending count order --------------
 
+  it('hides badge for cells with count === 1 even in multi-cell clusters', () => {
+    // Regression for issue #552: per spec §4.3 line 124 the badge is hidden
+    // when `cell.count === 1`, regardless of the cluster's total. The prior
+    // `totalCount > 1 || cellCount > 1` guard caused every count=1 cell in
+    // a multi-cell cluster to render a "1" badge. Fixture uses ties at 1 so
+    // we pin "count=1 → silhouette-only" rather than just descending order.
+    render(
+      <AdaptiveGridMarker
+        shape={SHAPE_2x2}
+        tiles={[rendered('a', 3), rendered('b', 1), rendered('c', 1), rendered('d', 1)]}
+        totalCount={6}
+        uniqueFamilies={4}
+        ariaLabel="Cluster: 6 observations, 4 families. Activate to zoom in."
+        onClick={noop}
+      />,
+    );
+    const badges = screen.queryAllByTestId('adaptive-grid-marker-badge');
+    expect(badges).toHaveLength(1); // only the count=3 cell gets a badge
+    expect(badges[0].textContent).toBe('3');
+  });
+
   it('renders 2×2 with 4 per-cell badges in descending count order', () => {
     render(
       <AdaptiveGridMarker
--- a/frontend/src/components/map/AdaptiveGridMarker.tsx
+++ b/frontend/src/components/map/AdaptiveGridMarker.tsx
@@ -54,7 +54,12 @@ export interface AdaptiveGridMarkerProps {
    * empty trailing slots render as transparent padders.
    */
   tiles: ReadonlyArray<AdaptiveTile>;
-  /** Cluster's full `point_count` (drives badge visibility for the 1×1 single-obs case). */
+  /**
+   * Cluster's full `point_count`. Not used for badge visibility (per spec
+   * §4.3 that's a per-cell decision driven by `cellCount > 1`) but kept on
+   * the public prop shape because callers thread it for aria-label
+   * construction and parity with sibling marker components.
+   */
   totalCount: number;
   /** For aria-label parity; not currently rendered (parent threads label). */
   uniqueFamilies: number;
@@ -93,7 +98,6 @@ export function AdaptiveGridMarker(props: AdaptiveGridMarkerProps) {
   const {
     shape,
     tiles,
-    totalCount,
     ariaLabel,
     describedByListId,
     describedByItems,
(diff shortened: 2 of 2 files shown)


julianken:
@Mergifyio queue

PR #620, opened 2026-05-18
infra(monitoring): budget alerts + CF Pages request tripwire (Phase 0 finishers)

julianken:
## Diagrams

```mermaid
flowchart LR
    GCP[GCP billing<br/>bird-maps-prod]
    Budget[google_billing_budget<br/>$100/mo]
    NC[email_julian<br/>notification channel]
    CF[Cloudflare Pages<br/>bird-watch-frontend]
    GHA[GH Actions cron<br/>cf-pages-request-tripwire]
    Email["julian.kennon.d@gmail.com"]

    GCP -->|spend| Budget
    Budget -->|0.5 warn / 1.0 crit| NC --> Email
    CF -->|GraphQL Analytics| GHA
    GHA -->|::warning:: 80k<br/>fail @ 95k| Email
```

## Summary

- Closes Phase 0 exit gates **P0.k** and **P0.l** on the amended going-national umbrella plan §7 (PR #618).
- **P0.k** — `google_billing_budget` ($100/mo, 0.5 + 1.0 thresholds) on `bird-maps-prod`, routed to the existing `email_julian` monitoring channel; default IAM recipients disabled to avoid billing-admin spam.
- **P0.l** — Daily GH Actions cron (`cf-pages-request-tripwire.yml`) queries the Cloudflare GraphQL Analytics API for the `bird-watch-frontend` Pages project; warns at 80k req/day, fails the job at 95k req/day. Approach B chosen because the v4 `cloudflare_notification_policy` `alert_type` enum has no per-Pages-project request-count threshold (`billing_usage_alert` is account-level; `pages_event_alert` covers deployments, not traffic).

## Screenshots

N/A — not UI

## Test plan

- [x] `terraform fmt -check` in `infra/terraform` — clean
- [x] `terraform validate` in `infra/terraform` — `Success! The configuration is valid.`
- [ ] Operator: before next `terraform apply`, add `gcp_billing_account_id` to `terraform.tfvars` (find via `gcloud beta billing accounts list`; confirm linkage via `gcloud beta billing projects describe bird-maps-prod`).
- [ ] Operator: after first `terraform apply`, verify the budget appears under GCP Console → Billing → Budgets & alerts.
- [ ] Post-merge: trigger `cf-pages-request-tripwire` once via `workflow_dispatch` to confirm the GraphQL query path works end-to-end against the live Pages project.

## Plan reference

[...]

PR #621, opened 2026-05-18
fix(infra): set google provider billing_project for Budgets API

julianken:
## Diagrams

N/A — two-line provider config change; no flow to diagram.

## Summary

- PR #620 added a `google_billing_budget` resource which fails to apply with `Error 403: ... billingbudgets.googleapis.com API requires a quota project, which is not set by default.`
- The Billing Budgets API requires a quota project on the ADC; `gcloud auth application-default set-quota-project` alone isn't enough — the terraform google provider must also be told to use the user's project for quota/billing via `billing_project` + `user_project_override = true`. See https://cloud.google.com/docs/authentication/adc-troubleshooting/user-creds
- No behavior change for existing resources — they already authenticate fine with implicit credentials. Only API-key-quota'd APIs (Billing Budgets) need the override. With this fix in place, PR #620's `budget.tf` applies cleanly.

## Screenshots

N/A — not UI

## Test plan

- [x] `terraform fmt -check` — clean
- [x] `terraform validate` (after `terraform init -backend=false`) — Success! The configuration is valid.
- [ ] Operator runs `terraform apply` post-merge to verify the budget resource applies without the 403.

## Plan reference

Out of plan — hotfix unblocking PR #620's budget alert resource.

---

### DIFF
--- a/infra/terraform/main.tf
+++ b/infra/terraform/main.tf
@@ -1,6 +1,11 @@
 provider "google" {
   project = var.gcp_project_id
   region  = var.gcp_region
+
+  # Required for google_billing_budget and other API-key-quota'd resources.
+  # See https://cloud.google.com/docs/authentication/adc-troubleshooting/user-creds
+  billing_project       = var.gcp_project_id
+  user_project_override = true
 }
 
 provider "neon" {


julianken:
@Mergifyio queue

PR #615, opened 2026-05-17
infra: Cloud SQL Stage 2 — Auth Proxy socket mount (cutover-safe)

julianken:
## Diagrams

```mermaid
graph LR
  subgraph CloudRun["Cloud Run services + jobs"]
    RA[read-api]
    AA[admin-api]
    IG[ingestor]
    IGP[ingestor-photos]
    IGD[ingestor-descriptions]
    IGPR[ingestor-prune]
  end

  RA -- DATABASE_URL --> NEON[(Neon — live traffic)]
  AA -- DATABASE_URL --> NEON
  IG -- DATABASE_URL --> NEON
  IGP -- DATABASE_URL --> NEON
  IGD -- DATABASE_URL --> NEON
  IGPR -- DATABASE_URL --> NEON

  RA -. "/cloudsql socket (idle)" .-> CSQL[(Cloud SQL — birdwatch-pg16)]
  AA -. "/cloudsql socket (idle)" .-> CSQL
  IG -. "/cloudsql socket (idle)" .-> CSQL
  IGP -. "/cloudsql socket (idle)" .-> CSQL
  IGD -. "/cloudsql socket (idle)" .-> CSQL
  IGPR -. "/cloudsql socket (idle)" .-> CSQL
```

Solid lines = live traffic (Neon, unchanged). Dotted lines = Auth Proxy socket mounted but idle until Stage 3 flips `DATABASE_URL` via a Secret Manager version bump.

## Summary

- Stage 2 of the Neon→Cloud SQL migration: adds `cloud_sql_instance` volume + `/cloudsql` `volume_mounts` to all 2 Cloud Run services and all 4 Cloud Run Jobs. Cutover-safe — `DATABASE_URL` is unchanged.
- Mounting the Auth Proxy ahead of time means Stage 3 cutover is a single Secret Manager version bump, not a coordinated 6-resource Terraform apply.
- Depends on Stage 1 (`cloud-sql.tf`) being applied in prod so `google_sql_database_instance.birdwatch.connection_name` resolves at apply time.

## Screenshots

N/A — not UI

## Test plan

- [x] `terraform fmt -check` — clean
- [x] `terraform validate` — Success
- [ ] Operator runs `terraform plan` — expects 6 in-place updates (2 services + 4 jobs), no replacements, no destroys
- [ ] Operator runs `terraform apply` — Cloud Run revisions roll forward; existing connections unaffected since DATABASE_URL is unchanged
[...]

PR #633, opened 2026-05-18
fix(infra): grant roles/cloudsql.client to Cloud Run SAs

julianken:
## Summary
- T2 (#615) mounted the Cloud SQL Auth Proxy volume on admin-api, ingestor, read-api but missed `roles/cloudsql.client` on the SAs. Surfaced 2026-05-18 08:00 UTC when ingestor dual-write went live — every secondary write logged `dual_write_secondary_failed kind=recent err=connect ENOENT /cloudsql/.../.s.PGSQL.5432` (24+ failures in the 08:00 tick alone).
- Adds three `google_project_iam_member` resources binding `roles/cloudsql.client` to `bird-admin-api`, `bird-ingestor`, `bird-read-api` SAs.

## Why
Without this grant the Auth Proxy sidecar can't authenticate to the Cloud SQL instance, so the unix socket never appears and the app sees `ENOENT`. The dual-write feature flag correctly logs+continues, but Cloud SQL stays empty — defeating the purpose of pre-T4 sync.

## Risk
Pure additive IAM grant. No write-path changes, no schema changes. Granting `cloudsql.client` is the standard Cloud Run + Cloud SQL pattern and is least-privilege (only allows connecting through the Auth Proxy, not direct API access).

## Test plan
- [ ] `terraform validate` clean (verified locally)
- [ ] CI: build, lint, test, e2e (no app code touched — should all pass)
- [ ] After apply: re-run a `/recent` tick and verify `ingest_runs.max(started_at)` advances on Cloud SQL within 5 min
- [ ] Verify no `dual_write_secondary_failed` log entries in the next ingestor execution

## Rollback
`terraform state rm google_project_iam_member.cloudsql_client_*` + `gcloud projects remove-iam-policy-binding bird-maps-prod --member=serviceAccount:<sa> --role=roles/cloudsql.client` for each. Or revert this PR. No data state changes to unwind.

## Screenshots
N/A — backend-only IAM grants, no UI surfaces touched.

### DIFF
--- a/infra/terraform/cloud-sql.tf
+++ b/infra/terraform/cloud-sql.tf
@@ -174,3 +174,27 @@ resource "google_secret_manager_secret_version" "cloudsql_db_url" {
   secret      = google_secret_manager_secret.cloudsql_db_url.id
   secret_data = local.cloudsql_pooled_url
 }
+
+# Grant Cloud SQL client role to the service accounts whose Cloud Run workloads
+# mount the Cloud SQL Auth Proxy sidecar. Without this, the proxy fails with
+# `boss::NOT_AUTHORIZED ... missing permission cloudsql.instances.get` and the
+# downstream connection sees `connect ENOENT /cloudsql/<conn>/.s.PGSQL.5432`.
+# Surfaced 2026-05-18 when ingestor dual-write went live: T2 (#615) mounted the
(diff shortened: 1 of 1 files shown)


julianken:
@Mergifyio queue

PR #682, opened 2026-05-20
chore(analytics): migrate analytics.ts from PostHog to Clarity + view tagging (#657-followup)

julianken:
## Diagrams

Before (PR #659 ships Clarity init, but the wrapper still routed through dead-wired PostHog):

```mermaid
flowchart LR
    A[url-state.ts<br/>url_corruption_recovered_511]
    B[SpeciesDetailSurface.tsx<br/>panel_opened / dwell / scroll]
    W[analytics.capture]
    P[posthog-js<br/>VITE_POSTHOG_KEY env var<br/>NEVER injected by deploy]
    DROP[(silent no-op in prod)]
    C[clarity.ts<br/>Clarity.init - PR #659]
    CS[Clarity backend]

    A --> W
    B --> W
    W --> P
    P --> DROP
    C --> CS
```

After (single live transport; setView adds the view dimension):

```mermaid
flowchart LR
    A[url-state.ts<br/>url_corruption_recovered_511]
    B[SpeciesDetailSurface.tsx<br/>panel_opened / dwell / scroll]
    APP[App.tsx<br/>useEffect on state.view]
    W[analytics.capture - unchanged API<br/>analytics.setView - new]
    CSDK["@microsoft/clarity<br/>Clarity.event(name)<br/>Clarity.setTag(key, value)"]
    C[clarity.ts<br/>Clarity.init - PR #659]
    CS[Clarity backend<br/>filterable by view tag]

    A --> W
    B --> W
    APP --> W
    W --> CSDK
    C --> CSDK
    CSDK --> CS
```

## Summary

- **PostHog dropped.** The `posthog-js` dep is gone; `VITE_POSTHOG_KEY` was never injected by `.github/workflows/deploy-frontend.yml`, so every `analytics.capture` call had been a silent no-op in prod since #357 merged. The four existing call sites (`url-state.ts` + `SpeciesDetailSurface.tsx`) are unchanged — only the underlying transport swapped.
- **Clarity now backs `analytics.capture()`.** A single `capture(name, props)` fans out to `Clarity.event(name)` + one `Clarity.setTag(key, String(value))` per prop so the event-with-dimensions intent survives Clarity's split API. A `clarityReady()` guard makes the wrapper safe in dev/test where `Clarity.init` is skipped.
- **`analytics.setView(view)` wired in App.tsx** via `useEffect([state.view])` so Clarity dashboards can filter sessions by surface (`feed | map | species | detail`) without us re-emitting an event per navigation.
[...]

PR #687, opened 2026-05-20
chore(cleanup): drop CF-WAF-blocked post-deploy check + sweep stale PostHog narrative

### LINKED ISSUE #681: ci(deploy): post-deploy bundle-hash check 403s on every deploy (CF WAF blocks Actions runner)
## Background

The `Post-deploy bundle-hash check` step in `.github/workflows/deploy-frontend.yml` has now red-X'd **at least 6 consecutive deploys**. The wrangler-action upload step before it succeeds every time; only the verification `curl` from the GitHub Actions runner fails.

```
curl: (22) The requested URL returned error: 403   (×6 retries)
##[error]Process completed with exit code 22.
```

The check fetches `https://bird-maps.com/`, extracts the first `/assets/*.js` from the HTML, fetches that bundle, and greps for `api.bird-maps.com` to confirm the build env propagated. The intent is a fail-forward guard: if the bundle didn't bake in the API URL, the previous CF Pages deploy stays live.

## Evidence

- ✅ Live site is healthy: HTTP 200 from any non-Actions IP; bundle contains `api.bird-maps.com`.
- ✅ wrangler-action upload always succeeds (visible in the same job logs).
- ❌ Only the curl from the runner gets 403'd — by Cloudflare, before the bundle is reached.

The 403 is consistent across all 6 deploys, all 5 retries per deploy. It's deterministic, not flaky.

## Hypothesis

CF rate-limit / WAF deployed in Phase 0 infra is blocking the GitHub Actions runner egress. Two plausible rule paths:
- **UA-based bot detection** — runners send `curl/X.Y.Z` UA; if a CF WAF rule challenges suspicious UAs, this trips it on every deploy.
- **IP-based reputation** — GitHub Actions egress IPs are shared with many other CI users; CF reputation may be poor and trigger a challenge.

It is probably *not* the rate-limit rule documented in [[feedback_audit_pacing_cf_rate_limit]] (60/min/IP) — the post-check makes ~6 requests per deploy, well under the threshold. But the adjacency is worth noting; same WAF, different rule.

## Impact

- **Noisy red ✗ on every merge to main** — hides any future *real* deploy failure inside this same check.
- **Mergify** sees `deploy-frontend` as failed in the run history, even though it's not currently a queue-required check, this could change if the workflow is ever moved into the required set.
- **No fix path is obvious** without CF dashboard access — and historically inspecting CF WAF rules has needed Julian's hands on the keyboard.

## Options (pick one, ranked by effort)

1. **Drop the check entirely.** wrangler-action's exit code + the `Deploy to Cloudflare Pages` step's success is sufficient for "the upload happened." The bundle-hash check was a belt-and-suspenders guard that has stopped earning its keep. ~5 lines deleted from the workflow.
2. **Replace curl with the CF API.** `GET /accounts/{id}/pages/projects/birdwatch/deployments` returns the latest deployment status and target bundle URL. Authenticated calls bypass the WAF rule. ~10 lines.
3. **Allow-list GitHub Actions IP ranges** in CF (admin action on `julianken/bird-sight-system`'s CF Pages WAF rule). Specific to bird-maps.com's WAF setup.
4. **Set a custom User-Agent** on the curl that bypasses the rule. Brittle — only works if the rule is UA-based, not IP-based.

My read: **Option 1.** The check is mostly redundant given wrangler-action's exit-code semantics, and the cases where wrangler succeeds-with-broken-bundle are vanishingly rare. If we want belt-and-suspenders later, do Option 2 (CF API) which is also more reliable than the curl-and-grep pattern.

## Related

- The new `feedback_audit_pacing_cf_rate_limit.md` memory documents the 60/min/IP rate limit found when bursting `/api/*` from a dev IP. The deploy-check 403 is probably a *different* WAF rule, but same WAF.
- The `lockfile-consistency` and `terraform-plan-drift-check` workflows are also "outside the Mergify gate" — same class of "informational" check, but those at least pass. This is the only one in that class that consistently fails.

## Acceptance

Picking and shipping one of the options above. If Option 1, also remove the now-dead `set -euo pipefail` script block. If Option 2 or 3, document the new behavior so future ops folks don't undo the rule.

julianken:
## Diagrams

```mermaid
flowchart LR
    A[deploy-frontend workflow] --> B[wrangler-action]
    B --> C[CF Pages deploy]
    C -.removed.-> D[Post-deploy bundle-hash check<br/>403 from CF WAF]
    style D stroke-dasharray: 5 5,stroke:#888
```

Workflow change drops the dead step; comment edits don't lend themselves to a diagram (no flow change). The flowchart above shows the deploy path after this PR — the dashed node is the removed step.

## Summary

- Closes #681 by dropping the post-deploy bundle-hash check that has 403'd on 6+ consecutive deploys because Cloudflare's WAF blocks the GitHub Actions runner IP. Its intended failure mode (wrangler succeeds + Vite silently strips an env var) is vanishingly rare; the check is dead weight hiding real deploy failures behind a recurring red ✗.
- Sweeps four stale present-tense PostHog comments flagged by @julianken-bot R7 SUGGESTION on #682 (`clarity.ts`, `state/url-state.ts`, `SpeciesDetailSurface.test.tsx`, `AttributionModal.test.tsx`). Migration-history comments in `analytics.ts` and `AttributionModal.tsx` are deliberate provenance and stay.

## Screenshots

N/A — workflow change + comment-only edits, no visible UI surface.

- [x] Every new/renamed `className` in this PR has a matching CSS rule (N/A — no className changes)
- [x] Multi-viewport design QA: N/A — not UI

## Test plan

- [x] `npm test --workspace @bird-watch/frontend` — green (907 tests, 65 suites)
- [x] No new unit/integration tests needed — comment-only edits don't change behavior; test assertions themselves don't reference PostHog text
- [x] No new Playwright e2e spec needed — non-UI change
- [x] Live confirmation of the workflow drop: the next merged PR's `deploy-frontend` run will be the proof point — no more ✗ on the Post-deploy bundle-hash step

## Plan reference

Out of plan — loose-ends cleanup addressing #681 (dead CI check) and the R7 SUGGESTION trail from #682's bot review.

---

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -5,12 +5,6 @@ DATABASE_URL=postgres://birdwatch:birdwatch@localhost:5432/birdwatch
 EBIRD_API_KEY=your-ebird-api-key-here
 FRONTEND_ORIGINS=https://bird-maps.com,https://www.bird-maps.com
 
-# PostHog instrumentation (issue #357). Leave empty in CI / local dev —
-# `frontend/src/analytics.ts` skips `posthog.init` on an empty key, which
-# is required for console-cleanliness assertions in the e2e suite. Set
-# in Cloudflare Pages production env vars only.
-VITE_POSTHOG_KEY=
-
 # Adaptive cluster grid — default ON since 2026-05-15.
 # Remove this flag (along with all gated code paths) after 2026-08-13 if no rollback needed.
 VITE_FF_ADAPTIVE_GRID=true
--- a/.github/workflows/deploy-frontend.yml
+++ b/.github/workflows/deploy-frontend.yml
@@ -40,25 +40,3 @@ jobs:
           apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
           accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
           command: pages deploy frontend/dist --project-name=birdwatch --branch=main --commit-dirty=true
-      - name: Post-deploy bundle-hash check
-        run: |
-          set -euo pipefail
-          # Fetch the live site, extract the first /assets/*.js path, fetch that
-          # bundle, grep for the production API base URL. If the deployed bundle
-          # does not bake in api.bird-maps.com, the build env did not propagate
-          # and the site is broken - fail the workflow so the previous CF Pages
-          # deployment stays live (fail-forward).
-          HTML=$(curl -sSf --retry 5 --retry-delay 3 --retry-all-errors https://bird-maps.com/)
-          ASSET=$(printf '%s' "$HTML" | grep -oE '/assets/[^"]+\.js' | head -1)
-          if [ -z "$ASSET" ]; then
-            echo "ERROR: no /assets/*.js found in served HTML" >&2
-            printf '%s\n' "$HTML" | head -40 >&2
-            exit 1
-          fi
-          echo "Checking bundle: $ASSET"
-          COUNT=$(curl -sSf --retry 5 --retry-delay 3 --retry-all-errors "https://bird-maps.com$ASSET" | grep -c 'api.bird-maps.com' || true)
-          echo "api.bird-maps.com occurrences in bundle: $COUNT"
-          if [ "$COUNT" -lt 1 ]; then
-            echo "ERROR: deployed bundle does not contain api.bird-maps.com" >&2
-            exit 1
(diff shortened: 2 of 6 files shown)


julianken:
@Mergifyio queue

PR #630, opened 2026-05-18
feat(read-api): zoom-aware aggregation on /api/observations

julianken:
## Diagrams

```mermaid
flowchart LR
    FE[MapCanvas / App.tsx<br/>bbox + zoom] -->|"GET /api/observations?bbox=...&zoom=N"| API[read-api]
    API -->|"zoom &lt; 6 AND bbox"| AGG[getObservationsAggregated<br/>PostGIS round/grid GROUP BY]
    API -->|"zoom &gt;= 6 or no bbox"| OBS[getObservations<br/>per-row]
    AGG -->|"mode: 'aggregated'<br/>buckets (array)"| FE
    OBS -->|"mode: 'observations'<br/>data (array)"| FE
    FE -->|"expand buckets to<br/>synthetic Observations"| Render[supercluster +<br/>AdaptiveGrid pills]
```

```mermaid
stateDiagram-v2
    [*] --> ObservationsMode
    ObservationsMode --> AggregatedMode: bbox && zoom < 6
    AggregatedMode --> ObservationsMode: zoom >= 6 OR bbox cleared
    AggregatedMode: mode='aggregated' · grid=2/4/8 at z=3/4/5
    ObservationsMode: mode='observations' · per-row
```

## Summary

- Adds a `zoom` query parameter to `/api/observations`. When `bbox` is set and `zoom < 6` the read-api returns `{ mode: 'aggregated', buckets: [...] }` from a PostGIS `round(coord * multiplier) / multiplier` GROUP BY; otherwise it returns the existing `{ mode: 'observations', data: [...] }` shape. Closes the CONUS-payload gate surfaced by #626 (2.78 MB → < 50 KB at zoom 3-4) without changing user-visible UI — the frontend expands buckets to synthetic Observations and feeds them through the same supercluster / AdaptiveGrid pill pipeline.
- Grid multipliers per zoom: 2 (z<=3, ~55 km), 4 (z=4, ~28 km), 8 (z=5, ~14 km). Pre-computed in the API handler via a closed switch so a tampered zoom can never bypass aggregation. Bbox / since / notable / species / family filters all flow into the aggregated query unchanged.
- `ObservationsResponse` becomes a discriminated union on `mode`. The frontend `ApiClient` normalizes three legacy envelope shapes (bare array, post-#456 `{data,meta}` without `mode`, new union) so a mixed-deploy rollout window can't crash the UI. Cache-Control header is unchanged (`s-maxage=300, stale-while-revalidate=600`).

## Screenshots

[...]

PR #722, opened 2026-05-26
fix(frontend,map): gate synthetic species codes from clickable rows — closes #715

### LINKED ISSUE #715: bug(frontend): species detail panel does not render when zoomed out far
## Symptom

When the map is zoomed out far (z < 6, the aggregated-mode threshold), clicking a species inside a cluster's CellPopover opens a species detail panel that renders only the chrome (close button on desktop, sheet handle on mobile) plus a near-invisible error line "Could not load species details". The photo, common name, scientific name, family, phenology chart, and description are all absent. The user perceives this as "nothing rendered" because the error line is small and unstyled-as-content.

## Reproduction (confirmed live, 2026-05-26)

The fastest deterministic repro bypasses the map UI:

1. Open `https://bird-maps.com/?detail=agg-3-anatidae-2` in any viewport (mobile or desktop).
2. Observe: the rail (≥1200px) or sheet (≤1199px) mounts, but the body of the panel contains only the StatusBlock "Could not load species details". Network: `GET /api/species/agg-3-anatidae-2 → 404`.

User flow that produces the same URL state:

1. Load `bird-maps.com` at default zoom (z=3, MIN_ZOOM, the CONUS view).
2. Hover any cluster pill on desktop (or tap on touch) to open the CellPopover.
3. Click any row — e.g. `53x anatidae`.
4. URL becomes `?detail=agg-N-anatidae-M&bbox=...`. The detail panel mounts and shows only the error line.

Screenshots captured during repro:
- Mobile (390×844) bug confirmation: `repro-12-bug-confirmed.png` — pink "Could not load species details" strip at sheet's peek position; no other content.
- Desktop (1440×900) bug confirmation: `repro-13-desktop-bug.png` — rail on the right contains only the close `×` and the error message.
- Healthy comparison (real speciesCode `norcar` at z=3): `repro-08-detail-norcar.png` — full panel renders correctly. Proves the API and panel chrome are zoom-independent in isolation; the bug is in the speciesCode that gets routed.

## Root cause

At zoom < 6 the Read API returns `{ mode: 'aggregated', buckets: AggregatedBucket[] }` — buckets carry `families[]` only, no real `speciesCode` (this is intentional; the bucket-level wire format saves bandwidth — see `services/read-api/src/observations-handler.ts` for the mode-switch).

`expandBucketsToSyntheticObservations` in `frontend/src/data/use-bird-data.ts:19-48` turns each bucket into `b.count` synthetic `Observation` rows. To make the existing supercluster/adaptive-grid renderer work unchanged, it fabricates a non-null `speciesCode` for each synthetic row:

```ts
// frontend/src/data/use-bird-data.ts:31-33
speciesCode: family
  ? `agg-${bi}-${family}-${i % Math.max(b.speciesCount, 1)}`
  : `agg-${bi}-${i}`,
```

The fabrication is required for the distinct-species count to be accurate (issue #630 made the codes unique across buckets so a viewport-level `new Set(...).size` doesn't undercount). But `speciesCode` is also the load-bearing signal in three places:

1. `frontend/src/components/map/adaptive-grid.ts:267-302` (`aggregateClusterSpecies`) preserves the synthetic `speciesCode` on its way into the `SpeciesAggregate` rows passed to `CellPopover`.
2. `frontend/src/components/map/CellPopover.tsx:100-118`: `const clickable = s.speciesCode !== null` — synthetic codes are non-null, so the popover renders each row as a clickable `<a role="link">` whose `onClick` calls `onSelectSpecies(code)`.
3. `frontend/src/components/map/ClusterListPopover.tsx:190-209`: the identical gate `const clickable = s.speciesCode !== null` — synthetic codes are non-null, so the expanded-family rows also render as clickable `<a role="link">` whose `onClick` calls `onSpeciesRowClick(code)`. This is a second entry point to the same broken chain at the same zoom level.

Then the chain plays out:

- `App.tsx:300-304`: `onSelectSpecies('agg-3-anatidae-2', bbox)` → `set({ detail: 'agg-3-anatidae-2', bbox })` → URL becomes `?detail=agg-3-anatidae-2&bbox=...`.
- `App.tsx:443-462`: `state.detail` truthy → mount `SpeciesDetailRail` (desktop) or `SpeciesDetailSheet` (mobile) keyed on `state.detail`.
- `frontend/src/data/use-species-detail.ts:79-93`: `client.getSpecies('agg-3-anatidae-2')` → server replies 404 → `setState({ loading: false, error, data: null })`.
- `frontend/src/components/SpeciesDetailSurface.tsx:164-172`: renders `<StatusBlock state="error" title="Could not load species details" surface="panel" />`.
- The fallthrough to `data.comName` at line 191 never executes — `data` is null.

The "panel renders nothing" perception is real: the error StatusBlock has no body text (`body` prop omitted at the call site), and the panel chrome (close button / sheet handle) provides no other content. The user does not see "Couldn't load" prominently because the text sits in a small region inside a 120px peek strip on mobile or beside a 32px close button on desktop.

## Why "zoomed out far" specifically

At z ≥ 6 the server switches to `mode: 'observations'` (per-obs path) and returns real `speciesCode` values; the synthetic-code branch in `expandBucketsToSyntheticObservations` never runs at that zoom, so CellPopover gets real codes that resolve cleanly via `/api/species/:code`. The bug is therefore exclusively a low-zoom (z < 6) phenomenon — at any zoom the user reaches by drilling into a cluster, the symptom disappears because `mode` flips.

## Files involved

- `frontend/src/data/use-bird-data.ts:19-48` — `expandBucketsToSyntheticObservations` (synthetic code source)
- `frontend/src/components/map/adaptive-grid.ts:267-302` — propagates the synthetic code into SpeciesAggregate
- `frontend/src/components/map/CellPopover.tsx:100-118` — treats non-null speciesCode as "clickable" (entry point 1)
- `frontend/src/components/map/ClusterListPopover.tsx:190-209` — identical gate; treats non-null speciesCode as "clickable" (entry point 2)
- `frontend/src/components/SpeciesDetailSurface.tsx:164-172` — renders the near-blank error block
- `frontend/src/data/use-species-detail.ts:79-93` — fetches the synthetic code and 404s

## Proposed fix

Two required changes (Fix A applied at both call sites) plus two optional polish changes.

**Fix A (required, minimal):** make synthetic rows non-clickable in both popovers by routing the discrimination through `speciesCode`.

- Decide on a sentinel pattern. Two viable shapes:
  - **A1:** in `expandBucketsToSyntheticObservations`, set `speciesCode: null` for aggregated rows. The distinct-species count metric (#630) currently relies on the fabricated codes being unique — if we drop them, we have to derive the count from a different field. `subId` is already globally unique (`agg:${bi}:${i}`) and would substitute fine for any "distinct row" counting; only "distinct species within a family" counts would need to switch their key. Audit: `frontend/src/components/map/adaptive-grid.ts` and any consumer of `SpeciesAggregate.speciesCode` for counting logic.
  - **A2 (recommended):** keep the codes but prefix them with a recognizable sentinel (already in place — they start with `agg-`). Add an `isSyntheticCode(code)` helper exported alongside the expand function:
    ```ts
    // frontend/src/data/use-bird-data.ts
    export const isSyntheticCode = (code: string | null): boolean =>
      code !== null && code.startsWith('agg-');
    ```
    Then update **both** clickable gates to exclude synthetic codes:

    - `frontend/src/components/map/CellPopover.tsx:100` — change `const clickable = s.speciesCode !== null;` to `const clickable = s.speciesCode !== null && !isSyntheticCode(s.speciesCode);`.
    - `frontend/src/components/map/ClusterListPopover.tsx:190` — change `const clickable = s.speciesCode !== null;` to `const clickable = s.speciesCode !== null && !isSyntheticCode(s.speciesCode);`. **This is not optional** — ClusterListPopover is a second entry point reachable at the same zoom level (cluster list expansion) and shipping Fix A only at CellPopover leaves the bug live via this path. Both gates must change in the same PR.

  A2 is the smaller diff and preserves the #630 distinct-count behavior. Recommend A2.

- When `clickable === false`, the existing branch in each popover (`CellPopover.tsx:120-128` and the parallel branch in `ClusterListPopover.tsx`) already renders the row as a static `<span>` instead of an `<a>`. The visual change is: "53x anatidae" still appears but isn't a link.

**Fix B (UX polish, optional but valuable):** improve the popover footer copy at aggregated zoom from `'Click or tap for full list'` to something like `'Zoom in for species detail'` so the user understands why the rows are non-interactive. Gate on `species.every(s => !s.speciesCode || isSyntheticCode(s.speciesCode))` to detect aggregated mode without threading zoom down. Applies to both popovers.

**Fix C (defense in depth, optional):** in `useSpeciesDetail`, short-circuit synthetic codes:

```ts
// frontend/src/data/use-species-detail.ts:57
if (speciesCode === null || isSyntheticCode(speciesCode)) {
  setState({ loading: false, error: null, data: null });
  return;
}
```

Rationale: this prevents an unnecessary 404 network round-trip and the associated React `setState` churn if a synthetic code somehow lands in URL state (deep-link, manual URL edit, race condition). Note that module cache poisoning is **not** a concern here — `_moduleCache.set(speciesCode, meta)` at line 86 runs only inside the `.then()` branch, so a 404 already cannot poison the cache. The benefit is narrower: skip the network RTT and the loading/error state flap when we can statically prove the code is unresolvable. Combined with Fix A, no user flow can reach this state in practice; Fix C is a guard for non-UI entry points (URL bar paste, browser history restore, external links).

## Acceptance

- At z=3 (or any zoom < 6), hovering/tapping a cluster opens the popover (either CellPopover or ClusterListPopover via the expanded-family path). Rows representing aggregated species are visible but non-clickable (cursor stays as `default`, keyboard navigation skips them). No URL change happens when they're activated. **Both popovers must satisfy this** — verify against both code paths.
- The user can still tap the cluster pill itself (outside the popover) to drill in via expansion-zoom — that path is unaffected.
- At z ≥ 6 (real `observations` mode), popover rows behave exactly as before: clickable, opening a fully-populated detail panel. Verify in both CellPopover and ClusterListPopover.
- E2E coverage: a new spec under `frontend/e2e/` that loads the app at z=3, opens a CellPopover and a ClusterListPopover, asserts (a) no popover rows have `role="link"` in either, (b) clicking/Enter/Space is a no-op (URL `?detail=` remains absent), (c) the popover footer reads the "zoom in" copy if Fix B is included.
- Unit coverage: extend `frontend/src/data/use-bird-data.test.tsx` to assert `isSyntheticCode` returns true for `agg-*` and false for real codes (e.g. `norcar`, `wlsfly`). Extend `frontend/src/components/map/CellPopover.test.tsx` AND add/extend `frontend/src/components/map/ClusterListPopover.test.tsx` to assert synthetic-coded rows render as `<span>` and ignore Enter/Space in both components.

julianken:
## Diagrams

```mermaid
sequenceDiagram
    autonumber
    participant User
    participant Popover as CellPopover / ClusterListPopover
    participant App as App.tsx (state-url)
    participant Hook as useSpeciesDetail
    participant API as GET /api/species/:code
    participant Surface as SpeciesDetailSurface

    rect rgb(255, 235, 235)
      Note over User,Surface: BEFORE — synthetic agg-* code leaks all the way to a 404
      User->>Popover: click "53x anatidae" row
      Popover->>App: onSelectSpecies("agg-3-anatidae-2")
      App->>Hook: useSpeciesDetail("agg-3-anatidae-2")
      Hook->>API: fetch /api/species/agg-3-anatidae-2
      API-->>Hook: 404
      Hook-->>Surface: { error, data: null }
      Surface-->>User: near-invisible "Could not load species details"
    end

    rect rgb(232, 245, 233)
      Note over User,Surface: AFTER — gate at both popovers&#59; defense-in-depth in the hook
      User->>Popover: click "53x anatidae" row
      Popover->>Popover: isSyntheticCode("agg-...") === true
      Note right of Popover: row renders as static span&#59; no callback, no URL change
    end
```

## Summary

- Add `isSyntheticCode(code)` in `frontend/src/data/use-bird-data.ts` — single source of truth for the `agg-*` prefix discriminator.
- Gate the clickable-row predicate in **both** `CellPopover` and `ClusterListPopover` on `!isSyntheticCode(...)`; synthetic rows fall through to the existing static-`<span>` branch (same treatment as spuh / slash / hybrid taxa).
- Short-circuit `useSpeciesDetail` for synthetic codes as defense in depth (URL paste, history restore, external links bookmarked at aggregated zoom); avoids the 404 RTT and the loading / error state flap.

The root cause is in the issue: at z<6 the Read API returns aggregated buckets; `expandBucketsToSyntheticObservations` fabricates `agg-${bi}-${family}-${i}` codes per row so the existing supercluster renderer works unchanged, and #630 requires those codes be unique across buckets for distinct-species counting. Both popovers' clickable-row gates were checking `speciesCode !== null` only — synthetic codes are non-null, so they read as clickable, the click wrote `?detail=agg-...` into the URL, `/api/species/agg-...` 404'd, and the detail panel rendered only the near-invisible error StatusBlock. Fix A (the popover gates) is required at both call sites — `ClusterListPopover` is a second entry point reachable at the same zoom level; shipping it only at `CellPopover` leaves the bug live via cluster-list expansion. Fix C (the hook short-circuit) is optional but tiny.

## Screenshots

All captures are `/?detail=agg-3-anatidae-2` (the issue's deterministic deep-link repro) against the local dev build of this branch. The load-bearing assertion: **the "Could not load species details" error block is absent from every capture** — the rail (desktop) or sheet-in-peek-state (mobile) mounts with an empty body, and the close affordance remains reachable. The local read-api had no DB attached during the wide-1920 captures (last two), so those show the page-level "Couldn't load bird data" panel; they still demonstrate that the detail-panel-specific error block does NOT render.

| Viewport | Light | Dark |
|---|---|---|
| Mobile 390×844 (iPhone 14 Pro) | mobile-390-light | mobile-390-dark |
| Tablet 768×1024 (iPad portrait) | tablet-768-light | tablet-768-dark |
| Tablet 1024×768 (iPad landscape) | tablet-1024-light | tablet-1024-dark |
| Desktop 1440×900 | desktop-1440-light | desktop-1440-dark |
| Wide 1920×1080 | wide-1920-light | wide-1920-dark |

- [x] Every new/renamed `className` in this PR has a matching CSS rule — N/A, no new classNames (Fix A reuses the existing `cell-popover__row` / `cluster-list-popover__row` static-`<span>` branch verbatim; Fix C lives in a hook).
- [x] Multi-viewport design QA: PR body has ≥10 `user-attachments` URLs (5 viewports × 2 themes).

## Test plan

- [x] `npm run test --workspace @bird-watch/frontend` — green (860 tests, including 3 new `isSyntheticCode` cases in `use-bird-data.test.tsx`, 3 new synthetic-code cases in `CellPopover.test.tsx`, 2 new in `ClusterListPopover.test.tsx`, 1 new in `use-species-detail.test.ts`).
- [x] `npm run build --workspace @bird-watch/frontend` — clean production build.
- [x] `npm run lint` — clean (forbidden-token grep passes; no workspace has a lint script, by design).
- [x] `npx knip --no-progress` — exit 0 (one pre-existing configuration hint, unrelated).
- [x] `bash scripts/check-orphan-classnames.sh` — PASS.
- [x] New Playwright e2e spec `frontend/e2e/synthetic-species-code-gate.spec.ts` — 2 tests (desktop 1440×900, mobile 390×844) green locally. Asserts (a) the "Could not load species details" error block does NOT render, (b) no `/api/species/agg-*` fetch fires (Fix C short-circuits), (c) the close affordance / sheet is still reachable.
- [x] Playwright MCP smoke — drove `?detail=agg-3-anatidae-2` at all 5 canonical viewports × 2 themes; screenshots above. The "Could not load species details" text is ABSENT in every capture, which is the load-bearing UI assertion.

## Plan reference

Out of plan — production bug fix for #715, triaged from a live repro on bird-maps.com (2026-05-26). No `docs/plans/` entry; the fix is small (single helper + two one-line predicate updates + one hook short-circuit) and the test plan is captured in the issue.

---


julianken:
@Mergifyio queue

PR #724, opened 2026-05-26
fix(frontend,map): max-zoom cluster pill opens ClusterListPopover

### LINKED ISSUE #717: bug(frontend,map): max-zoom cluster pill silently no-ops; should open ClusterListPopover
## Symptom

When a cluster pill represents observations at (or near) coincident coordinates, the user can zoom in until the camera reaches `CLUSTER_MAX_ZOOM` (22) but the pill never breaks apart — supercluster has run out of resolution. Clicking the pill does nothing.

The expected behavior is the same as the coarse-pointer `AdaptiveGridMarker` outer-tap path: open the `ClusterListPopover` listing every species (grouped by family), so the user can pick one and read its detail.

## Root cause (confirmed by static trace)

`handleGroupClick` in `frontend/src/components/map/MapCanvas.tsx:1284-1362`:

```ts
const targetZoom = Math.min(Math.max(...zooms), CLUSTER_MAX_ZOOM);
const currentZoom = map.getZoom();
if (
  targetZoom > currentZoom &&
  anchor.longitude !== undefined &&
  anchor.latitude !== undefined
) {
  map.easeTo({ ... });
}
// no else — falls off the end of the try block, silent no-op
```

At max zoom on a coincident cluster, `Math.max(...zooms)` returns ≤ 22, `Math.min(22, 22) = 22`, `22 > 22` is `false`. The function returns without doing anything. Same failure mode for any `targetZoom <= currentZoom` (not strictly the max-zoom case — also fires if a cluster wants the camera to zoom out, which `easeTo` would honor but the gate forbids).

Adjacent failure modes the fix should also defend against:
- `getClusterExpansionZoom` returning a value `<= currentZoom` for a still-clustered marker (any expansion that doesn't escape the current zoom level).
- `getClusterExpansionZoom` resolves to `NaN` (maplibre/supercluster library-error state — not normal cluster geometry; the empty-array case is already guarded at `MapCanvas.tsx:1319-1334` because all-silhouette groups exit early). `Math.max(NaN)` is `NaN`, comparison is `false`, silent no-op. Treat as "open the popover" too.

Routing context:
- `ClusterPill` (`frontend/src/components/ds/ClusterPill.tsx`): a plain `<button>` whose only callback is `onClick: () => void`. No popover state. Wired in `MapCanvas.tsx:1538` as `onClick={() => handleGroupClick(g)}`.
- `AdaptiveGridMarker` (`frontend/src/components/map/AdaptiveGridMarker.tsx`): on **coarse pointer** the outer-button click branches into `setIsClusterListOpen(true)` internally (lines 268-279, 379-396) — it does NOT reach `handleGroupClick`, so coarse-pointer grid markers already do the right thing. On **fine pointer** each cell is its own `<button>` with `CellPopover`; the outer is `role="group"` and not clickable. The fine-pointer grid is therefore not affected.
- The bug surface is specifically the `ClusterPill` route on every pointer kind, plus the (rare) coarse-pointer route for *very dense* clusters where `pickGridShape` returns `{ tag: 'pill' }` (uniqueFamilies > 16 OR pointCount > 64).

### Option A (smaller diff, recommended): fallback in `handleGroupClick`

Add an else-branch path that fetches leaves, aggregates into the shape `ClusterListPopover` already expects, and opens it. Plumb an anchor element through `ClusterPill.onClick` so the popover can attach.

Insertion point — `MapCanvas.tsx:1344-1354` (replace the existing `if (targetZoom > currentZoom && ...)` block with an if/else):

```ts
const shouldEase =
  Number.isFinite(targetZoom) &&
  targetZoom > currentZoom &&
  anchor.longitude !== undefined &&
  anchor.latitude !== undefined;

if (shouldEase) {
  map.easeTo({
    center: [anchor.longitude!, anchor.latitude!],
    zoom: targetZoom,
    ...(prefersReducedMotion ? { duration: 0 } : {}),
  });
} else {
  // Camera already at the zoom where this cluster bottoms out (or
  // getClusterExpansionZoom resolved to NaN — library-error guard).
  // Open ClusterListPopover so the user can still drill in by species —
  // mirrors the coarse-pointer AdaptiveGridMarker outer-tap path.
  await openClusterListFromGroup(group);
}
```

New helper, declared next to `handleGroupClick`:

```ts
const openClusterListFromGroup = useCallback(async (group: DeconflictGroup) => {
  const map = mapRef.current?.getMap();
  if (!map) return;
  const source = map.getSource('observations') as {
    getClusterLeaves?: (id: number, limit: number, offset: number) => Promise<unknown[]>;
  } | undefined;
  if (!source?.getClusterLeaves) return;

  // Aggregate leaves across every real cluster member (silhouettes filtered).
  const realIds = group.memberIds.filter((id) => id > 0);
  if (realIds.length === 0) return;

  try {
    const leafBatches = await Promise.all(
      realIds.map((id) => source.getClusterLeaves!(id, 64, 0) as Promise<ClusterLeafFeature[]>),
    );
    const leaves = leafBatches.flat();
    const families = aggregateClusterFamilies(leaves);
    const speciesByFamily = aggregateClusterSpecies(leaves);
    setClusterList({
      families,
      speciesByFamily,
      totalCount: group.anchor.point_count,
      uniqueFamilies: group.anchor.uniqueFamilies,
      anchorEl, // passed in from handleGroupClick via e.currentTarget
    });
  } catch {
    // getClusterLeaves can reject for recycled cluster_ids; match the
    // err-swallow pattern from the easeTo branch above.
  }
}, []);
```

State + render:

```ts
const [clusterList, setClusterList] = useState<{
  families: FamilyAggregate[];
  speciesByFamily: ReadonlyMap<string, ReadonlyArray<SpeciesAggregate>>;
  totalCount: number;
  uniqueFamilies: number;
  anchorEl: HTMLElement | null;
} | null>(null);

// ... near <ObservationPopover ...>:
{clusterList && clusterList.anchorEl && (
  <ClusterListPopover
    families={clusterList.families}
    speciesByFamily={clusterList.speciesByFamily}
    totalCount={clusterList.totalCount}
    uniqueFamilies={clusterList.uniqueFamilies}
    anchorEl={clusterList.anchorEl}
    onDismiss={() => setClusterList(null)}
    onSelectSpecies={(code) => {
      onSelectSpecies?.(code, getClusterBbox(/* synthesize from clusterList */));
      setClusterList(null);
    }}
  />
)}
```

### Anchor element (the load-bearing piece)

`ClusterListPopover` requires `anchorEl: HTMLElement` (for focus return on dismiss, click-outside detection, and positioning the popover relative to the marker — `ClusterListPopover.tsx:37`, `100`).

**Pass the click event through.** Widen `ClusterPillProps.onClick` from `() => void` to `(e: ReactMouseEvent<HTMLButtonElement>) => void`. Pass `e.currentTarget` into `handleGroupClick` as a second arg (`handleGroupClick(g, e.currentTarget)`), thread the element into `openClusterListFromGroup`. This is race-free — `e.currentTarget` is guaranteed to be the clicked button at the moment of the click, and it's the correct element for `anchorEl.focus()` on dismiss. A ref-map fallback (`pillRefs.current.get(group.key)`) was considered and rejected: it can return null if the ref callback races, silently suppressing the popover. Existing tests at `ClusterPill.test.tsx` need updates but the signature change is isolated.

## Files to touch

- `frontend/src/components/ds/ClusterPill.tsx` — widen `onClick` signature to `(e: ReactMouseEvent<HTMLButtonElement>) => void`.
- `frontend/src/components/map/MapCanvas.tsx` — add `ClusterListPopover` import, add state, add `openClusterListFromGroup` helper, restructure `handleGroupClick` else branch, render the popover.
- `frontend/src/components/map/MapCanvas.test.tsx` — add cases:
  - max-zoom coincident pill click opens `ClusterListPopover` with the aggregated species list (re-uses the existing `getClusterLeaves` mock fixture, e.g. lines 454-457).
  - `targetZoom > currentZoom` case still calls `easeTo` (regression guard for the existing path at lines 356-376).
  - **library-error guard**: when `getClusterExpansionZoom` resolves to `NaN` (mock the maplibre source method to return NaN — this is a supercluster/maplibre internal error state, not normal cluster geometry), `handleGroupClick` opens the popover instead of silently no-op'ing.
  - `easeTo` is NOT called in the max-zoom case.
- Optional: extend `frontend/e2e/cluster-interactions.spec.ts` (or nearest spec) with a tap-at-max-zoom test. Defer if it requires fixture engineering disproportionate to the bug.

## Acceptance

- AC1: At zoom 22 with a pill anchored at coincident coordinates, clicking the pill opens `ClusterListPopover` listing every species, grouped by family, sorted descending by count.
- AC2: At any zoom where `targetZoom > currentZoom`, the camera still `easeTo`s and the popover does NOT open (regression guard).
- AC3: Clicking `Done` / ESC / outside dismisses the popover and returns focus to the pill button (existing `ClusterListPopover` behavior, must hold under the new mount point).
- AC4: WCAG 2.5.5 tap target from #565 unaffected — `ClusterListPopover` rows keep their existing min height. Pill button stays ≥ its current size.
- AC5: No regression on `AdaptiveGridMarker` coarse-pointer path — the marker still opens its own internal `ClusterListPopover`. The new `MapCanvas`-level popover only mounts when the click came from a `ClusterPill` (or, optionally, a coarse-pointer fine-grained grid that wants to escalate, but that path already works internally so the simpler implementation is "pill-only").
- AC6: Console clean — no warnings about missing `anchorEl`, no React 18 strict-mode double-mount complaints from popover state.

## Related

- #565 (popover species-row tap target — must not regress)
- #556 (per-cell species popover epic — context for `ClusterListPopover` lineage)
- #559 (Phase 2 issue that introduced `ClusterListPopover` for coarse pointer)
- #554 (deconflict + unified `DeconflictGroup` — `handleGroupClick` lives here)

julianken:
## Diagrams

```mermaid
sequenceDiagram
    participant User
    participant ClusterPill
    participant handleGroupClick
    participant supercluster
    participant ClusterListPopover

    User->>ClusterPill: click (camera at max zoom)
    ClusterPill->>handleGroupClick: (group, e.currentTarget)
    handleGroupClick->>supercluster: getClusterExpansionZoom(id)
    supercluster-->>handleGroupClick: 22 (or NaN)
    Note over handleGroupClick: targetZoom <= currentZoom<br/>OR !Number.isFinite(targetZoom)
    handleGroupClick->>supercluster: getClusterLeaves(id, 64, 0)
    supercluster-->>handleGroupClick: leaves
    handleGroupClick->>ClusterListPopover: setClusterList({families, speciesByFamily, anchorEl})
    ClusterListPopover-->>User: species list, grouped by family
```

## Summary

- Closes #717. `handleGroupClick` previously fell off the end of the try block when `targetZoom <= currentZoom` — every coincident-hotspot pill clicked at zoom 22, plus the maplibre/supercluster NaN-return error path. The user saw a dead click with no way to drill in.
- Adds an `else` branch that mounts `ClusterListPopover` at the MapCanvas level via a new `openClusterListFromGroup` helper. Mirrors the coarse-pointer `AdaptiveGridMarker` outer-tap path — same component, new mount point. `ClusterPill.onClick` widens to receive the synthetic event so `e.currentTarget` is the race-free popover anchor (ref-map fallback was considered and rejected per issue body).
- `Number.isFinite(targetZoom)` defends against library-error NaN returns from supercluster's `getClusterExpansionZoom`.

## Screenshots

Captured via a Node + Playwright script that stubs the read-api with a coincident hotspot (60 demo species, 20 families, one lng/lat). Two clicks: the first eases the camera; the second hits the new max-zoom `else` branch and mounts `ClusterListPopover`.

**Desktop 1440×900**

Light | Dark
:---:|:---:
 | 

**Wide desktop 1920×1080**

Light | Dark
:---:|:---:
 | 

**iPad landscape 1024×768**

Light | Dark
:---:|:---:
 | 

**iPad portrait 768×1024**

Light | Dark
:---:|:---:
 | 

**iPhone 14 Pro 390×844**

Light | Dark
:---:|:---:
 | 

- [x] Every new/renamed `className` in this PR has a matching CSS rule (N/A — no new classNames; the popover component already ships its own CSS for the existing `AdaptiveGridMarker` mount point)
- [x] Multi-viewport design QA: PR body has ≥10 `user-attachments` URLs (10 above, 5 viewports × 2 themes)

## Test plan

- [x] `npm run typecheck && npm run test` — green (854/854 frontend unit tests, including 3 new MapCanvas.test.tsx cases for #717)
- [x] New unit tests added — max-zoom pill click opens popover, regression guard for `targetZoom > currentZoom` (easeTo still fires), and the NaN library-error guard
- [ ] New Playwright e2e spec added — deferred per issue body ("Optional: extend `frontend/e2e/cluster-interactions.spec.ts` ... Defer if it requires fixture engineering disproportionate to the bug.")
- [x] UI verified via a Node Playwright capture script (5 viewports × 2 themes): each capture shows `ClusterListPopover` mounted with "Cluster: 60 observations, 20 families" and the family/species tree (Aegithalidae, Cardinalidae, Certhiidae, Columbidae, Corvidae, Fringillidae, Hirundinidae, ...)

## Plan reference

Out of plan — bug fix surfaced after the production ship (#717).

---


julianken:
@Mergifyio queue

Issue #764, opened 2026-05-29
test(frontend,e2e): artboard verification — a11y + scope-transition + label-bleed coverage, clean-console gate, manual 5×2 design-review (#760)


julianken:
Background — part of epic #760 (state artboard mask)

This is **SUB3 of 3** under epic #760. It adds the verification layer on top of the assembled artboard feature. Two things to be honest about up front, because they drive the whole shape of this issue:

1. **The deterministic regression coverage lives in the UNIT layer, not e2e.** Every meaningful artboard assertion (the padded-clamp value, the layer-count delta, the label-isolation `within` filter, reduced-motion durations) reads WebGL/`maplibre` state that **does not populate in headless CI** — the e2e `__birdMap` reads skip cleanly there and therefore gate nothing on their own. The precedent is already in the repo: `state-scope.spec.ts` asserts the camera via a URL round-trip *precisely because* it can't rely on WebGL, and `map-symbol-layer.spec.ts` `test.skip()`s when `onLoad` never fires headless. So this issue (a) keeps finding-1 + label-bleed regression coverage in the **unit** suite (`mask.test.ts` / `MapCanvas.test.tsx` from SUB1, plus SUB2's `applyLabelIsolation` filter unit test), and (b) treats the `__birdMap` e2e assertions as **best-effort** — useful when WebGL is available locally, skipped (not failed) when it isn't.

2. **The 5-viewport × 2-theme `ui-design` dispatch is the MANUAL pre-merge gate, not automated coverage.** It is a deliverable of *this PR's review process*, run live by the implementer/orchestrator — it is not a CI check and it does not provide automated regression coverage. The text below frames it that way and stops claiming the e2e covers what only the manual gate covers.

Depends on **#762** (mask fill + camera-decouple + asset/`mask.ts`) and **#763** (label isolation + halo/outline float) already merged to `main`. This sub-issue writes **no production code** — it is test-only under `frontend/e2e/**` plus the manual design-review pass.

What to reuse from the prototype (`feat/state-artboard-mask-760`, mined reference — NOT merged)
[...]

Issue #762, opened 2026-05-29
feat(map): state-artboard core — mask fill + camera + client polygon asset (#760)


julianken:
## Background

This is sub-issue **1 of 3** under epic **#760** (state-artboard mask). It delivers the *core* layer of the agreed v3 artboard design: a `?state=US-XX` scope renders as a Sketch-style artboard — only the selected state shows basemap + observations, everything outside its polygon is flat opaque theme-aware gray, and the user can zoom OUT to watch the state shrink on the gray field, bounded by a padded artboard margin (not an infinite void).

A working prototype was built in isolation on branch **`feat/state-artboard-mask-760`** (commits `0526ebb`, `c85a2a7`, `3cef141`, `1198637`, `61315d4`, plan `0e850cd`). **Mine it as reference — it is NOT merged and NOT done.** The prototype validated the geometry, hook, camera-decouple, and render wiring against real viewports; this sub-issue re-lands that proven code with the corrections the v3 review demands. Canonical layer ids (mined from the prototype, locked for SUB2/SUB3): mask **source `state-mask`**, **fill layer `state-mask-fill`**.

The three v3 corrections layered on top of the prototype:

1. **Author the client-asset emit net-new inside `scripts/generate-state-boundaries.mjs`** — the prototype used a *separate* `scripts/generate-state-mask-polygons.mjs`; that script lives only on the prototype branch and **does not exist on `main`**. There is nothing to delete; the emit is appended to the existing generator (preferred resolution of reviewer finding 3).
2. **Correct the mask fill colors** — the prototype's `#e7e2d6` / `#161b27` were a dark-on-dark mistake; v3 uses `#d8d8d8` (light) / `#06090e` (dark, *darker* than positron-dark land so the state floats as the lit element).
3. **Add `renderWorldCopies: false` — CONDITIONAL on `maskPolygon`** (the prototype never set it). It must be gated so the in-place `state→us` prop update does not break the `?scope=us` wide-viewport view (see Implementation note 4).

This sub-issue is **render + camera + data only**. Label isolation, layer sinking, and halo/outline float layers are **#763**; e2e + clean-console + design-review are **#764** (see epic #760 for siblings).

## Scope

In:
- **Append** the polygon-stripping emit to `scripts/generate-state-boundaries.mjs` so `data/us-state-polygons.geojson` and `frontend/public/state-polygons.json` are emitted from **one** generator run. Do NOT create a separate script.
[...]

PR #766, opened 2026-05-29
feat(frontend,map): artboard fidelity — label isolation + clean exterior + float layers (#763)

### LINKED ISSUE #763: feat(frontend,map): artboard fidelity — label isolation + clean exterior + float layers (#760)
Part of epic **#760** (state-artboard mask). This is **SUB2 of 3**, sequenced after #762 (mask-fill core) and before #764 (verification). It is the highest-risk sub-issue: pure imperative MapLibre layer manipulation on a live style, re-applied across `setStyle` theme swaps. It gets its own focused **opus** review and a live Playwright pass before review.

## Background

#762 lands the inverse-mask fill: a single `fill` layer (`state-mask` source → `state-mask-fill` layer) painting flat opaque theme-aware gray everywhere except the selected state, plus the fit/clamp decouple, `MIN_ZOOM=2`, and the `maskPolygon`-conditional `renderWorldCopies:false`. With #762 the artboard reads correctly as a *shape* — but the basemap symbol layers (place names, settlement labels, POI labels) still render across the whole world, so labels for OTHER states bleed onto the gray field, and labels straddling the state border get visually sliced by the opaque fill. #762 also does not add the artboard "float" (halo/outline) that the v3 design calls for.

SUB2 makes the artboard look *finished*: exterior labels gone, interior labels whole and un-sliced, no boundary/coastline bleed into the gray, and a theme-aware halo + crisp outline so the state floats above the field.

**Inherited epic a11y ACs (#760):** the state boundary is a load-bearing accessibility element, not decoration — the mask fill alone fails WCAG 1.4.11 non-text contrast (verified plausible against #762's v3 tokens: `MASK_FILL_DARK = '#06090e'` ≈ 1.05:1 dark, `MASK_FILL_LIGHT = '#d8d8d8'` ≈ 1.26:1 light, both against the unmasked state land). SUB2 owns making that boundary meet **≥3:1 non-text contrast in both themes**. (The polite-live-region scope announcement is owned by **#762** — #764 only verifies it; SUB2 must not regress it.)

**Reuse from the prototype branch `feat/state-artboard-mask-760` (mined reference, NOT merged):** SUB2 is NEW work not present in the prototype (the prototype validated the fill + camera only — commits `0526ebb`, `c85a2a7`, `3cef141`, `1198637`, `61315d4`). What you reuse is the prototype's *integration surface*, all of which #762 will have landed on `main` by the time you start:
- `frontend/src/components/map/mask.ts` — `buildMaskFeature` returns a `Feature<Polygon>`; you need the SAME state polygon (a `MultiPolygon` from `useStatePolygon`) as the basis for the `['within', …]` filter geometry (then OUTWARD-BUFFER it for the label test — see "Buffered within vs. exact fill"). Reuse the polygon, do not re-fetch it.
- `MapCanvas.tsx` MutationObserver (prototype ~L1511–1528, calls `map.setStyle(style)` then `setMaskTheme(next)`; current `main` shifts after #762) — this is the single theme-swap site you hook into.
- The canonical `state-mask` source / `state-mask-fill` layer id (#762) is the z-order anchor: SUB2's "sink stray layers" step moves bleed-prone basemap fill/line layers *below* it, and the float (halo/outline) layers go *above* it.

## Scope

In `frontend/src/components/map/MapCanvas.tsx` (and a new helper module — see Implementation notes), when a state scope is active (`maskPolygon` set):

1. **Within-filter the basemap symbol layers** to the state polygon so exterior labels do not render and none get sliced. Capture each touched layer's ORIGINAL filter first, merge `['within', isolationPolygon]` into it (combining with the original via `['all', original, withinExpr]` when an original exists), and restore the captured original when the mask unmounts (scope → us/chooser) so the whole-US and chooser views are untouched. **Critical:** the geometry fed to `['within', …]` is a slightly OUTWARD-BUFFERED state polygon, NOT the exact simplified `maskPolygon` — see "Buffered within vs. exact fill" below; getting this wrong silently drops near-border interior labels.
2. **Match symbol layers by type + name heuristic**, NOT by hardcoded id — positron(light) and dark layer IDs are not stable across the two styles. Iterate `map.getStyle().layers`, select `layer.type === 'symbol'` whose id matches a conservative name pattern (e.g. `/(^|[-_])(place|settlement|poi|label|town|city|village|state|country)([-_]|$)/i`). The pattern must **fail open**, not closed: a new/unmatched symbol layer in a future basemap release renders exterior (detectable in QA) rather than throwing or blanking the whole map.
3. **Re-apply across theme swaps — but SPLIT the re-apply by reconcile timing (see the sequencing blocker in Implementation notes).** `map.setStyle()` clears and asynchronously reloads ALL layers, dropping both the merged filters and the float layers. Label-filter isolation re-applies inside `style.load` (basemap symbol layers exist immediately and need no reference layer); the float `addLayer` + the `moveLayer` stray-sink re-apply from a separate `maskPolygon`-watching effect that runs *after* react-map-gl re-adds `state-mask-fill`. Re-capture originals against the NEW style.
4. **Sink stray basemap fill/line layers beneath the mask** via `map.moveLayer(strayId, 'state-mask-fill')` so no boundary/coastline/water-outline bleeds *on top of* the gray. Guard with `map.getLayer('state-mask-fill')` first (see blocker below).
5. **Halo + outline float layers** above the mask, theme-aware: a blurred halo `line` layer (`line-blur`; a soft drop-shadow in light, a soft glow in dark) plus a **crisp `line` outline that is the a11y-load-bearing boundary** (≥3:1 contrast target — item below), both tracing the state polygon's exterior so the artboard "floats". These mount/unmount with the mask and re-tint on theme swap.

Out of scope (other sub-issues): the mask fill + camera + asset/hook/generator + scope-announcement live region (#762); e2e specs, console-gate suppression, axe scan, and the 5-viewport × 2-theme ui-design design-review gate (#764). SUB2 still drives a live Playwright pass for its own implementer verification, but the formal design-review gate is #764's.

## Implementation notes

*(Line numbers are approximate — read current `main` after #762 merges; the prototype diff shifted MapCanvas by ~120 lines and #762 will shift it again.)*

### [blocker] moveLayer / style.load reconcile-sequencing — do NOT re-apply everything in `style.load`

This is the dominant technical blocker and the reason the re-apply is split. `moveLayer(strayId, 'state-mask-fill')` called inside `style.load` throws `Cannot move layer before non-existing layer`: react-map-gl re-adds its managed declarative layers (including `state-mask-fill`) on the *next* React reconcile pass, which has NOT happened yet when `style.load` fires. So `state-mask-fill` does not exist at `style.load` time. Same hazard for the float `addLayer` if it uses `state-mask-fill` as a `beforeId`/insert anchor. SPLIT the re-apply:

- **(a) In the `style.load` handler, re-apply ONLY label-filter isolation.** Basemap symbol layers exist immediately on style load; `applyLabelIsolation` needs no reference to `state-mask-fill`, so this is safe here.
- **(b) Move the float-layer `addLayer` + the `moveLayer` stray-sink into a `mapReady`-gated `useEffect` that watches `maskPolygon`.** It re-fires on the next render — AFTER react-map-gl's reconcile has re-added `state-mask-fill` — so the reference layer exists by then. **Guard with `map.getLayer('state-mask-fill')` before any `moveLayer`/insert; if absent, `console.warn` and return — do NOT call through** (calling through is the exact throw we are avoiding).
- **Document the sequencing dependency** in a comment at the split so a future implementer doesn't "simplify" it back into a single `style.load` re-apply and silently reintroduce the throw.

*(Orchestrator-only optional, exercise only if SUB2 review stalls: split SUB2 into 2a = float layers (trivial, near-zero-risk cosmetics) and 2b = label isolation + theme re-apply. Note these are coupled through the same `maskPolygon`-gated effect and the same reconcile-sequencing fix, so the split does not cleanly de-risk; keep SUB2 whole by default.)*

### Helper module

- **Put the imperative logic in a new, unit-testable helper module** rather than inline in MapCanvas, so it can be tested without a full React render. Suggested: `frontend/src/components/map/artboard-layers.ts`. **Public API is the function exports below** (`SavedFilters` may be a private alias or an export — implementer's choice, see knip note below):
  - `applyLabelIsolation(map, isolationPolygon): SavedFilters` — iterates `map.getStyle().layers`, matches symbol layers (item 2), captures `{ [layerId]: originalFilter }`, calls `map.setFilter(id, mergedExpr)` with `['within', isolationPolygon]`, ends with `map.triggerRepaint()`, returns the saved map. **`isolationPolygon` is the OUTWARD-BUFFERED polygon (see "Buffered within vs. exact fill"), NOT the exact `maskPolygon`.**
  - `bufferIsolationPolygon(maskPolygon, distanceKm)` (or equivalent inline helper) — returns an outward-expanded copy of the state polygon used ONLY for the `within` label test. **Units matter:** `@turf/buffer`'s distance is in **kilometers** by default (NOT degrees) — use **~8 km** (tunable), enough to clear the 5%-simplified edge. `~0.05–0.1°` would be the WRONG unit for turf (≈5–11 km only near the equator, and far smaller in linear terms than intended), and the "bbox strictly larger" AC would pass anyway, so the mistake is silent. The dependency-free fallback (per-ring outward nudge / padded-bbox) works in **degrees** (~`0.1°`) — keep degrees ONLY for that path.
  - `restoreLabelIsolation(map, saved)` — `map.setFilter(id, original)` for each captured entry (passing `undefined`/`null` restores "no filter"), then `map.triggerRepaint()`.
  - `sinkStrayLayersBelowMask(map, maskLayerId)` — `moveLayer` basemap fill/line layers under the mask (caller has already `getLayer`-guarded `state-mask-fill`).
  - `addFloatLayers(map, maskPolygon, maskLayerId, theme)` / `removeFloatLayers(map)` — add the halo + crisp-outline `line` layers above the mask; theme param drives the paint.
  - `applyArtboardFidelity(map, maskPolygon, theme)` — composite for the float/sink half (item 3b); returns the saved-filters handle from the isolation half where appropriate for teardown.
  - **`SavedFilters` may stay a private `type` alias inside `artboard-layers.ts` as a minor style preference** — at the `MapCanvas.tsx` call site, `ReturnType<typeof applyLabelIsolation>` is a fine explicit annotation, so the alias need not be exported. This is a readability nicety only, **NOT a CI requirement**: a type/alias exported from `artboard-layers.ts` and consumed cross-file in `MapCanvas.tsx` is plainly "used" by knip and would NOT be flagged (knip's `ignoreExportsUsedInFile: true` at `knip.ts:134` is *same-file* suppression only — it has no bearing on a cross-file consumer). Export it or keep it private; either is knip-clean.
  - **Knip-safety:** the real knip constraint is no *orphan* exports — every export above (function or type) must have at least one consumer in the SAME PR. Co-locate; do not stage a helper ahead of its consumer.

- **`triggerRepaint` (defensive, with a why-comment):** `setFilter` already schedules a render on its own, so this is belt-and-suspenders — a force-flush in case the map is idle when the filter changes. Frame it that way in the comment (defensive idle-map flush), NOT as a load-bearing premise that the filter won't otherwise apply. Call `map.triggerRepaint()` once at the end of `applyLabelIsolation` AND of `restoreLabelIsolation`.

- **Within-filter expression:** the geometry argument to `['within', geom]` is a GeoJSON `Polygon`/`MultiPolygon` *geometry object* — the OUTWARD-BUFFERED state polygon (`isolationPolygon`, see "Buffered within vs. exact fill" below), NOT the exact `maskPolygon` and NOT the inverse world-ring `Feature` that `buildMaskFeature` produces (that one is for the fill). Merge with the captured original as `['all', original, ['within', isolationPolygon]]`; when `getFilter` returns nothing, the merged filter is just `['within', isolationPolygon]`.

- **Where to hook the re-apply:** the MutationObserver (prototype ~L1524 `map.setStyle(style)`, ~L1528 `setMaskTheme(next)`; shifted by #762 on `main`) is the swap site. Inside it, register `map.once('style.load', …)` (or guard on `styledata` + `map.isStyleLoaded()`) to re-run **the label-isolation half only** (3a). The float/sink half (3b) lives in the separate `maskPolygon`-watching `mapReady`-gated effect. Do the initial application the same way: run isolation in the existing `handleLoad`/`onLoad` path and the float/sink in the `maskPolygon` effect. Mirror the existing `mapReady`-gated effect pattern and the `prevThemeRef` same-value guard (prototype ~L1496–1505) so a no-op theme write doesn't thrash the layers.

- **Teardown contract:** on scope → us/chooser (mask unmounts) and on unmount, call `restoreLabelIsolation` + `removeFloatLayers`. Wrap MapLibre calls in `try/catch` (or guard with `map.getLayer(id)`/`map.getStyle()`) — the existing code already guards `getLayer` calls because layers/the map can be gone after a style swap or disposal. Match that defensive style.

- **Float-layer ids** must be stable, app-owned ids (`state-artboard-halo`, `state-artboard-outline`) so re-apply can `getLayer`-guard and `removeLayer` them idempotently across swaps.

### Buffered within vs. exact fill (label-survival — the one piece NOT caught by unit/e2e)

MapLibre's `['within', geom]` returns **FALSE for a point ON or OUTSIDE the boundary**. The `maskPolygon` is the 5%-mapshaper-simplified state set (`data/us-state-polygons.geojson`, generated by `scripts/generate-state-boundaries.mjs` with `-simplify 5% keep-shapes visvalingam`), whose simplified edge does NOT coincide with the basemap's native label anchors. So a legitimate **INTERIOR city near the border** (e.g. Yuma, AZ; small border towns) can fall on the wrong side of the simplified edge, evaluate `within → false`, and VANISH — directly contradicting the "interior labels render whole" AC.

This failure is **invisible to the test suite**: the deterministic unit test uses mock geometry (it asserts the filter *shape*, `['within', …]` / `['all', …, ['within', …]]`, not whether a real near-border anchor passes), and the headless e2e skips label-bleed entirely (#764). It surfaces ONLY in the manual 5×2 design review.

Fix: feed `within` a slightly **OUTWARD-BUFFERED** state polygon so near-border interior anchors land inside the buffered ring and survive `within → true`. Buffer width: **~8 km via `@turf/buffer`** (turf distance is **kilometers**, NOT degrees — `0.05–0.1°` is the wrong unit and silently under-buffers); the dependency-free fallback (per-ring outward nudge / bbox-pad) uses **~`0.1°`**. Tunable, but wide enough to clear the 5%-simplified edge. The mask **FILL keeps the EXACT (unbuffered) simplified `maskPolygon`** so the gray edge still matches the `ST_Intersects` data-clip against `state_boundaries` (#762/#733) — buffering the fill would desync the visible gray edge from the actual clipped data. Two polygons, two jobs: buffered for the label `within` test, exact for the fill. **Tradeoff:** an outward buffer symmetrically re-admits FOREIGN-state labels within the band, partly fighting "exterior labels gone" — pick the smallest width that saves interior near-border anchors without re-admitting across-the-border labels (verify BOTH directions, see AC).

### Crisp outline as the a11y boundary (≥3:1 non-text contrast)

The mask fill alone fails WCAG 1.4.11 (≈1.05:1 dark / ≈1.26:1 light against the adjacent state land). The crisp `state-artboard-outline` line is therefore the load-bearing boundary, not a cosmetic flourish. Give it an explicit **≥3:1 non-text contrast target against BOTH (i) the mask fill (`#06090e` dark / `#d8d8d8` light — the v3-corrected colors from #762, NOT the prototype's `#161b27`/`#e7e2d6`) and (ii) the adjacent in-state land** in both themes, and verify the computed ratios. Coordinate the edge-visibility seam with #762: either #762's AC acknowledges that its fill-only merged state has an edge-invisible boundary until SUB2 lands, OR #762 folds a minimal 1px high-contrast outline so no merged commit on `main` ever has an invisible boundary. Note in the PR which path was taken.

### Unit-mock extension (required — current mock lacks these methods)

The shared MapCanvas test mock-map factory (prototype `MapCanvas.test.tsx` ~L55–115; on `main` after #762) defines `getLayer`, `addLayer`, `removeLayer`, `setStyle`, `getSource`, `on/off/once` etc. as `vi.fn()`, but it has **no** `getStyle`, `getFilter`, `setFilter`, `moveLayer`, `triggerRepaint`, and `getLayer` returns `undefined` by default. SUB2 must extend the mock:
- `getStyle: vi.fn(() => ({ layers: [ … ] }))` returning a representative layer list mixing `symbol` layers (some matching the name heuristic, some not — to assert the heuristic is selective and fails open) and `fill`/`line` basemap layers (to assert sinking).
- `getLayer: vi.fn((id) => layersById[id])` returning `{ id, type, filter }` so capture-of-original works (one prototype test ~L1409 does `getLayer.mockReturnValue({ id })`; generalize it).
- `getFilter: vi.fn((id) => layersById[id]?.filter)` so original-filter capture is testable.
- `setFilter: vi.fn()`, `moveLayer: vi.fn()`, `triggerRepaint: vi.fn()` spies to assert the merged filter expression (including that its `within` geometry is the BUFFERED `isolationPolygon`, distinct from the fill's exact `maskPolygon`), the restore-to-original call, the `moveLayer(strayId, 'state-mask-fill')` order, and the repaint flush.
- `addLayer` already exists — assert it's called with the halo/outline specs and (where order matters) the `beforeId`/insert-position arg.

Keep these additions in the shared factory so #762's existing assertions still pass; only ADD methods, don't change existing return shapes.

## Acceptance criteria

- [ ] **Reconcile-sequencing split is implemented and documented:** label-isolation re-applies in `style.load`; float `addLayer` + `moveLayer` stray-sink re-apply from a separate `maskPolygon`-watching `mapReady`-gated effect; `moveLayer` is guarded by `map.getLayer('state-mask-fill')` (warn-and-return if absent, never call through). A sequencing comment prevents future collapse. Unit-asserted: a simulated `setStyle` + `style.load` re-invokes `setFilter` (not `moveLayer`); the `maskPolygon` effect is where `moveLayer`/`addLayer` fire; `moveLayer` is NOT called when the mock's `getLayer('state-mask-fill')` returns `undefined`.
- [ ] Symbol layers are matched by type + name heuristic against `getStyle().layers` — no hardcoded layer ids; the heuristic fails OPEN. Unit-asserted with a mock layer list containing matching + non-matching + non-symbol layers.
- [ ] Each touched symbol layer's ORIGINAL filter is captured before merge and RESTORED on mask unmount; whole-US (`?scope=us`) and chooser views render with no `['within', …]` filter applied. Unit-asserted: `setFilter` called with merged expr on apply, with the captured original on teardown.
- [ ] Merged filter is `['within', isolationPolygon]` when no original exists, else `['all', original, ['within', isolationPolygon]]`. **Deterministic unit test asserts `applyLabelIsolation` produces `within`-shaped filters against a mock `getStyle().layers` on both branches** — this is where the label-bleed regression guard actually lives (the headless e2e skips, see #764).
- [ ] **The `within` label test uses the OUTWARD-BUFFERED polygon (`isolationPolygon`) while the mask FILL uses the EXACT (unbuffered) `maskPolygon`.** Unit-asserted: `applyLabelIsolation` is given a buffered geometry distinct from the fill's `maskPolygon` (e.g. assert the within-geometry's bbox is strictly larger than the fill polygon's, or that `bufferIsolationPolygon` is invoked); the #762 fill layer's geometry is unchanged.
- [ ] **Near-border INTERIOR labels survive isolation, AND across-the-border FOREIGN labels stay gone** — the manual design-review check pairs BOTH directions: named near-border interior cities (e.g. Yuma, AZ; ≥1 other border town) still render under an active state scope (NOT dropped by `within`), AND a just-across-the-border foreign-state label (e.g. a CA/NV town just over the AZ line) is confirmed ABSENT — the outward buffer must not be wide enough to re-admit foreign labels. (Manual-only: NOT catchable by the unit/e2e suite — see "Buffered within vs. exact fill"; it MUST be verified in the manual 5×2 design review.)
- [ ] `map.triggerRepaint()` is called after `setFilter` in both `applyLabelIsolation` and `restoreLabelIsolation`, each with a why-comment framing it as a defensive idle-map flush (since `setFilter` already schedules a render), not as load-bearing. Unit-asserted on the `triggerRepaint` spy.
- [ ] Stray basemap `fill`/`line` layers ordered above `state-mask-fill` are moved beneath it via `moveLayer(strayId, 'state-mask-fill')` (no boundary/coastline bleed onto the gray). Unit-asserted on `moveLayer` calls.
- [ ] Halo (`line` + `line-blur`) and crisp outline (`line`) float layers mount above the mask, are theme-aware (drop-shadow in light / glow in dark), and are removed on unmount; stable app-owned ids; `removeLayer` is idempotent + guarded. Unit-asserted on `addLayer`/`removeLayer`.
- [ ] **The crisp outline meets ≥3:1 non-text contrast against both the mask fill and the adjacent in-state land, in both themes** — computed ratios recorded in the PR; the edge-invisibility seam with #762 is resolved (either #762 acknowledges fill-only edge-invisibility or folds a 1px boundary).
- [ ] No new ORPHAN exports in `artboard-layers.ts` — every export (function or type) has a consumer in this PR; knip clean. (Whether `SavedFilters` is exported or kept a private alias via `ReturnType<typeof applyLabelIsolation>` is implementer's choice — both are knip-clean since a cross-file-consumed export is "used"; this is not a CI gate.)
- [ ] All teardown paths guard against a disposed map / missing layers (try/catch or `getLayer`/`getStyle` guard) — no throw when the map is gone after a swap or unmount.
- [ ] The mask-fill, camera/clamp, `renderWorldCopies`, scope-announcement live region, and asset behavior from #762 are unchanged (no regression in #762's MapCanvas/mask tests).

## Dependencies (sequence)

- **Depends on #762** (artboard core) being merged to `main` first — SUB2 targets the `state-mask`/`state-mask-fill` layer and the `maskPolygon`/MutationObserver/theme-swap wiring that #762 lands. Rebase SUB2 onto the advancing `main` after #762 merges.
- **Blocks #764** (verification) — #764's e2e label-bleed checks, axe scan, and the design-review gate run against the assembled #762+SUB2 feature.
- **MapCanvas contention:** #762, SUB2, #764 all touch `frontend/src/components/map/MapCanvas.tsx`. They MUST be implemented strictly sequentially (#762 → SUB2 → #764), each rebased onto the prior's merged HEAD. Do NOT run a second MapCanvas implementer concurrently.

## CI / console / design-review gates

- **CI 4-check gate green on this PR's own HEAD** (`test`, `lint`, `build`, `e2e`) — no "leave it red, fix in #764" seam. Plus `knip (informational)`, which is a REQUIRED check that blocks Mergify: confirm zero new ORPHAN (truly unused) exports in `artboard-layers.ts` before posting `@Mergifyio queue`. Note: exporting `SavedFilters` does NOT trip knip if `MapCanvas.tsx` consumes it (a cross-file-used export is "used"); the only knip risk is a helper export with no consumer in this PR.
- **If `@turf/buffer` is added** for the outward-buffered `isolationPolygon`, it touches `frontend/package.json` + the root lockfile — the `lockfile-consistency` workflow fires on the PR (catches `package-lock.json` drift). Run `npm install` so the lockfile is regenerated in the same commit, and confirm `lockfile-consistency` is green. (If the cheap fallback nudge/bbox-pad is used instead, no dependency is added and this gate is moot.)
- **Frontend UI PR — clean-console gate applies.** The positron POI sprite warning `Image "circle-11" could not be loaded` must be handled via SCOPED EXACT-MATCH suppression (never a blanket console filter); #764 owns the formal console-capture spec, but SUB2's own live Playwright pass must not surface NEW console errors/warnings beyond that one known sprite warning — and in particular must not surface any `Cannot move layer before non-existing layer` throw (the sequencing blocker).
- **Live Playwright verification (implementer, this PR):** drive `?state=US-AZ` (and at least one small state, e.g. `US-RI`) — confirm exterior labels are gone, interior labels render whole and un-sliced, **named near-border INTERIOR cities survive (e.g. Yuma, AZ — the buffered-`within` check; this is the one failure NOT caught by unit/e2e, so verify it by eye here and in the manual 5×2 design review)**, no coastline/boundary bleed onto the gray, theme toggle does NOT throw, and the halo/crisp-outline float reads with ≥3:1 edge contrast in BOTH themes (toggle `data-theme` via `browser_evaluate`). The formal 5-viewport × 2-theme ui-design design-review gate is #764's, but the risky imperative work here warrants its own focused **opus** review.
- **PR body:** `.github/PULL_REQUEST_TEMPLATE.md` verbatim (all 5 sections). Screenshots REQUIRED (frontend visible-UI change) via the `pr-screenshots-via-user-attachments` skill (`user-attachments/assets/<uuid>`; never commit PNGs). Bot review dispatches through the `julianken-bot` Agent subagent (opus), never `gh pr review`.

## References

- Epic **#760** (state-artboard mask) — see sibling sub-issues #762 (core) and #764 (verification).
- Prototype branch `feat/state-artboard-mask-760` (mined reference, NOT merged) — commits `0526ebb` (generator), `c85a2a7` (mask.ts util), `3cef141` (hook), `1198637` (mask render + clamp decouple), `61315d4` (MapSurface/App wiring). Mine the integration surface; treat it as proven brainstorming code, not done.
- `frontend/src/components/map/mask.ts` — `buildMaskFeature` (returns `Feature<Polygon>`), `padBounds`, `ARTBOARD_PAD`, `MASK_FILL_LIGHT = '#d8d8d8'` / `MASK_FILL_DARK = '#06090e'` (#762).
- `data/us-state-polygons.geojson` + `scripts/generate-state-boundaries.mjs` (`-simplify 5% keep-shapes visvalingam`) — the 5%-simplified source of `maskPolygon`. Its simplified edge is why `within` must use a buffered copy for labels (see "Buffered within vs. exact fill") while the fill keeps the exact polygon to stay aligned with the `ST_Intersects` clip against `state_boundaries` (#733). `@turf/buffer` is NOT currently a dependency (only `@turf/boolean-point-in-polygon` is) — adding it triggers `lockfile-consistency`.
- `frontend/src/components/map/MapCanvas.tsx` — theme-swap MutationObserver (prototype ~L1511–1528: `setStyle` then `setMaskTheme`), `prevThemeRef` same-value guard (~L1496–1505), defensive `getLayer` guard pattern, canonical `state-mask` source / `state-mask-fill` layer render (#762, prototype ~L1847–1852).
- `frontend/src/components/map/MapCanvas.test.tsx` — shared mock-map factory to extend with `getStyle`/`getFilter`/`setFilter`/`moveLayer`/`triggerRepaint` (prototype ~L55–115; `getLayer.mockReturnValue` example ~L1409).
- `knip.ts:134` — `ignoreExportsUsedInFile: true` is *same-file* suppression only; it does NOT govern cross-file consumption, so it is not a reason `SavedFilters` must stay private. Keeping the alias private is a style choice (`ReturnType<typeof applyLabelIsolation>` at the call site), not a knip/Mergify requirement.
- Risk register (epic holistic findings): **moveLayer/style.load reconcile-sequencing** (dominant blocker — the 3a/3b split above); **LABEL BLEED RISK** — heuristic must fail open, guarded by the deterministic unit test; **NEAR-BORDER INTERIOR LABEL DROP** — `within` is FALSE on/outside the simplified edge, so labels near the border vanish unless `within` uses the outward-buffered polygon (fill stays exact); uncatchable by unit (mock geometry) or headless e2e — manual 5×2 design-review only; **a11y non-text contrast** — the crisp outline carries the boundary; **renderWorldCopies:false trap** — a future scope-change remount (note the #761 always-mounted-map re-architecture) could drop both `renderWorldCopies:false` and these re-applied layers (caught by #764 design review). Pull fresh `maplibre-gl` 5.x docs from context7 for `within`-expression, `setFilter`, `moveLayer`, `triggerRepaint`, and `style.load` semantics before writing the imperative code (5.x differs from 4.x).

julianken:
## Diagrams

```mermaid
flowchart TD
    subgraph stateScope["state scope active (maskPolygon set)"]
        SL["map 'style.load'<br/>(fires on each theme setStyle swap)"]
        EFF["maskPolygon-watching<br/>mapReady-gated useEffect<br/>(runs AFTER react-map-gl re-adds state-mask-fill)"]
    end

    SL -->|"3a: label isolation only"| ALI["applyLabelIsolation(map, bufferIsolationPolygon(poly))<br/>merge ['within', isolationPolygon] into each<br/>basemap symbol layer; capture originals"]
    EFF -->|"3b: float + sink<br/>(getLayer-guarded on state-mask-fill)"| AAF["applyArtboardFidelity"]
    AAF --> SINK["sinkStrayLayersBelowMask<br/>moveLayer(stray, 'state-mask-fill')"]
    AAF --> FLOAT["addFloatLayers<br/>halo (line-blur) + crisp outline (line)<br/>via ONE named source state-artboard-line"]

    ALI -.->|"teardown on scope→us/chooser"| REST["restoreLabelIsolation + removeFloatLayers"]

    BUF["maskPolygon (5%-simplified)"] -->|"OUTWARD buffer 0.2 km"| ISO["isolationPolygon<br/>(label 'within' test ONLY)"]
    BUF -->|"EXACT, unbuffered"| FILL["#762 state-mask-fill geometry<br/>(stays aligned with ST_Intersects clip)"]
```

## Summary

- **Why:** #762 lands the inverse-mask fill but the basemap symbol layers still render worldwide, so OTHER-state labels bleed onto the gray and labels straddling the state border get visually SLICED by the opaque fill. On prod this surfaced as labels (Arlington, Alexandria, "Virginia", the river label) clipped at the artboard edge — worst on tiny `US-DC`. SUB2 makes the artboard look *finished*: exterior labels REMOVED (not clipped), interior labels whole, no boundary/coastline bleed, plus a theme-aware halo + crisp outline so the state floats.
- **How:** A new unit-testable helper `artboard-layers.ts` merges `['within', isolationPolygon]` into each basemap symbol layer (matched by a fail-open type+name heuristic, never hardcoded ids), sinks stray fill/line layers below the mask, and adds the float layers. The `within` test uses an **outward-buffered** polygon (calibrated **0.2 km** via `@turf/buffer`) while the fill keeps the **exact** `maskPolygon` — two polygons, two jobs — so near-border interior anchors (Yuma) survive while across-the-border foreign cities (Heroica Nogales, MX) stay gone. The reconcile-sequencing blocker is handled by SPLITTING re-apply: label isolation in `style.load`; float + stray-sink in a separate `maskPolygon`-watching `mapReady`-gated effect guarded by `getLayer('state-mask-fill')`.
- **a11y / robustness:** the crisp outline is the WCAG 1.4.11 load-bearing boundary with an explicit **≥3:1** non-text-contrast target against BOTH the mask fill and the adjacent land, in both themes (light outline `#1a1d24`: ≈12.3:1 vs fill / ≈16.3:1 vs land; dark outline `#e8edf4`: ≈16.4:1 / ≈15.6:1). Both float layers share ONE explicit **named** source (`state-artboard-line`) so rapid theme swaps never orphan an anonymous source — fixing the maplibre `coalesceChanges` "Cannot convert undefined or null to object" render-time throw. **This FIXES the cover-mode label clipping the user reported.**

## Screenshots

**`?state=US-AZ`** — 5 canonical viewports × light + dark (10 captures). Crisp AZ outline + halo float on the gray; interior labels whole (Phoenix/Tucson/Flagstaff/Yuma); zero Mexican-border or coastline bleed.

| Light | Dark |
|---|---|
| az-390-light | az-390-dark |  
| az-768-light | az-768-dark |  
| az-1024-light | az-1024-dark |  
| az-1440-light | az-1440-dark |  
| az-1920-light | az-1920-dark |  

**`?state=US-DC`** — the user-flagged worst case (tiny district surrounded by VA/MD). Diamond outline whole; the only labels rendered are DC-interior (Washington + two interior freeways); exterior VA/MD labels (Arlington, Alexandria, "Virginia", the river label) are REMOVED, not clipped. Both themes.

| Light | Dark |
|---|---|
| dc-1440-light | dc-1440-dark |

## Test plan

- [x] `npm run typecheck && npm run test` — green (frontend: 69 files, 1022 tests pass; shared-types built first)
- [x] New unit / integration tests added — `artboard-layers.test.ts` (415 lines) + extended `MapCanvas.test.tsx` mock-map factory (`getStyle`/`getFilter`/`setFilter`/`moveLayer`/`triggerRepaint`)
- [x] New Playwright e2e spec added (if user-visible behavior changed) — N/A here; the formal e2e label-bleed + axe specs are owned by sibling #764 (this is SUB2's implementer pass)
- [x] `npm run build` — clean production build; `npm run lint` 0 errors; `npx knip` no new findings; `npx tsc -p frontend/tsconfig.test.json` 109 (baseline)
- [x] (UI only) Playwright MCP smoke — ran `npm run dev` against a local mock read-api on :8787; drove `?state=US-DC` (flagged worst case), `?state=US-AZ`, `?state=US-RI` at all 5 canonical viewports × light+dark. **DC:** only 3 labels render and ALL are interior (Washington, Anacostia Freeway, Southeast Freeway) — zero foreign labels (Arlington/Alexandria/Virginia REMOVED, not clipped); diamond outline whole; no edge-slice. **AZ:** Yuma (lat 32.72) + Nogales **AZ** (lat 31.34, NORTH of the Sonora line) survive; **zero** place features render south of lat 31.3322 — Heroica Nogales MX (lat 31.329) ABSENT; no coastline bleed. **RI:** Providence/Newport whole; MA/CT cities absent. 5 rapid `data-theme` swaps left all artboard layers intact + recolored with **NO** `coalesceChanges` throw and **NO** `Cannot move layer` throw. Console: 0 errors. The only warnings are missing-positron-sprite warnings (`circle-11` per #764, and `wood-pattern` — the same basemap-sprite class, surfaced by the DC landcover tiles; neither references any artboard layer and neither is introduced by this PR).

## Plan reference

Part of epic **#760** (state-artboard mask), SUB2 of 3 (after #762 core, before #764 verification). See `docs/plans/2026-05-29-state-artboard-mask.md`.

Closes #763

---

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -14,6 +14,7 @@
   "dependencies": {
     "@bird-watch/shared-types": "*",
     "@microsoft/clarity": "^1.0.2",
+    "@turf/buffer": "^7.3.5",
     "maplibre-gl": "^5.24.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
--- a/frontend/src/components/map/MapCanvas.test.tsx
+++ b/frontend/src/components/map/MapCanvas.test.tsx
@@ -60,7 +60,44 @@ function makeFakeMap() {
   // stateful pair (maplibre's default is `true`). Lets tests both exercise the
   // effect without throwing AND assert the imperative value it lands on.
   let renderWorldCopiesState = true;
+  // #763 — a representative style.layers list for the artboard-fidelity
+  // imperative work (label isolation, sink, float). Mixes symbol layers (some
+  // matching the place/label heuristic, some NOT — to assert the heuristic is
+  // selective and fails open) with basemap fill/line layers (to assert
+  // sinking) and the #762 mask fill (the z-order anchor). The state-mask-fill
+  // entry is conditionally present so a test can simulate the
+  // reconcile-sequencing window where the mask layer does NOT yet exist (the
+  // moveLayer guard).
+  let styleHasMaskLayer = true;
+  const baseStyleLayers = () => {
+    const layers: Array<{ id: string; type: string; filter?: unknown }> = [
+      { id: 'background', type: 'background' },
+      { id: 'water', type: 'fill' },
+      { id: 'place_country', type: 'symbol', filter: ['==', 'class', 'country'] },
+      { id: 'place_city', type: 'symbol' },
+      { id: 'poi_z14', type: 'symbol' },
+      { id: 'transit_route_ref', type: 'symbol' }, // symbol, no place/label token
+    ];
+    if (styleHasMaskLayer) layers.push({ id: 'state-mask-fill', type: 'fill' });
+    // Stray basemap line/fill layers painted ABOVE the mask (sink targets).
+    layers.push({ id: 'boundary_country', type: 'line' });
+    layers.push({ id: 'landcover_glacier', type: 'fill' });
+    return layers;
+  };
+  let styleLayers = baseStyleLayers();
+  const layersById = () =>
+    Object.fromEntries(styleLayers.map((l) => [l.id, l]));
   return {
+    // #763 test affordances (not part of the maplibre API surface): let a test
+    // simulate the reconcile window where state-mask-fill is absent, and reset
+    // the style layer list between style swaps.
+    __setMaskLayerPresent: (present: boolean) => {
+      styleHasMaskLayer = present;
+      styleLayers = baseStyleLayers();
+    },
+    __resetStyleLayers: () => {
+      styleLayers = baseStyleLayers();
+    },
     on: vi.fn(
       (
         event: string,
@@ -85,7 +122,19 @@ function makeFakeMap() {
     queryRenderedFeatures: vi.fn(),
     querySourceFeatures: vi.fn(() => []),
     getSource: vi.fn(),
-    getLayer: vi.fn(),
+    // #763 — getLayer resolves against the in-memory style layer list so the
+    // moveLayer/float guards and original-filter capture are testable. (The
+    // #762 cluster tests that call getLayer.mockReturnValue still override it.)
+    getLayer: vi.fn((id?: string) => (id ? layersById()[id] : undefined)),
+    // #763 — style introspection + imperative layer ops for artboard fidelity.
+    getStyle: vi.fn(() => ({ layers: styleLayers })),
+    getFilter: vi.fn((id: string) => layersById()[id]?.filter),
+    setFilter: vi.fn((id: string, filter: unknown) => {
+      const layer = layersById()[id];
+      if (layer) layer.filter = filter;
+    }),
+    moveLayer: vi.fn(),
+    triggerRepaint: vi.fn(),
     getCanvas: vi.fn(() => canvas),
     getContainer: vi.fn(() => container),
     easeTo: vi.fn(),
@@ -121,8 +170,17 @@ function makeFakeMap() {
     unproject: vi.fn(() => [-111, 34]),
     addSource: vi.fn(),
     removeSource: vi.fn(),
-    addLayer: vi.fn(),
-    removeLayer: vi.fn(),
+    // #763 — addLayer/removeLayer mutate the in-memory style layer list so the
+    // float-layer lifecycle is realistic: an imperatively-added halo/outline is
+    // then visible to getLayer, and removeFloatLayers can guard-and-remove it.
+    addLayer: vi.fn((layer: { id?: string; type?: string }) => {
+      if (layer?.id && !styleLayers.some((l) => l.id === layer.id)) {
+        styleLayers.push({ id: layer.id, type: layer.type ?? 'line' });
+      }
+    }),
+    removeLayer: vi.fn((id: string) => {
+      styleLayers = styleLayers.filter((l) => l.id !== id);
+    }),
     addImage: vi.fn((id: string) => {
       sprites.add(id);
     }),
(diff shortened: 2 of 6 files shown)


julianken:
@Mergifyio queue

Issue #761, opened 2026-05-29
feat(frontend): map-first re-architecture — full-viewport map canvas with floating overlay UI


julianken:
## Problem

Today the map is *windowed inside the UI*, not the application itself. The app shell (`frontend/src/App.tsx`) is a vertical flex column: `.app { display: flex; flex-direction: column; height: 100vh }` (`styles.css:77-81`) with a sticky `AppHeader` banner (`.app-header { position: sticky; top: 0; z-index: 10 }`, `styles.css:164-174`) above a single scrollable `<main id="main-surface">` that owns `flex: 1; min-height: 0; padding: var(--space-lg); overflow-y: auto; display: flex; flex-direction: column` (`styles.css:86-93`).

The chrome — not the map — is the structural frame. View switching inside `<main>` swaps between mutually-exclusive surfaces driven entirely by the URL `?view=` parameter via the custom `useUrlState` hook (`frontend/src/state/url-state.ts`); there is no client-side router library. `MapSurface` (`frontend/src/components/MapSurface.tsx`) renders only when the view is map/detail, hosting the lazy-loaded `MapCanvas` (`React.lazy()`, `MapSurface.tsx:16-18`), a `.map-context-strip` section (MapLede + FilterSentence + freshness), and a `FamilyLegend` overlay.

The result is a **content-model architecture** — "surfaces inside containers" — and the map feels constrained because:

1. `#main-surface` adds 16px padding on all four sides around every surface (`styles.css:89`).
2. The `.map-context-strip` stacks *above* the canvas as a sibling, consuming ~85–110px of vertical height before the map even begins (`MapSurface.tsx:249-267`, `styles.css:1228-1232`, `1212-1219`).
3. `.map-surface { position: relative; width: 100%; flex: 1; min-height: 0 }` (`styles.css:686-691`) makes the canvas a controlled flex child capped by the padded parent — the inner `[data-testid=map-canvas]` and `MapView` are `width:100% / height:100%` of *that* constrained box, not of the viewport.

[...]

Issue #740, opened 2026-05-29
feat(frontend): wire scope end-to-end in App.tsx


julianken:
**Goal:** Wire scope end-to-end in `frontend/src/App.tsx` (Task C6). Render the landing chooser (#742) when **unscoped** and **suppress the cold-load CONUS `/api/observations` fetch + the map render until a scope is chosen**; then route `?state=US-XX`→clip filter + scoped camera, ZIP `onResolve`→state scope + flyTo-within-state, `?scope=us`→CONUS map. This is the integration task — it owns the render-gating and fetch-gating that #742 (chooser UI), #736 (camera), #737 (in-state control), and #738 (runtime label) only supply pieces of.

**Files:** `frontend/src/App.tsx` (+ `frontend/src/App.test.tsx`)

**Depends on:** #735 (URL three-state model + exported `DEFAULTS`), #736 (controllable `MapCanvas` camera — `fitBounds`/`flyTo`/dynamic `MAX_BOUNDS`), #737 (in-state `ScopeControl`), #738 (`regionLabelFor(scope)`), #739 (`ZipInput` + `ScopeResolution` + `ZIP_FLYTO_ZOOM`), #742 (`ScopeChooser`). All seven must merge before this one — C6 is the join point.

---

### Context: what App.tsx does today (the baseline to change)

`App.tsx` today (`b640fee`) fires `useBirdData(apiClient, {…, bbox: debouncedBbox, zoom: debouncedZoom})` **unconditionally on mount** with `DEFAULT_BBOX_CONUS = [-125, 24, -66, 50]` at `zoom: 3` (the eager cold-load this issue suppresses), and renders `<MapSurface …>` whenever `mapVisible` (`state.view === 'map' || 'detail'`). The viewport-idle refetch loop is `onViewportChange` (App.tsx:185–223) → a 250ms `bboxDebounceRef` debounce → `setDebouncedBbox`/`setDebouncedZoom`. **That debounce is wired to map pan/zoom (`idle`) events only** — a `stateCode`/scope change is a discrete URL-state transition, not a viewport event, so the existing debounce does **not** coalesce it. (This is the gap the bot flagged twice on the shell; the mechanism is specified concretely below, drawn from prototype finding (d).)

### The model (chooser-first REMOUNT — locked decision 4b)

[...]

PR #790, opened 2026-05-30
perf(frontend): scope-gated maplibre prefetch (lazy split preserved) — closes #781

### LINKED ISSUE #781: perf(frontend): scope-gated maplibre prefetch (lazy split preserved)
**Goal:** Warm the `maplibre-gl` + `MapCanvas` chunk *before* it is needed, but only once a scope is known — fire a low-priority `import()` (or `<link rel="modulepreload">`) of the lazy MapCanvas chunk on a **scoped landing** and on a **chooser scope-pick**, and **NEVER** on the unscoped/chooser landing. This keeps the deliberately fetch-light chooser landing untouched (locked decision #4 / #740/C6) while removing the cold-chunk-fetch latency between "user picks a scope" and "map paints" on the path where the map is actually about to render. Additive, no behavior change to data/URL/caching — can ship standalone.

## Problem

`MapCanvas` is code-split behind a `React.lazy()` boundary that lives in `MapSurface`, not in `MapCanvas` itself:

```ts
// frontend/src/components/MapSurface.tsx:16-18
const MapCanvas = React.lazy(() =>
  import('./map/MapCanvas.js').then((m) => ({ default: m.MapCanvas })),
);
```

This is the correct split and **must be kept** (OQ5 in the report) — the `maplibre-gl` chunk is **1,027.57 kB raw / 272.93 kB gzip** (measured in `dist/assets/maplibre-gl-*.js`), so eager-inlining it onto the index chunk would regress every cold start, including the unscoped chooser landing where the map is not even rendered. The split also keeps the index chunk maplibre-free (verified: `index.js` contains zero maplibre refs).

The `React.lazy` boundary does mean the chunk fetch only *starts* when `<MapSurface>` first mounts — i.e. after the user has already chosen a scope. On a scoped landing (a bookmarked `?state=US-XX` / `?scope=us` URL) the chunk fetch is on the critical path between first paint and map paint; on a chooser→scope-pick interaction it is on the critical path between the click and the map paint. Nothing today warms the chunk earlier. There is **no existing prefetch / modulepreload / `requestIdleCallback` machinery** in the frontend — the only dynamic imports are this lazy `MapCanvas` (`MapSurface.tsx:17`) and the dev-only `DsPreview` shim (`main.tsx:18`). The `index.html` head has `preconnect` hints to the API + tile host but no module preload.

The fetch gate that distinguishes "scope known" from "unscoped chooser" already exists and is the load-bearing half of the #740/C6 zero-`/api/observations`-on-landing guarantee:

```ts
// frontend/src/App.tsx:134
const scopeActive = state.scope.kind !== 'unscoped';
```

and the three scope-pick callbacks are concrete firing points already in `App.tsx`:

- `onPickState` (`App.tsx:418-424`)
- `onPickWholeUs` (`App.tsx:426-429`)
- `onResolveZip` (`App.tsx:431-446`)

with the unscoped early-return (the landing we must NOT prefetch on) at `App.tsx:541-552`.

A stale comment compounds the confusion: `MapSurface.tsx:13` claims the chunk is "217 KB" — the real cost is ~4.7× that raw / ~1.25× gzip. This issue corrects it on the same branch so future readers size the prefetch decision against reality.

## Scope

Exact files to add/modify:

1. **`frontend/src/prefetch.ts` (NEW).** A tiny, side-effect-light prefetch helper that warms the MapCanvas chunk exactly once, idempotently, and only when called:
   - Export `prefetchMapCanvas()`. It must reference the **same module specifier** as the lazy boundary (`'./map/MapCanvas.js'` — co-locate so the bundler maps both to the same chunk; if the path resolves differently from `prefetch.ts`, use the correct relative path to `components/map/MapCanvas.js` and verify the built `dist/` chunk graph dedupes to one chunk, not two).
   - Guard with a module-level `let warmed = false` so repeated calls (re-renders, multiple scope picks) issue the underlying `import()` at most once.
   - Schedule the `import()` at low priority: prefer `requestIdleCallback(() => { import(...) }, { timeout: 2000 })` with a `setTimeout(fn, 1)` fallback when `requestIdleCallback` is undefined (Safari/jsdom). The `import()` promise's rejection must be swallowed (`.catch(() => {})`) so a transient prefetch failure never surfaces as an unhandled rejection — the real render path still has its own Suspense/ErrorBoundary.
   - Guard `typeof window === 'undefined'` so it is a strict no-op under SSR/jsdom test environments that lack the scheduling primitives.

2. **`frontend/src/App.tsx` (MODIFY).** Wire `prefetchMapCanvas()` at the two scoped trigger points, never the unscoped one:
   - One `useEffect` keyed on `scopeActive` (`App.tsx:134`) that calls `prefetchMapCanvas()` when `scopeActive === true` — this covers the **scoped landing** (a `?state=`/`?scope=us` deep link mounts with `scopeActive` already true). The effect early-returns when `scopeActive === false`, so the unscoped chooser landing never warms the chunk.
   - Call `prefetchMapCanvas()` inside `onPickState` (`App.tsx:418`), `onPickWholeUs` (`App.tsx:426`), and `onResolveZip` (`App.tsx:431`) — this covers the **chooser scope-pick** path, warming the chunk on the click rather than waiting for `<MapSurface>` to mount. (The `scopeActive` effect also fires on the resulting state change, but the idempotency guard makes the double-call a no-op; calling in the handler shaves the extra render-commit latency.) Because `prefetchMapCanvas` is referenced by these `useCallback`s, the import is a stable module-level function (no new dep-array churn).

3. **`frontend/src/components/MapSurface.tsx:11-14` (MODIFY — comment only).** Correct the stale "217 KB" figure to the measured cost (1,028 kB raw / 273 kB gzip) and note that the chunk is warmed early by `prefetchMapCanvas()` on scoped paths while the lazy boundary remains the actual mount-time loader. No code change to the `React.lazy` boundary itself.

Do **NOT**:
- Eager-inline maplibre (no static `import 'maplibre-gl'` on the index path).
- Add `manualChunks` / `build.rolldownOptions.output` to `vite.config.ts` — the default Vite/Rolldown split is already clean (index chunk has zero maplibre refs); explicit chunking is unnecessary and risks splitting the prefetch target away from the lazy target.
- Add a static `<link rel="modulepreload">` to `frontend/index.html` — a static preload fires on **every** landing including the unscoped chooser, violating the locked decision. The prefetch must be runtime-conditional on scope. (If a `<link rel="modulepreload">` is preferred over `import()` for any reason, it must be injected at runtime by `prefetchMapCanvas()` via `document.head.appendChild`, gated identically — but the `import()` approach is simpler and is the recommended implementation.)

## Out of scope

- **Scope-gating `onViewportChange`** so a persistent unscoped map fires zero `/api/observations` — that is a *separate* sub-issue (epic-relative O→S4 / WS8.2, R1) and depends on the shell inversion landing first. O9 does not make the map persistent; it only warms the chunk earlier. The fetch gate (`scopeActive`/`enabled`) is untouched here.
- The shell inversion / always-mounted map (epic S1/S2) — O9 ships against the *current* windowed shell where `<MapSurface>` still mounts on scope-pick. It will continue to work unchanged once the map becomes persistent.
- Recording the `dist/` chunk-size baseline as a tracked AC artifact — that is the separate WS8.4 sub-issue (P3). O9 only corrects the inline comment; it does not add a size-budget gate.
- Memoizing relocated persistent overlays / theme-toggle render-count guards (WS8.3 / O8).
- LCP / Lighthouse perf budgets — no performance-budget CI gate is added here.

## Acceptance criteria

- [ ] `prefetchMapCanvas()` exists in `frontend/src/prefetch.ts`, references the same `MapCanvas` chunk specifier as the `MapSurface.tsx:17` lazy boundary, is idempotent (warms at most once via a module-level guard), schedules at low priority (`requestIdleCallback` with a `setTimeout` fallback), swallows its own rejection, and is a strict no-op when `typeof window === 'undefined'`.
- [ ] `App.tsx` calls `prefetchMapCanvas()` on a scoped landing (a `useEffect` keyed on `scopeActive`, App.tsx:134, that early-returns when `scopeActive === false`) AND on each scope-pick (`onPickState`/`onPickWholeUs`/`onResolveZip`, App.tsx:418/426/431).
- [ ] The prefetch **never** fires on the unscoped/chooser landing: a bare-URL landing schedules **zero** network requests for the MapCanvas/maplibre chunk and **zero** `/api/observations` requests (the #740/C6 AC `expect(obsRequests).toHaveLength(0)` at `state-scope.spec.ts:113-114` and `zip-scope.spec.ts:179` stay green and unmodified).
- [ ] The `React.lazy(() => import('./map/MapCanvas.js'))` boundary at `MapSurface.tsx:16-18` is unchanged — the map still mounts via Suspense; the prefetch only warms the chunk earlier. No eager static `import 'maplibre-gl'`, no `manualChunks`/`rolldownOptions` in `vite.config.ts`, no static `<link rel="modulepreload">` in `index.html`.
- [ ] The stale "217 KB" comment at `MapSurface.tsx:13` is corrected to the measured cost (1,028 kB raw / 273 kB gzip).
- [ ] The built `dist/` graph still resolves the prefetch target and the lazy target to a **single** chunk (verify after `npm run build` that there are not two MapCanvas/maplibre chunks).
- [ ] **This PR alone leaves all four CI checks green — `test`, `lint`, `build`, `e2e` — plus `knip`. No "leave CI red, fix next PR".** If `knip` flags `prefetch.ts` as unused (it is reached only via the static `import` in `App.tsx`, which knip resolves — but the new file must not become orphaned), address it on the same branch per the repo's knip false-positive workflow rather than deleting the file. The `orphan-classname-check` is `N/A` (no new `className`s).

## Test plan

**Unit (vitest — `frontend/src/prefetch.test.ts`, NEW):**
- Mock `import('./map/MapCanvas.js')` (e.g. `vi.mock` or a dynamic-import spy) and assert `prefetchMapCanvas()` triggers exactly **one** underlying import even when called multiple times (idempotency).
- Assert calling it without `requestIdleCallback` defined (delete/stub it) falls back to `setTimeout` and still triggers the import.
- Assert a rejected import does not produce an unhandled rejection (swallowed `.catch`).

**Unit (vitest — extend `frontend/src/App.test.tsx`):**
- Spy on `prefetchMapCanvas` (mock the `./prefetch.js` module). Assert it is **NOT** called on the unscoped/chooser render (`state.scope.kind === 'unscoped'`).
- Assert it **IS** called when a scope becomes active: render with `?state=US-AZ` (scoped landing) → called; and simulate `onPickState`/`onPickWholeUs`/`onResolveZip` → called. These extend the existing `App.test.tsx` scope coverage (#740) rather than duplicating it.

**E2E (`@playwright/test` — extend `frontend/e2e/state-scope.spec.ts`):**
- Add a network-counting block (mirror the existing `obsRequests` collector at `state-scope.spec.ts:75-78`) that counts requests whose URL matches the built MapCanvas/maplibre chunk (match on `/maplibre/i` or the `MapCanvas` chunk name in the request URL). On the bare-URL chooser landing (`gotoRaw('')`, the existing case at `state-scope.spec.ts:82-114`), assert **zero** maplibre-chunk requests over an 800ms window — and keep the existing `expect(obsRequests).toHaveLength(0)` assertion green and unmodified.
- On the state-select round-trip (the existing scoped case in the same file) assert the maplibre chunk **is** fetched (count ≥ 1) — confirming the prefetch warms the chunk on the scoped path. (If the e2e dev/preview server name-hashes the chunk, match on the stable substring the bundler emits, e.g. `maplibre-gl`.)
- No new spec file is required; the assertions ride the existing scope specs. Do not change `playwright.config.ts` workers/retries. The no-DB-write grep stays clean (this PR adds no writes).

## Dependencies

Part of #761. Epic-relative blockers: **none** — O9 is additive and can ship standalone against the current windowed shell (the epic graph lists O9 as "Additive — can ship standalone"). It composes cleanly with the later shell-inversion sub-issues (S1/S2) and the separate `onViewportChange` scope-gate (S4) without depending on them.

## Plan traceability

- Research report: `docs/analyses/2026-05-29-map-first-rearchitecture-761/report.md` — §5 OQ5 ("Bundle strategy for eager maplibre" → keep React.lazy split + scope-gated prefetch, never inline, never `manualChunks`), §6 WS8 sub-issue 8.1 ("Prefetch maplibre/MapCanvas on scoped paths + fix stale comment"), §7 precondition 3 ("Prefetch, not inline"), risk **R4** (273 KB gzip cold-start regression if inlined), and §10 phase-4 item **O9**.
- Epic workstream: #761 §8 ("Performance of an always-mounted map") and open question 5 ("Bundle strategy for eager `maplibre-gl`: Suspense placeholder vs. main-chunk vs. prefetch?").

julianken:
## Diagrams

```mermaid
flowchart TD
    Landing{Landing URL}
    Landing -->|bare URL / unscoped| Chooser[ScopeChooser]
    Landing -->|"?state= / ?scope=us<br/>(scopeActive === true)"| ScopedMount[scoped mount]

    Chooser -.->|NEVER prefetch| NoWarm[(chunk stays cold)]
    Chooser -->|pick state / whole-US / ZIP| Pick[scope-pick handler]

    Pick -->|prefetchMapCanvas&#40;&#41; on click| Warm
    ScopedMount -->|useEffect keyed on scopeActive| Warm[prefetchMapCanvas&#40;&#41;]

    Warm -->|"idempotent guard<br/>requestIdleCallback / setTimeout"| Import["import('./components/map/MapCanvas.js')<br/>(same chunk as React.lazy boundary)"]
    Import --> Cache[(browser module cache warmed)]

    MapSurface["MapSurface React.lazy<br/>import('./map/MapCanvas.js')"] -->|Suspense reads warmed cache on mount| Cache
```

## Summary

- **Why:** `MapCanvas` (+ its ~1,028 kB raw / ~273 kB gzip `maplibre-gl` dep) is code-split behind the `React.lazy` boundary in `MapSurface.tsx`, which only *starts* the chunk fetch when `<MapSurface>` mounts — i.e. after the user has already chosen a scope. That puts the cold-chunk fetch on the critical path between scope intent and map paint. This warms the chunk **earlier**, but only once a scope is known.
- **What:** a tiny idempotent, low-priority `prefetchMapCanvas()` (`requestIdleCallback` + `setTimeout` fallback, rejection swallowed, SSR no-op) that references the **same** MapCanvas specifier as the lazy boundary so the bundler dedupes both to one chunk. `App.tsx` fires it on a scoped landing (`useEffect` keyed on `scopeActive`) and on each scope-pick (`onPickState`/`onPickWholeUs`/`onResolveZip`), and **NEVER** on the unscoped chooser landing — preserving the #740/C6 fetch-light landing guarantee.
- **Kept the split:** the `React.lazy(() => import('./map/MapCanvas.js'))` boundary is unchanged — no eager `maplibre` import, no `manualChunks`/`rolldownOptions`, no static `<link rel="modulepreload">`. Inlining would regress every cold start including the map-less chooser landing (risk R4). Also corrects the stale "217 KB" comment in `MapSurface.tsx` to the measured cost.

## Screenshots

N/A — not UI (additive perf/loading only; no visible change to data, URL, or rendered surfaces).

- [x] Every new/renamed `className` in this PR has a matching CSS rule — `N/A — no new classNames`
- [x] Multi-viewport design QA — `N/A — not UI`

## Test plan

- [x] `npm test --workspace @bird-watch/frontend` — green (67 files, 972 tests). New: `frontend/src/prefetch.test.ts` (5 tests: idempotency, `setTimeout` fallback, `requestIdleCallback` path, swallowed rejection, SSR no-op) + 6 wiring tests in `App.test.tsx` (NOT called unscoped; called on scoped `?state=`/`?scope=us` landing; called on each scope-pick).
- [x] New unit tests added — prefetch helper + App wiring (above).
- [x] New Playwright e2e added — `state-scope.spec.ts`: a MapCanvas/maplibre chunk-request collector asserts **zero** chunk requests on the bare-URL chooser landing (alongside the unmodified `expect(obsRequests).toHaveLength(0)` C6 AC) and **≥1** on the state-select round-trip. `zip-scope.spec.ts` + `map-cold-load.spec.ts` re-run green (no regression).
- [x] `npm run build --workspace @bird-watch/frontend` — clean. Lazy split preserved: **1** `maplibre-gl-*.js` chunk (identical hash to baseline), **1** `MapCanvas-*.js` chunk (both import sites dedupe to it), **0** maplibre refs in `index.js`.
- [x] `npm run lint` — clean (exit 0). `npx knip` — clean (`prefetch.ts` resolved via the static `import` in `App.tsx`; not flagged). No-DB-write e2e grep — clean.
- [x] (UI only) Playwright MCP smoke — `N/A — not UI`. (Behaviour was still verified live via dev-server request capture: unscoped landing fires 0 chunk requests; scoped landing fires `MapCanvas.tsx` + `maplibre-gl-*.js`.)

## Plan reference

Part of #761 (epic O9 / WS8 sub-issue 8.1). See `docs/analyses/2026-05-29-map-first-rearchitecture-761/report.md` — §5 OQ5 (keep React.lazy split + scope-gated prefetch), §6 WS8.1, §7 precondition 3 ("Prefetch, not inline"), risk R4, §10 phase-4 item O9. Closes #781.

---


julianken:
@Mergifyio queue

PR #791, opened 2026-05-30
test(frontend,e2e): decouple readiness gate + #main-surface selectors from the literal `<main>` tag (POM-only resilience, LAND FIRST) — closes #774

### LINKED ISSUE #774: test(e2e): decouple readiness gate + #main-surface selectors from the literal `<main>` tag (POM-only resilience, LAND FIRST)
**Goal:** Decouple the e2e suite's two structural couplings to the literal `<main id="main-surface">` element — the `waitForAppReady` readiness gate and every hard-coded `main#main-surface` / `#main-surface` selector — so that the map-first shell inversion (which moves the map out of `<main>` and may rename/relocate the readiness-bearing element) does not turn the whole suite red in a single PR. **This is a pure test-resilience refactor: no app behavior, no `frontend/src/**` runtime change.** It MUST land before any structural shell PR (epic Phase 3) so the suite never goes all-red under `retries: 0` (epic risk **R2**).

## Problem

The e2e suite is transitively gated on the literal `main` tag carrying `data-render-complete`, and a second cluster of specs hard-codes the `main#main-surface` selector. The map-first inversion (#761) hoists the map out of `<main>` and re-tiers/renames `#main-surface` (it "either disappears for the map or wraps only non-map surfaces" — epic Scope §1/§2), which silently invalidates both.

**1. The readiness gate keys on the literal `main` tag.** `frontend/e2e/pages/app-page.ts:148-152`:

```ts
async waitForAppReady(timeout = 10_000) {
  await this.page
    .locator('main[data-render-complete="true"]')
    .waitFor({ state: 'attached', timeout });
}
```

`waitForAppReady()` is the canonical "app finished initial load" wait and is called by ~30 of 34 specs (every data-path spec gates on it before asserting). The selector requires the element carrying `data-render-complete` to be a `<main>`. Today `App.tsx:604-605` renders `<main id="main-surface" data-render-complete={renderComplete} ...>`, so this holds — but the inversion moves the readiness-bearing element off `<main>`. When that lands, `waitForAppReady` hangs to timeout and ~30 specs fail at once.

**2. A duplicate inline copy of the same gate** lives at `frontend/e2e/family-legend.spec.ts:118`:

```ts
await page.locator('main[data-render-complete="true"]').waitFor({ state: 'attached', timeout: 15_000 });
```

— same `main`-tag coupling, bypassing the POM. It must move to the same resilient selector.

**3. The loading-state marker is also tag-coupled** at `frontend/e2e/map-cold-load.spec.ts:96`: `page.locator('main[data-render-complete="false"]')` — the `false` (still-loading) variant of the same attribute. Same decoupling applies.

**4. Hard-coded `main#main-surface` / `#main-surface` selectors** appear across a second cluster, each baking in the tag (six sites) or the `main-surface` id (all eight). Note: only the **selector call-sites** below are in scope — comments, assertion messages, and test titles that mention `main#main-surface` (e.g. `error-states.spec.ts:43`, `family-legend-collapse-visibility.spec.ts:134/157/179/214`) are prose, not couplings, and are explicitly NOT touched (see Out of scope):

| File:line | Selector / usage | Context |
|---|---|---|
| `frontend/e2e/pages/app-page.ts:54` | `this.mainSurface = page.locator('main#main-surface')` | Shared POM accessor — **stays**, becomes the sole re-point site (out of `*.spec.ts` glob) |
| `frontend/e2e/a11y.spec.ts:65` | `page.locator('main#main-surface').getAttribute('aria-busy')` | `page.locator` |
| `frontend/e2e/error-states.spec.ts:47` | `expect(page.locator('main#main-surface')).toHaveAttribute('aria-busy', 'false', ...)` | `page.locator` |
| `frontend/e2e/sheet-snap.spec.ts:28,35` | `expect(page.locator('#main-surface')).toHaveAttribute('inert', '')` / `.not.toHaveAttribute('inert', '')` | `page.locator` |
| `frontend/e2e/a11y-bundle-d.spec.ts:178` | `document.querySelector('main#main-surface')` (tabIndex read) | **in `page.evaluate`** |
| `frontend/e2e/axe.spec.ts:299` | `const main = page.locator('#main-surface')` (inert assertion before axe scan) | `page.locator` |
| `frontend/e2e/family-legend-collapse-visibility.spec.ts:141` | `document.querySelector<HTMLElement>('#main-surface')` (chip-bounds clip-guard) | **in `page.evaluate`** (already tag-free today) |
| `frontend/e2e/prod-smoke.preview.spec.ts:16` | `expect(page.locator('main#main-surface')).toHaveAttribute('aria-busy', 'false', ...)` | `page.locator` |

These reference the same DOM anchor through eight independently-maintained selector strings; the inversion changes the anchor once and breaks all eight.

**The two context classes need different fixes.** Six sites are `page.locator(...)` calls in Node test context — those funnel through the shared `app.mainSurface` POM accessor. But **two sites — `a11y-bundle-d.spec.ts:178` and `family-legend-collapse-visibility.spec.ts:141` — run `document.querySelector(...)` *inside* `page.evaluate`**, i.e. in the browser context. A Playwright `Locator` cannot cross into `page.evaluate`, so the POM accessor is structurally unusable there; the only in-browser hook is a CSS selector string. The fix funnels both concerns through the POM where it is reachable, and uses a **tag-AND-id-free** attribute selector in the two browser-context sites so the readiness anchor is referenced by the one attribute (`data-render-complete`) that the map-first inversion carries forward onto whatever element becomes the readiness root.

The fix is to funnel both concerns through the POM so there is a **single** place to re-point when the element moves: make the readiness selector **tag-agnostic** (`[data-render-complete="true"]`, dropping the `main` qualifier) and make `page.locator('#main-surface')` consumers go through the shared `app.mainSurface` accessor (already exists at `app-page.ts:54`) rather than re-declaring `main#main-surface` inline; the two in-browser `page.evaluate` sites resolve via `document.querySelector('[data-render-complete]')` (see Scope). `prod-smoke.preview.spec.ts` is included even though its preview job runs **off** the Mergify queue gate (epic gap 10) — leaving it coupled would let it break silently against the live read-api.

## Scope

Test-only. Touches `frontend/e2e/**` (POM + specs + the manual-test doc). **No `frontend/src/**` change** — the `<main id="main-surface" data-render-complete=...>` element stays exactly as it is today; this PR only makes the *tests* tolerant of it later moving.

### Modify `frontend/e2e/pages/app-page.ts`
- **`waitForAppReady` (lines 148-152):** drop the `main` tag qualifier — change `'main[data-render-complete="true"]'` → `'[data-render-complete="true"]'`. Update the docstring (lines 142-152) to say the readiness flag is keyed on the `data-render-complete` *attribute*, not the `<main>` landmark, and note that map-first (#761) may move it off `<main>`. Keep `{ state: 'attached' }` and the timeout signature unchanged.
- **`mainSurface` accessor (line 54):** this stays the single source of truth for the readiness-surface element **in Node test context** (`page.locator` call-sites). Keep it as `page.locator('main#main-surface')` for now (no app change yet), but add a short comment that it is the **shared handle** every `page.locator`-based spec must use instead of inlining `main#main-surface`, that #761 re-points only this line, and that browser-context (`page.evaluate`) consumers cannot use this Locator and instead select on the tag-agnostic `[data-render-complete]` attribute.

### Modify the readiness-gate duplicates (use the POM, not the literal `main` tag)
- **`frontend/e2e/family-legend.spec.ts:118`:** replace the inline `page.locator('main[data-render-complete="true"]').waitFor(...)` with `await new AppPage(page).waitForAppReady(15_000)` (preserving the 15s timeout this spec passes), importing `AppPage` from `./pages/app-page.js` if not already imported. This routes the gate through the POM.
- **`frontend/e2e/map-cold-load.spec.ts:96`:** make the loading-state marker tag-agnostic — `'main[data-render-complete="false"]'` → `'[data-render-complete="false"]'`. (Leave the `map-cold-load.spec.ts:158` bare `main` wait alone — it asserts the `<main>` landmark itself exists, which is a distinct, still-valid concern today; do not change it in this PR.)

### Re-point the `#main-surface` selector cluster — by context

**Context class A — Node test context (`page.locator(...)`): route through the shared `app.mainSurface` POM accessor.** Replace the inline `main#main-surface` / `#main-surface` locator with `app.mainSurface` (constructing an `AppPage` where one is not already in scope). The assertion semantics (`aria-busy`, `inert`) MUST be byte-for-byte preserved:
- `frontend/e2e/a11y.spec.ts:65` — `aria-busy` poll → `app.mainSurface`.
- `frontend/e2e/error-states.spec.ts:47` — `aria-busy=false` race arm → `app.mainSurface`. (The comment at `:43` mentioning `main#main-surface` is prose — leave it untouched, or freely reword it to track the attribute-keyed framing; it is not a coupling.)
- `frontend/e2e/sheet-snap.spec.ts:28,35` — `inert` present/absent → `app.mainSurface`.
- `frontend/e2e/axe.spec.ts:299` — `inert` assertion → `app.mainSurface`.
- `frontend/e2e/prod-smoke.preview.spec.ts:16` — `aria-busy=false` smoke → `app.mainSurface` (kept in scope despite the off-gate preview job, per epic gap 10).

**Context class B — browser context (`document.querySelector(...)` inside `page.evaluate`): a Playwright `Locator` cannot cross into `page.evaluate`, so `app.mainSurface` is unusable here.** Replace the in-browser selector string with **`document.querySelector('[data-render-complete]')`** — tag-AND-id-free, keyed on the single attribute the map-first inversion carries onto the readiness root. This is verified byte-for-byte semantics-preserving: `App.tsx:604-605` puts `id="main-surface"` and `data-render-complete={renderComplete}` on the **same single element**, and `data-render-complete` is rendered on exactly one element in `frontend/src/**` (App.tsx:605; the only other occurrences are a comment at App.tsx:127 and a CSS comment at styles.css:85), so `[data-render-complete]` resolves to the identical node `#main-surface` resolves to today. Cast/narrow exactly as today (`instanceof HTMLElement` at a11y-bundle-d; `querySelector<HTMLElement>` at fl-collapse) so the `tabIndex` and `getBoundingClientRect()` reads are unchanged:
- `frontend/e2e/a11y-bundle-d.spec.ts:178` — `document.querySelector('main#main-surface')` → `document.querySelector('[data-render-complete]')` (tabIndex read; keep the `main instanceof HTMLElement ? main.tabIndex : null` shape verbatim).
- `frontend/e2e/family-legend-collapse-visibility.spec.ts:141` — `document.querySelector<HTMLElement>('#main-surface')` → `document.querySelector<HTMLElement>('[data-render-complete]')` (chip-vs-main clip rect; the chip selector `.map-surface button[aria-expanded="false"]` is out of scope here — only the `#main-surface` anchor at `:141` is re-pointed). The docblock at `:134`, assertion message at `:157`, and test titles at `:179`/`:214` are prose referencing the surface by name — leave them as-is; they are not selectors.

> **Why `[data-render-complete]` and not tag-free `#main-surface` for the two evaluate sites:** the id-cluster done-signal grep below keys on the literal `#main-surface` id. `document.querySelector('#main-surface')` (tag dropped, id kept) STILL matches that grep, so it cannot reach empty; indeed `fl-collapse:141` is *already* tag-free `#main-surface` today and the grep flags it. Selecting on `[data-render-complete]` is both the genuinely tag-AND-id-free hook (so the grep reaches empty) and the more inversion-resilient anchor (the attribute, not the renamed id, is what rides the new readiness element). This is the explicit resolution of the prior review's contradiction between the Scope wording ("query without the `main` tag qualifier") and the id-keyed done-signal grep.

### Update `frontend/e2e/manual-test.md`
Lines 6-7, 31, 47, 86 describe the readiness gate as `<main id="main-surface"> ... data-render-complete`. Reword to "the element carrying `data-render-complete`" (attribute-keyed, tag-agnostic) so the manual-test doc matches the POM after the inversion. Keep the example check scripts working today.

## Out of scope

- **No `frontend/src/**` change.** Renaming/relocating `#main-surface`, hoisting the map out of `<main>`, and moving `data-render-complete` to a new root element all belong to the shell-inversion sub-issues (epic Phase 3, S1/S2). This PR only makes the tests tolerant of those moves.
- **Prose mentions of `main#main-surface` are NOT couplings and are NOT in scope.** The comment at `error-states.spec.ts:43`, the docblock at `family-legend-collapse-visibility.spec.ts:134`, the assertion message at `:157`, and the test titles at `:179`/`:214` describe behavior in English — they carry no selector and break nothing when the anchor moves. Do not delete or rewrite them to "satisfy" a guard grep; the guards below are anchored to selector call-sites precisely so prose is never flagged.
- **The `aria-busy` semantics bug** (`App.tsx:606` gates `aria-busy` on `state.view === 'feed'`, so it is vacuously false on the map — epic R8) is **not** touched here; re-pointing the selector preserves the existing (trivially-passing) assertion. Fixing the gate + re-baselining is epic sub-issue O1 (WS7.2).
- **Adding a real camera data-attribute** (epic R14 / WS9.3) is a separate scope-spec sub-issue — not this PR. (This PR reuses the *existing* `data-render-complete` attribute as an in-browser hook; it adds no new data-attribute.)
- **Re-baselining the scope-chooser DOM-count assertions** (`state-scope.spec.ts` / `zip-scope.spec.ts` `toHaveLength(0)` and `mapCanvas`/`mainSurface` count-0 — epic R10) belongs to the chooser-as-overlay sub-issue (S1/WS4.2).
- **The `family-legend-collapse-visibility` chip-clip rect re-target** (epic R11 — full-bleed removes the `#main-surface` clip parent) is deferred: here we only re-point the *selector* (to `[data-render-complete]`); whether the rect assertion is re-targeted to the viewport or the spec retired happens in the full-bleed PR (S2/WS9.5).
- **map-cold-load.spec.ts:158**'s bare `main` landmark wait stays as-is (distinct concern, still valid today).

## Acceptance criteria

- `waitForAppReady` (`app-page.ts:148-152`) matches `[data-render-complete="true"]` with **no `main` tag qualifier**; its docstring states the gate is attribute-keyed and notes #761 may move it off `<main>`.
- The inline readiness gate at `family-legend.spec.ts:118` routes through `AppPage.waitForAppReady`; the loading marker at `map-cold-load.spec.ts:96` is tag-agnostic (`[data-render-complete="false"]`).
- Every **Node-context (`page.locator`) site** in the `#main-surface` cluster (a11y:65, error-states:47, sheet-snap:28/35, axe:299, prod-smoke.preview:16) references the surface through the shared `app.mainSurface` POM accessor — **zero** remaining inline `main#main-surface` / `#main-surface` literals at those `page.locator` call-sites.
- Every **browser-context (`page.evaluate` → `document.querySelector`) site** (a11y-bundle-d:178, family-legend-collapse-visibility:141) selects on **`[data-render-complete]`** — tag-AND-id-free, so it carries no `main#main-surface` *or* `#main-surface` literal — with `tabIndex` / bounding-rect reads byte-for-byte unchanged.
- The lone permitted `main#main-surface` selector literal is `app-page.ts:54`, the shared POM accessor (already excluded by `--include='*.spec.ts'`).
- **Verification grep — id cluster (anchored to selector call-sites; this is the authoritative done-signal):**
  ```sh
  grep -rnE "(locator|querySelector)[^)]*#main-surface" frontend/e2e/ --include='*.spec.ts'
  ```
  must return **nothing**. This matches `page.locator('…#main-surface…')` and `document.querySelector('…#main-surface…')` call-sites (whether `main#main-surface` or tag-free `#main-surface`); it does NOT match comments, assertion-message strings, or test titles. Today it returns exactly the 8 in-scope selector sites (a11y:65, a11y-bundle-d:178, error-states:47, prod-smoke:16, sheet-snap:28/35, axe:299, family-legend-collapse-visibility:141). After the swaps it returns empty — the six `page.locator` sites move to `app.mainSurface`, and the two `page.evaluate` sites move to `[data-render-complete]` (which contains no `#main-surface` substring, so the grep clears). **Note `fl-collapse:141` is already tag-free `#main-surface` today and the grep flags it — re-pointing it to `[data-render-complete]` is what clears it; merely "dropping the `main` tag" would not, which is the contradiction this revision fixes.** The 5 prose hits (`error-states:43`, `family-legend-collapse-visibility:134/157/179/214`) are never matched by this command and must remain untouched.
- **Verification grep — readiness gate (already selector-precise):**
  ```sh
  grep -rnE "main\[data-render-complete" frontend/e2e/ --include='*.spec.ts'
  ```
  must return **nothing**. Today it returns exactly `family-legend.spec.ts:118` and `map-cold-load.spec.ts:96`, both of which the Scope converts; only the tag-agnostic form remains afterward.
- **End-state assertion (selector-resolution, not just literal-text):** every `page.locator(...)` and every in-`page.evaluate` `document.querySelector(...)` in `frontend/e2e/**` that targets the readiness surface resolves through either `app.mainSurface` (Node context) or a tag-AND-id-free `[data-render-complete]` query (browser context). No spec re-declares the `main#main-surface` *or* `#main-surface` anchor inline. Both guard greps above pass **precisely** when the Scope-listed sites — and only those — are converted; the two evaluate sites and the id-cluster grep now agree (the prior contradiction is resolved).
- **Every assertion's semantics are byte-for-byte preserved** — `aria-busy`, `inert`, `tabIndex`, and bounding-rect checks assert the same values against the same DOM element they do today. `[data-render-complete]` is verified to resolve to the identical node as `#main-surface` today (both attributes ride the same single `App.tsx:604-605` element, and `data-render-complete` is rendered on exactly one element in `frontend/src/**`), so the browser-context swaps are a no-op at runtime. No assertion is weakened or removed.
- `manual-test.md` describes the readiness gate as attribute-keyed, not `<main>`-tag-keyed.
- **This PR alone leaves all four CI checks green — `test`, `lint`, `build`, `e2e` — plus `knip`. No "leave CI red, fix next PR."** Because no app code changes, the full e2e suite must pass against the *unchanged* `#main-surface` shell exactly as it does on `main` today; the refactor is a no-op at runtime. (Required-check reminder: `knip (informational)` and `orphan-classname-check` also gate the Mergify queue — keep both clean; this PR adds no `className`s, so orphan-classname is `N/A — no new classNames`.)
- Per CLAUDE.md: this is a **test-only** PR under `frontend/**` with no visible UI change — mark the PR's Screenshots section `N/A — not UI` and skip live Playwright UI verification.

## Test plan

No new product behavior, so **no new unit tests** and **no new e2e specs** — this is a re-baseline of *existing* specs' selectors with semantics held constant. Concretely:
- **POM:** `frontend/e2e/pages/app-page.ts` — `waitForAppReady` + `mainSurface` are the shared chokepoints for Node-context consumers; all `page.locator` consumers exercise them transitively. The two browser-context `page.evaluate` sites cannot route through the POM Locator and instead select on `[data-render-complete]`.
- **Re-baselined specs** (selector swap, assertions unchanged): `frontend/e2e/family-legend.spec.ts`, `frontend/e2e/map-cold-load.spec.ts`, `frontend/e2e/a11y.spec.ts`, `frontend/e2e/error-states.spec.ts`, `frontend/e2e/sheet-snap.spec.ts`, `frontend/e2e/a11y-bundle-d.spec.ts`, `frontend/e2e/axe.spec.ts`, `frontend/e2e/family-legend-collapse-visibility.spec.ts`, `frontend/e2e/prod-smoke.preview.spec.ts`.
- **Validation:** run the full e2e suite (`npm run test:e2e --workspace @bird-watch/frontend`, or the repo's e2e script) locally and confirm green; run the two **anchored** grep guards from the Acceptance criteria (`(locator|querySelector)[^)]*#main-surface` and `main\[data-render-complete`, both scoped `--include='*.spec.ts'`) and confirm both return empty; run `knip` clean. The no-DB-write grep must stay clean: `grep -rE "request\.(post|patch|delete|put)|fetch\(.*method:|fetch\(.*[\"']POST[\"']" frontend/e2e/` returns nothing. Because the prod-smoke preview job runs off-queue against a live read-api, also confirm `prod-smoke.preview.spec.ts` compiles and its selector resolves (it is exercised in the preview workflow, not the queue gate).
- **Spot-check the two `page.evaluate` swaps resolve to the same node:** the a11y-bundle-d:178 `tabIndex` test asserts `toBe(0)` and the fl-collapse:141 clip-guard asserts the chip is inside the surface rect — both must still pass, confirming `[data-render-complete]` resolves to the same element `#main-surface` did. (If either fails, the attribute-uniqueness premise is wrong and the swap must be revisited — but `frontend/src/**` shows `data-render-complete` on exactly one rendered element.)
- **Do NOT use the unanchored `grep -rn "main#main-surface" …` as a done-signal** — it matches the 5 prose lines the Scope deliberately leaves alone, so it can never reach empty without illegitimately editing comments/titles. The anchored selector-call-site grep above is the correct guard. Equally, do NOT treat "dropped the `main` tag" as sufficient for the two `page.evaluate` sites — tag-free `#main-surface` still trips the id-cluster grep; only `[data-render-complete]` clears it.

## Dependencies

Part of #761. Epic-relative blockers: **none** — this is a Phase 1 pure-refactor that **must merge before any structural shell PR** (it unblocks the Phase 3 inversion: F1/S1/S2 and the overlay sub-issues all assume a tag-agnostic readiness gate). The epic holds the authoritative numbered dependency graph; this issue corresponds to **P2** in epic §10 "Recommended sub-issue sequencing" (Phase 1, "must merge before any shell PR").

## Plan traceability

- Research report: `docs/analyses/2026-05-29-map-first-rearchitecture-761/report.md` — §6 **WS9 (Test-coupling migration)**, sub-issue **9.1** ("POM readiness gate + selector resilience — **LAND FIRST**"); §5 **OQ6** ("Land a POM-only resilience PR FIRST … the POM PR must merge before any shell PR or the whole suite goes red at once"); §8 risk **R2** (readiness gate ~30/34 specs); §10 Phase 1 **P2**; §9 gap **10** (prod-smoke + error-states `main#main-surface` coupling; preview job off the Mergify gate).
- Epic workstream: #761 Scope §9 "Test-coupling migration" (the `a11y.spec.ts` `main#main-surface` `aria-busy` selector + windowed-structure couplings).

julianken:
## Diagrams

How the two structural couplings funnel to a single re-point site after this PR:

```mermaid
flowchart TD
    subgraph Before["Before — 2 couplings, 10 inline sites"]
        G1["waitForAppReady<br/>main[data-render-complete=true]"]
        G2["family-legend:118 inline gate<br/>map-cold-load:96 loading marker"]
        S1["6× page.locator('main#main-surface')<br/>a11y · error-states · sheet-snap×2 · axe · prod-smoke"]
        S2["2× page.evaluate querySelector('#main-surface')<br/>a11y-bundle-d:178 · fl-collapse:141"]
    end
    subgraph After["After — tag-agnostic, single re-point site for #761"]
        P["app.mainSurface (POM, app-page.ts:54)<br/>= page.locator('main#main-surface')<br/>← #761 re-points ONLY this line"]
        A["[data-render-complete] attribute<br/>(tag-AND-id-free, rides the readiness root)"]
    end
    G1 -->|drop main qualifier| A
    G2 -->|route via POM / tag-agnostic| A
    S1 -->|Node ctx → shared accessor| P
    S2 -->|browser ctx, Locator can't cross| A
```

## Summary

- **Why:** the map-first inversion (#761, Phase 3) hoists the map out of `<main>` and re-tiers/renames `#main-surface`; today the e2e suite is transitively gated on that literal `<main>` tag (`waitForAppReady`, ~30/34 specs — epic **R2**) plus a second cluster hard-coding `main#main-surface` across 8 independent selector strings. Landing the shell inversion before this would turn the whole suite red at once under `retries: 0`. This **must land first** (epic §10 **P2**, Phase 1).
- **What:** pure test-resilience refactor — `waitForAppReady` + both gate duplicates key on the tag-agnostic `[data-render-complete]` attribute; the 6 Node-context `page.locator` sites route through the shared `app.mainSurface` POM accessor (the sole #761 re-point site); the 2 browser-context `page.evaluate` sites (where a Playwright `Locator` can't cross) select on `[data-render-complete]`, which resolves to the same single element `#main-surface` does today (App.tsx renders the attribute on exactly one element).
- **No `frontend/src/**` change** — the `<main id="main-surface" data-render-complete>` element is untouched; the refactor is a runtime no-op and every assertion (aria-busy, inert, tabIndex, bounding-rect) is byte-for-byte preserved.

## Screenshots

N/A — not UI (test-only change under `frontend/e2e/**`; no `frontend/src/**` runtime change — per the CLAUDE.md test-only exemption).

## Test plan

- [x] `npm test --workspace @bird-watch/frontend` — green (66 files, **961 tests pass**)
- [x] `npm run lint` — green (root delegates `--workspaces --if-present`; no workspace defines a lint script, vacuous as on `main`)
- [x] `npx knip --no-progress` — clean (exit 0; single output line is a pre-existing config *hint*, not a finding)
- [x] e2e: all **46 tests across every touched spec pass** (single-worker dev-server run, DB seeded with the CI e2e seed SQL). Whole-suite parallel runs hit `ERR_CONNECTION_REFUSED` from the local read-api being OOM-killed (`code 137`) under repeated heavy concurrent load — an environment crash, never a selector/assertion failure. CI runs read-api as its own healthy service container.
- [x] **Guard grep — id cluster** (authoritative done-signal): `grep -rnE "(locator|querySelector)[^)]*#main-surface" frontend/e2e/ --include='*.spec.ts'` → **empty**
- [x] **Guard grep — readiness gate**: `grep -rnE "main\[data-render-complete" frontend/e2e/ --include='*.spec.ts'` → **empty**
- [x] **No-DB-write grep**: `grep -rE "request\.(post|patch|delete|put)|fetch\(.*method:|fetch\(.*[\"']POST[\"']" frontend/e2e/` → **empty**
- [x] The 5 prose mentions of `main#main-surface` (`error-states:43`, `fl-collapse:134/157/179/214`) left untouched; the lone permitted selector literal is `app-page.ts:54` (shared POM accessor)
- [x] No new unit tests / e2e specs — selector re-baseline with semantics held constant (no new product behavior)
- [x] orphan-classname: N/A — no new classNames

## Plan reference

Part of #761 (epic), sub-issue **P2** (Phase 1, "must merge before any shell PR"). Research report: `docs/analyses/2026-05-29-map-first-rearchitecture-761/report.md` — §6 WS9.1 ("POM readiness gate + selector resilience — LAND FIRST"), §5 OQ6, §8 risk R2, §10 Phase 1 P2, §9 gap 10 (prod-smoke off the Mergify gate). Closes #774.

---


julianken:
@Mergifyio queue

Issue #783, opened 2026-05-30
O5 — Cap FamilyLegend mobile width (R6) + rewire SpeciesDetailSheet/rail z-index above the legend/popover band (R3)


julianken:
**Goal:** Stop two mobile-overlay defects that the map-first inversion turns from latent to user-visible. (1) Cap `FamilyLegend` at a real overlay width on mobile — today it widens to almost the full viewport (~94%, ~366px at 390px), and a previously-expanded user gets a near-full-width legend on cold mobile load (**R6**). (2) Lift the full-snap `SpeciesDetailSheet` (and the rail's stacking expectation) above the legend/popover band — today the full sheet's `z-index:20` sits **below** the legend (`40`) and observation popover (`41`), masked only by the `inert`-on-`#main-surface` containment that O1 is about to dismantle (**R3**).

This is a Phase-4 overlay-generalization sub-issue. It lands the WS6.2 + WS6.3 work and clears risks R3 and R6 from the epic's register.

---

### R6 — the legend reaches ~94% of a 390px viewport, not 67%

`.family-legend` is a `position:absolute` overlay anchored bottom-left over the canvas with a base `max-width: 240px` (`frontend/src/styles.css:864-876`, z-index `var(--z-panel)` = 40 at `styles.css:868`). Under the global mobile breakpoint it is **widened**, not tightened:

```css
/* frontend/src/styles.css:976-983 */
@media (max-width: 760px) {
  .family-legend {
    max-width: calc(100vw - 2 * var(--space-md));
  }
  .family-legend-entries { max-height: 240px; }
}
```

With `--space-md: 12px` (`styles.css:12`), at a 390px viewport that resolves to `390 - 24 = 366px` — **~94% of the width**, not the ~67% the epic's WS6 prose claims. This rule borrows the global content breakpoint (760px) to size an *overlay*, which is exactly the anti-pattern the report's OQ2 flags ("size/anchor breakpoints should be decoupled from the 760px content breakpoint").

[...]

Issue #773, opened 2026-05-30
S3: Resolve TODO(#737) — asymmetric `fitBounds` top-padding (header + scope chrome clearance) + verify `map.resize()` / `isMobile` on the S2 flex→fixed transition


julianken:
**Goal:** Resolve `TODO(#737)` in `MapCanvas.tsx` so `fitBounds` clears the floating `AppHeader` + `ScopeControl` chrome (markers stop hiding under the bar), and verify whether the map needs an explicit `map.resize()` on the flex-child → fixed-inset container resize introduced by S2 (hoist map out of `<main>`). Today `fitBounds` uses a uniform `padding: 48` at both the mount-frame (`MapCanvas.tsx:506`) and the imperative camera effect (`MapCanvas.tsx:661`), and `MapCanvas.tsx` itself never calls `map.resize()` on a container-box change — it has no `ResizeObserver` of its own.

This is the third shell-inversion PR (epic-relative **S3**). It depends on **S2** having already moved `.map-surface` to `position: fixed; inset: 0` — only then does the map box change size on the flex→fixed transition, and only then does the floating header/scope chrome overlap the canvas top edge.

## Problem

Two grounded concerns, both invisible while the map was a padded flex child and both surfaced by S2's full-bleed extraction. The first is a defect that must be fixed in this PR; the second is a verify-first item that becomes a fix only if a repaint flash is demonstrated.

**1. Uniform `fitBounds` padding occludes the framed region under floating chrome.**

The scope-driven camera-intent effect frames a state/region envelope with uniform padding (two-arg call — `bounds` then an options object, no third callback/eventData arg):

```ts
// frontend/src/components/map/MapCanvas.tsx:655-665
map.fitBounds(activeBounds, {
  // TODO(#737): asymmetric top padding == ScopeControl bar height — the
  // in-state on-map ScopeControl floats over the canvas, so uniform 48
  // would let the bar occlude the top of the framed state. Land uniform
  // 48 until #737 fixes the bar height, then switch to
  // `{ top: <barHeightPx>, bottom: 48, left: 48, right: 48 }`.
  padding: 48,
  maxZoom: 12,
  essential: true,
  duration: prefersReducedMotion ? 0 : 600,
});
```

The same uniform `48` is also baked into the **mount-frame** first-paint, so a scoped *deep link* (not just an in-session scope change) lands occluded too:

```ts
// frontend/src/components/map/MapCanvas.tsx:504-508
const initialViewStateRef = useRef(
[...]

PR #471, opened 2026-05-11
fix(styles): add CSS rules for 8 orphan classNames — W2 orphan-CSS sweep (#455)

julianken:
## Diagrams

```mermaid
graph LR
    subgraph Before
        A[".map-lede"] -->|"bare h1 ~32px bold"| B["browser default"]
        C[".species-detail-sheet--peek"] -->|"no modifier rule"| D["fallback to base only"]
        E[".feed-surface"] -->|"bare div"| F["no padding/max-width"]
        G[".map-loading-skeleton"] -->|"collapses to 0px"| H["invisible on slow loads"]
    end
    subgraph After
        I[".map-lede"] -->|"26px semibold, no UA margin"| J["spec newspaper-lede"]
        K[".species-detail-sheet--peek"] -->|"z-index 10 + overflow hidden"| L["correct snap styling"]
        M[".feed-surface"] -->|"padding-lg box-sizing"| N["surface-level constraint"]
        O[".map-loading-skeleton"] -->|"token shimmer animation"| P["visible loading state"]
    end
```

## Summary

- Eight `className` strings used in JSX had no matching CSS rule at `origin/main SHA 85efd1e`, causing visible defects: `.map-lede` rendered as a bare `<h1>` at ~32px (spec calls for 26px lede-size), `.species-detail-sheet--peek` and `--half` had no modifier rules to distinguish snap states, and `.map-loading-skeleton` collapsed to zero height on slow loads.
- All 8 rules added to `frontend/src/styles.css` (append-only, no selector reordering) using semantic tokens from `tokens.css`; the inline sr-only escape-hatch in `FilterSentence.tsx:121` is deleted now that `.filter-sentence-live` has a proper CSS rule.
- Pre-existing silhouette-sprite warnings (3x, `famA`/`famB` test data) remain unchanged; zero JS errors on any touched surface.

## Screenshots

- [x] Every new/renamed `className` in this PR has a matching CSS rule in `frontend/src/styles.css`
- [ ] Multi-viewport design QA: chrome-devtools-mcp browser session was closed during paste-flow; screenshots captured in Playwright MCP session but user-attachments upload could not complete. Bot reviewer should drive `npm run dev --workspace @bird-watch/frontend` against PR head SHA and capture independently per CLAUDE.md §UI verification protocol.

**Playwright MCP validation completed (local dev server, commit c02ad42):**

[...]

PR #795, opened 2026-05-30
feat(frontend): invert app shell to a full-viewport map root  closes #775

### LINKED ISSUE #775: feat(frontend): S2 — invert app shell to full-viewport map root (map-layer wrapper + floating fixed header)
**Part of #761** (map-first re-architecture). This is sub-issue **S2** — the core shell inversion: the map stops being a flex child of the padded `<main>` and becomes a `position: fixed; inset: 0` viewport-root layer, with the chrome floating on top.

> **Branch note (post-S1 baseline).** S2 builds **on top of S1 (#772)**, not on current `main`. S1 has already converted the unscoped early-return at `App.tsx:541-552` (the `if (state.scope.kind === 'unscoped') return <ScopeChooser/>` block, verified present on `main`) into an **always-render-shell + inert focus-trapped scrim over a mounted, idle map** (locked decision 1; #761 Map lifecycle). So by the time S2 starts, the chooser is a **scrim sibling inside the shell**, no longer a pre-`.app` early-return, and the map is **always mounted** — idle under the scrim when unscoped, live when scoped. The zero-`/api/observations` landing AC is preserved by the **existing `scopeActive` fetch gate** (`App.tsx:134`, `state.scope.kind !== 'unscoped'`), which is orthogonal to `mapVisible` (`App.tsx:273`) and keeps the network panel empty even though the map is mounted. Line anchors below that point at the *pre-S1* early-return are given for traceability against today's `main`; S2 operates on the post-S1 tree where that block is a scrim, and the map block S2 hoists is the **always-mounted** one.

## Problem

Today the map is *windowed inside the UI*, not the application itself. The app shell is a vertical flex column — `.app { display: flex; flex-direction: column; height: 100vh }` (`frontend/src/styles.css:77-81`) — and `MapSurface` + `ScopeControl` render **inside** the single scrollable `<main id="main-surface">`, which owns `flex: 1; min-height: 0; padding: var(--space-lg); overflow-y: auto; display: flex; flex-direction: column` (`styles.css:86-93`).

Concretely, in `App.tsx` the map block is gated by `mapVisible` (`App.tsx:273`, `= state.view === 'map' || state.view === 'detail'`) and rendered as a child of `<main>` — and post-S1 it is the **always-mounted** map (idle under the scrim when unscoped, live when scoped); `mapVisible` is the view-tier gate, `scopeActive` (`App.tsx:134`) is the orthogonal fetch gate:

- `<main ref={mainRef} id="main-surface" … tabIndex={0}>` opens at `App.tsx:602-614`.
- The `{mapVisible && (<><ScopeControl …/><MapSurface …/></>)}` block lives **inside** that `<main>` at `App.tsx:633-679` — note `ScopeControl` (`App.tsx:641-648`) and `MapSurface` (`App.tsx:649-677`) are **siblings inside the fragment**, not parent/child.
- `</main>` closes at `App.tsx:689`.
- Only *after* `</main>` do `SpeciesDetailRail` / `SpeciesDetailSheet` render as **viewport siblings** (`App.tsx:690-710`) — the proven `position: fixed` floating-over-the-map pattern (#663).

So the map inherits the 16px `--space-lg` padding on all four sides (`styles.css:89`) and is capped by the flex box that the sticky `AppHeader` (`styles.css:164-174`, `position: sticky; top: 0; z-index: 10`) displaces down. `.map-surface` itself is a *controlled flex child* — `position: relative; width: 100%; flex: 1; min-height: 0` (`styles.css:686-691`) — so the canvas fills only that constrained box, never the viewport.

The detail surfaces already point at the right answer: they sit as fixed siblings of `<main>` (`App.tsx:690-710`; rail `position: fixed` at `styles.css:1081`+, `z-index: calc(var(--z-panel) + 5)`=45 at `styles.css:1092`). **S2 generalizes that pattern to the map + ScopeControl.** Padding alone will not free the map: the sticky header occupies real flow height, so a padded-but-in-`<main>` map stays trapped beneath it. The fix is **DOM extraction** — hoist the block out of `<main>` and make it viewport-filling via a **mandatory** fixed-inset wrapper.

> **Note on the unscoped (scrim) landing.** Post-S1 the map is always mounted — when unscoped it sits **idle beneath the inert scrim** (locked decision 1), not unmounted. S2's hoist therefore applies to that always-mounted block in **both** states; the wrapper (`#map-layer`, below) mounts in the unscoped case too, with the canvas idle behind the scrim and the `scopeActive` fetch gate (`App.tsx:134`) holding `/api/observations` at zero. S2 must **not** make `#map-layer` conditional on `scopeActive`, and must **not** newly gate it such that it disappears on the unscoped scrim landing — doing so would silently break S1's always-mounted-under-scrim contract and the camera-preservation / zero-`/api` landing model the epic depends on. Whether the hoisted block stays under the existing `{mapVisible && …}` gate or moves to an always-mounted position is specified in Scope; the invariant is that the **unscoped scrim landing still mounts `#map-layer` (idle)**.

> **Note on `.scope-control`'s containing block (corrected — this is the load-bearing reason the wrapper is mandatory):** `.scope-control` is `position: absolute; inset-block-start/inline: var(--space-md); z-index: var(--z-panel)` (`styles.css:1903-1920`). Critically, in the JSX it is a **sibling** of `MapSurface`, **not a child of `.map-surface`** (`App.tsx:641` vs `App.tsx:649`, both direct children of the `{mapVisible && (<>…</>)}` fragment). An `absolute` element resolves against its nearest *positioned* ancestor. Today that ancestor is **neither** `.map-surface` (which is its sibling, not an ancestor) **nor** `#main-surface` (`styles.css:86-93` sets no `position`, so it is statically positioned) — it is the **initial containing block (the viewport)**, which is why ScopeControl currently anchors to the viewport top edge despite living inside an in-flow `<main>`. After the hoist, ScopeControl must *keep* resolving against a box that equals the viewport. Simply making `.map-surface` `position: fixed; inset: 0` does **not** help ScopeControl, because `.map-surface` is still ScopeControl's sibling, not its ancestor. Therefore S2 **requires** a wrapper element that contains both ScopeControl and MapSurface and is itself positioned. **The `#map-layer` wrapper is mandatory, not optional** (see Scope): `<div id="map-layer">` becomes `position: fixed; inset: 0`, ScopeControl's `absolute` offsets then resolve against that wrapper (≡ viewport), and `.map-surface` fills the wrapper. `.family-legend` (`styles.css:864-876`, `position: absolute; z-index: var(--z-panel)`) and `.observation-popover` (`styles.css:996-1007`, `z-index: calc(var(--z-panel) + 1)`) live *inside* `MapSurface`, so once `.map-surface` is positioned they anchor to it correctly and ride along — but ScopeControl, being a sibling, depends on the wrapper specifically.

> **Note on the detail-sheet z-stack (do not touch in S2):** the bottom-sheet snap tiers are bare-integer z-indices on their own stacking comparison — base `.species-detail-sheet` `z-index: 10` (`styles.css:1132`), `--peek` 10 (`styles.css:1303`), `--half` 15 (`styles.css:1311`), `--full` 20 (`styles.css:1148`). These are **below** the `--z-panel`=40 band. Today the sheet visually covers the header only because the header is *also* a bare-integer sticky context (`z-index: 10`, `styles.css:173`) in the normal flow. The moment the header goes `position: fixed` on the 40+ `--z-chrome` token, it outranks the entire 10/15/20 sheet stack, so a full-snap sheet will **no longer** occlude it — and `inert` cannot compensate, because the header is a **sibling** of `#main-surface` (never a descendant) and S2 keeps `inert` wired to `#main-surface` (`SpeciesDetailSheet.tsx:42-43`). Reconciling the sheet stack onto the named scale is epic #761's **mobile-overlay workstream (O4-family)**, not S2 — see Out of scope. S2 makes **no claim** about sheet-vs-header z-order.

## Scope

**Modify `frontend/src/App.tsx`**
- Hoist the `{mapVisible && (<><ScopeControl/><MapSurface/></>)}` block (`App.tsx:633-679`) **out of `<main>`** so it renders as a viewport sibling of `<main>` — adjacent to where `SpeciesDetailRail`/`Sheet` already render (after `App.tsx:689`). The map block is fully prop-threaded from App-level state already (every prop on `MapSurface`/`ScopeControl` comes from `App` scope), so no new prop-drilling is required — only moving the JSX out of the `<main>` element.
- **Preserve the always-mounted-under-scrim invariant (post-S1).** The hoisted block is the **always-mounted** map; on the unscoped scrim landing `#map-layer` still mounts (idle, behind S1's inert scrim) — `#map-layer` must **not** become conditional on `scopeActive`. Keep the existing `{mapVisible && …}` gate as-is (it is a view-tier gate — `'map' || 'detail'` — that is already true on the scrim landing, which renders the `'map'` view beneath the scrim, so the map mounts when unscoped); do **not** introduce a new `scopeActive`-keyed gate around `#map-layer`. The `scopeActive` fetch gate (`App.tsx:134`) stays the sole mechanism holding `/api/observations` at zero on the unscoped landing.
- **Wrap the hoisted block in a mandatory `<div id="map-layer">` container** that holds **both** `ScopeControl` and `MapSurface`. This wrapper is **required, not optional** — it is the positioned ancestor that ScopeControl's `position: absolute` resolves against post-hoist (see the corrected `.scope-control` note above; without it ScopeControl would resolve against whatever the new sibling/ancestor chain yields, not a guaranteed viewport box). It also gives a future inert/aria-busy retarget (sub-issue O1) one stable node to target. This issue only introduces the wrapper; it does **not** rewire inert (that stays on `#main-surface` here — see Out of scope).
- `#main-surface` then wraps **only the non-map view surface**: the `state.view === 'feed'` `FeedSurface` block at `App.tsx:615-632` (live until F1 lands). Keep `<main id="main-surface">`, keep `mainRef`, keep `data-render-complete`, keep `tabIndex={0}` — the readiness gate and the scroll-bypass affordance (#586) stay. (The App-level error path is a *pre-shell* early-return — `App.tsx:554-562` returns a bare `<StatusBlock>` before `.app`, `AppHeader`, or `#main-surface` are ever rendered — so it is not a `#main-surface` child and is untouched by this inversion. See the R15 bullet and Out of scope.)

**Modify `frontend/src/components/AppHeader.tsx`**
- **Retarget the Map tab's `aria-controls` from `"main-surface"` to `"map-layer"`** (`AppHeader.tsx:125`). The single tab is the **"Map view"** tab (`TABS = [{ value: 'map', label: 'Map', accessibleName: 'Map view' }]`, `AppHeader.tsx:39-41`), and today its `aria-controls="main-surface"` is *correct* because the map renders inside `<main id="main-surface">`. After S2's hoist the map lives in the sibling `#map-layer` and `#main-surface` wraps only the (downstream-removed) feed branch — so leaving `aria-controls="main-surface"` would silently point the "Map view" tab at a region that **no longer contains the map**, an a11y semantic regression introduced *by* this PR. Retargeting to `"map-layer"` keeps the tab→region semantics honest: the tab controls the region it actually toggles into view. (This is the one a11y-attribute change S2 owns; the inert/aria-busy/aria-live retarget on `#map-layer` stays O1 — see Out of scope — but the **tab's** `aria-controls`, which O1 does not touch, must be fixed in the same PR that moves the map out of `#main-surface`, or the regression ships uncaught.)

**Modify `frontend/src/styles.css`**
- **Give `#map-layer` `position: fixed; inset: 0`** with a z-index below the chrome band (it sits beneath `--z-chrome`; the canvas is the backdrop). With the wrapper fixed-inset, `.map-surface` (`styles.css:686-691`) returns to filling its parent: replace the flex-child constraints `flex: 1; min-height: 0; width: 100%` with `position: absolute; inset: 0` (or `width:100%; height:100%`) of the `#map-layer` box. The inner `MapCanvas` is already `width:100% / height:100%` of `.map-surface`, so it fills the viewport once the wrapper does, and ScopeControl (sibling of `.map-surface`, child of `#map-layer`) anchors to the fixed wrapper. Leave `#main-surface` padding (`styles.css:89`) untouched: it now only pads the non-map view surface.
- **AppHeader → fixed floating chrome.** Change `.app-header` (`styles.css:164-174`) from `position: sticky; z-index: 10` to `position: fixed; top: 0; left: 0; right: 0;` on the new `--z-chrome` token introduced by S1. `--z-chrome` must sit **above** the map-assist overlays (legend / scope / observation-popover at `--z-panel`=40 / `--z-panel`+1=41, `styles.css:868`/`1907`/`998`) and **below** the detail rail (`calc(--z-panel + 5)`=45, `styles.css:1092`) — so the in-place rail (#663) still floats over the floating header. The bottom-sheet snap tiers (10/15/20) are a separate, already-below-`--z-panel` stack that S2 deliberately does not reconcile (see the detail-sheet note above and Out of scope). (S1 owns the named scale; S2 consumes the `--z-chrome` token it defines.)
- **R15 — every chrome-flow surface that the fixed header would overlap must clear it.** When the header leaves the flow (`position: fixed`), the surfaces that previously sat below it in normal flow rise under it. Two such surfaces exist, and **both** must be cleared in S2:
  1. **`#main-surface`** — give it a `padding-top` (and `scroll-margin-top`) equal to the header height so the one surface it actually wraps — the live `FeedSurface` branch (removed downstream, but live until F1 lands) — does not render underneath the floating chrome.
  2. **`.filters-panel`** (`App.tsx:581-600`; `styles.css:1663-1665`, `position: relative`, **no `z-index`**) — this panel is a **sibling of `<main>` rendered before it**, opened via the header's `onOpenFilters` (`App.tsx:578`), and **reachable on the map view** (the trigger lives in `AppHeader`, present on every view). Because it sits at the top of the flow with no z-index, the now-`position: fixed` header overlaps its top edge and occludes the close button / first filter control. S2 must clear the header for it too — the minimal fix is a `margin-block-start`/`top` offset equal to the header height (or wrap it so it opens below the chrome). This is an interim clearance only; the full conversion of `.filters-panel` to a floating sheet + backdrop with proper z-layering is **O4** — S2 just keeps it from sliding under the fixed header.
  - The **map layer is exempt** — `#map-layer` is `inset: 0` beneath the header on purpose (chrome floats over the canvas, Google-Maps-style). The **App-level error path is also not in scope** for this clearance: it never renders inside the shell (it early-returns a bare `StatusBlock` before `.app`/`AppHeader`/`#main-surface` mount — `App.tsx:554-562`), so there is no fixed header above it to clear.

**Verify (assert in this PR):**
- **The Map tab's `aria-controls` resolves to the region it now controls.** After the retarget, `AppHeader.tsx:125` reads `aria-controls="map-layer"` and that id resolves to the present, hoisted `<div id="map-layer">` (the region that actually contains the map). Assert the **semantics survive**, not merely that *some* element is present: the "Map view" tab's controlled region is the map layer, not the now-feed-only `#main-surface`. (Do **not** leave a presence-only "still resolves to a present element" check against `#main-surface` — that would assert the regression passes.)
- `<main id="main-surface">` is retained (with `id`, `data-render-complete`, `mainRef`, `tabIndex={0}`) and still wraps the feed branch — but it is **no longer** the target of the Map tab's `aria-controls`.
- On the unscoped scrim landing, `#map-layer` mounts (idle) and `/api/observations` stays at zero (existing `scopeActive` gate) — i.e. the always-mounted-under-scrim invariant survives the hoist.

> **Map `resize` on the flex→fixed transition is OUT of scope — owned by S3 (#773).** Moving the map container from `flex: 1` to a `position: fixed; inset: 0` wrapper changes the box out from under the MapLibre GL canvas, which does not auto-observe its container, and a corrective `map.resize()` may be needed for the canvas to adopt the new box. **That resize is S3's, not S2's** — the research report assigns "`map.resize` on transition" to S3 (`report.md:251`, "S3. … + `map.resize` on transition (WS2.2 + gap 8)") and lists it as an unhandled gap owned by WS2/S3 (`report.md:222`, gap 8). Firing it from App is also structurally blocked here: App holds no map ref, `MapSurfaceProps` (`MapSurface.tsx:49`+, signature `:180-202`) exposes **no** ref / imperative handle / resize trigger, and the only `mapRef` lives inside the `React.lazy`-loaded `MapCanvas` (`MapCanvas.tsx:482` `useRef<MapRef>`; `getMap()` at `:638`/`:1715` etc.). Wiring a corrective resize from App would demand a net-new `forwardRef`/`useImperativeHandle` chain App→MapSurface→MapCanvas (or a transition-keyed effect inside MapCanvas) — which contradicts S2's "only moving JSX, no new prop-drilling" contract and is not in S2's files-to-modify list (only `App.tsx` + `styles.css` + the single `aria-controls` retarget in `AppHeader.tsx`). S2's full-bleed geometry is proven as **steady-state** after `waitForAppReady` (see AC + Test plan); a settled rect does not depend on which call sized the canvas, so deferring the resize to S3 does **not** weaken S2's geometry proof. If S3's resize work surfaces a stale-size repaint that S2's steady-state e2e somehow exposes first, **note it and hand it to S3 — do not add ref plumbing or a ResizeObserver in S2.**

## Out of scope (defer to the named downstream sub-issue)

- **Corrective `map.resize()` on the flex→fixed transition**, plus the `forwardRef`/`useImperativeHandle` (or transition-keyed internal effect) plumbing it would require — App holds no map ref and `MapSurfaceProps` exposes none; the `mapRef` lives only inside lazy `MapCanvas` (`MapCanvas.tsx:482`). → **S3** (`report.md:251` / gap 8 `report.md:222`).
- **Asymmetric `fitBounds` top-padding** so markers clear the floating header/ScopeControl — `MapCanvas` uniform 48px padding with `TODO(#737)`. → **S3**.
- **`map.resize` ResizeObserver / `isMobile` width re-read** hardening (`isMobile` width read at `MapCanvas.tsx:1093` now sees the full viewport). → **S3**.
- **`.filters-panel` → floating sheet + backdrop with proper z-layering.** S2 only adds the interim header-clearance offset so the panel is not occluded on the map view; converting it to a top-layer floating sheet (backdrop, focus trap, z-index on the named scale) is **O4**. → **O4**.
- **Detail-sheet snap-tier z-order vs. the fixed header.** The bottom-sheet tiers are bare-integer `z-index: 10/15/20` (`styles.css:1132`/`1311`/`1148`) on a stack **below** `--z-panel`=40; once the header is `position: fixed` on `--z-chrome` (necessarily ≥41), the header outranks the sheet and `inert` on `#main-surface` cannot mute a sibling header. Lifting `.species-detail-sheet--full` (and the half/peek tiers) onto the named `--z-chrome`+N scale so a full-snap sheet occludes the header is epic #761's **mobile-overlay workstream** — explicitly **not** S2. S2 makes no occlusion/inert claim about the sheet. → **O4-family / mobile-overlay sub-issue** (per #761 §6 "Guarantee sheet z-index ≥ map controls + popovers without a z-index war").
- **Scope-gating `onViewportChange`** so a persistent unscoped map fires zero `/api/observations`. → **S4** (this PR keeps the existing chooser/scrim from S1 plus the `scopeActive` fetch gate, which already suppress the fetch).
- **Retargeting `inert` / `aria-busy` / `aria-live`** from `#main-surface` to the `#map-layer` wrapper, and the camera data-attribute. → **O1**. (Sheet `inert` stays wired to `mainRef`/`#main-surface` in this PR — `SpeciesDetailSheet.tsx:42-43`.) **Exception — the Map tab's `aria-controls` is NOT deferred to O1:** O1 retargets the wrapper's inert/busy/live state, but the *tablist* tab→region pointer (`AppHeader.tsx:125`) is a different attribute O1 does not touch, and leaving it pointed at `#main-surface` would silently ship a "Map view" tab that controls a map-free region the moment S2 moves the map out. So S2 itself retargets the tab's `aria-controls` to `"map-layer"` (see Scope / Verify / AC) in the same PR that creates the regression window — it does not hand this one attribute to a downstream sub-issue.
- **Relocating the `.map-context-strip`** (lede/filter-sentence/freshness) to a floating card + index.html inline-CSS / map-lede-CLS re-baseline. → **O3**.
- **Hoisting skip-link + FamilyLegend to a persistent App-root** so they don't unmount with the view. → **O2**.
- **Named z-index token scale itself** (`--z-chrome`, `--z-modal`, etc.) — defined by **S1**; this PR only consumes the `--z-chrome` token.
- **Feed archival / route shim.** → **F1** (the `state.view === 'feed'` branch stays present and functional here).
- **Error-as-overlay.** → **O7**. The App-level error path is a pre-shell early-return (`App.tsx:554-562` renders a bare `StatusBlock` — `status-block__*` classes, not `.error-screen` — and returns *before* `.app`/`AppHeader`/`#main-surface` mount), so S2 neither moves it nor clears for it; it is unchanged. (Note: `.error-screen` — `styles.css:94-98` — is consumed only inside `MapSurface.tsx:225` and by the `ErrorBoundary.tsx:39` fallback, never as a non-map `#main-surface` child. The in-map `MapSurface` `.error-screen` rides along inside the extracted map layer; reframing *that* in-map error as a viewport overlay is also O7, not S2.)

## Acceptance criteria

- On a scoped landing at 1440×900, the MapLibre canvas fills `100vw × 100vh` (verify via `getBoundingClientRect()` on `.map-surface` ≈ viewport, with **no 16px gutters**), with the `AppHeader` floating over the top edge — the canvas is **not** inset by `#main-surface` padding and is not capped by the header's flow height. This is a **steady-state** assertion taken after `waitForAppReady` settles; it proves the box reaches full-bleed and does not encode any claim about *which* call sized the canvas (the corrective `map.resize()` is S3's — see Out of scope — so S2 asserts only the settled geometry, never a resize-causality claim).
- The `{mapVisible && …}` map block renders inside a **`#map-layer` wrapper that is a sibling of `<main>`**, not a descendant — confirmed by DOM structure (`#map-layer` / `.map-surface` is not contained within `#main-surface`).
- **On the unscoped scrim landing, `#map-layer` still mounts (idle, behind S1's inert scrim) and `/api/observations` stays at zero.** Verify the always-mounted invariant: with no scope (the S1 scrim landing), `#map-layer` is present in the DOM (the map is mounted, not unmounted), the `scopeActive` fetch gate (`App.tsx:134`) holds net `/api/observations` requests at **0**, and `#map-layer` is **not** gated away by `scopeActive`. This AC fails if S2 makes the wrapper conditional on `scopeActive` (which would silently break the always-mounted-under-scrim contract).
- **`.scope-control` anchors to the viewport after the hoist.** Because ScopeControl is a sibling of `MapSurface` (not a child of `.map-surface`), its `position: absolute` offsets must resolve against the fixed `#map-layer` wrapper, which equals the viewport. Verify at 1440×900 (scoped landing) that `.scope-control` `getBoundingClientRect()` sits at the documented `--space-md` inset from the **viewport** top/sides (its `top` ≈ `--space-md`, not `--space-md + headerHeight`, and it is centered via `margin-inline:auto` within the viewport-width wrapper) — i.e. it did not collapse onto `#main-surface`'s box or shift under the header. This AC fails if the wrapper is omitted (the original review gap).
- `#main-surface` still exists with `id="main-surface"`, `data-render-complete`, `mainRef`, and `tabIndex={0}` intact.
- **The Map tab's `aria-controls` is retargeted to `"map-layer"` and resolves to the hoisted `#map-layer`, not the now-feed-only `#main-surface`.** Verify (unit and/or a11y e2e) that `getByRole('tab', { name: 'Map view' })` carries `aria-controls="map-layer"` and that `document.getElementById('map-layer')` is present — i.e. the single tab points at the region that actually contains the map post-hoist. This AC **fails** if the tab is left with `aria-controls="main-surface"` (which would resolve to a present element but no longer the controlled map region — the silent a11y semantic regression this hoist would otherwise introduce). The broader inert/aria-busy retarget on `#map-layer` is O1 and is not asserted here; only the **tab's** `aria-controls` is in S2's scope.
- The detail rail/sheet (`App.tsx:690-710`) continue to float over the live, interactive map (no regression to #663); the map stays mounted on scope/detail transitions.
- The live non-map view surface that `#main-surface` actually wraps — the `FeedSurface` branch — does not render underneath the fixed header: `#main-surface` `padding-top`/`scroll-margin-top` clears the header height (R15). (The App-level error path is excluded by construction — it early-returns before any header exists — so there is no error-surface-under-header AC to assert in S2.)
- **The open `.filters-panel` is not occluded by the fixed header on the map view** (R15, second surface). Open the filters panel from the header on a scoped map landing and verify via `getBoundingClientRect()` that the panel's top edge (and its close button at `styles.css:1667-1690`) clears the fixed-header height — the panel's effective top is ≥ the header's bottom, and the close button is fully within the visible region, at both 390×844 (mobile) and 1440×900 (desktop). This is the interim clearance; the full floating-sheet conversion is O4.
- `AppHeader` is `position: fixed` on `--z-chrome`, layered **above** legend/scope/popover (`--z-panel`/`--z-panel`+1) and **below** the detail rail (`calc(--z-panel + 5)`=45) — so the in-place rail (#663) still floats over the header. **S2 asserts no z-order relationship between the fixed header and the bottom-sheet snap tiers** (that 10/15/20 stack reconciliation is the deferred mobile-overlay workstream); the only requirement here is that the header gains its `--z-chrome` rank without introducing a regression to the rail or the map-assist overlays.
- **This PR alone leaves all four CI checks — `test`, `lint`, `build`, `e2e` — and `knip` green. No "leave CI red, fix next PR."** Every coupled spec that asserts the old windowed DOM (map inside `#main-surface`, header sticky) is re-baselined **on this branch**. If extraction surfaces a knip finding (e.g. an unused export from the move), it is addressed on the same branch (add a `knip.ts` ignore rule with the dated-comment convention if it is a genuine static-analysis false positive, otherwise remove the dead code).
- Zero console errors and zero console warnings at all five canonical viewports × both themes (the documented upstream basemap worker warning at zoom ≥ 10 is the single tolerated line and must be filtered, not failed on).

## Test plan

This PR depends on **S1** having already landed (a) the inert-scrim-over-mounted-map conversion (chooser is a scrim, the map is always mounted) and (b) the POM readiness-gate resilience (the `waitForAppReady` decoupling from the `main` tag and `#main-surface` selector hardening) — confirm both are on `main` before starting, so the suite does not go red en masse when the map leaves `<main>`.

**Re-baseline (same branch as the behavior change):**
- `frontend/e2e/pages/app-page.ts` — `mainSurface = page.locator('main#main-surface')` (line 54) and `waitForAppReady` keying on `main[data-render-complete="true"]` (lines 148-152). The `<main>` element survives, so these should keep resolving; verify the map is no longer queried *through* `mainSurface` anywhere. Add/confirm a `mapLayer` accessor pointing at `#map-layer` and/or a `mapSurface` accessor at `.map-surface` for tests that assert the map's viewport geometry, plus a `scopeControl` accessor for the new anchoring assertion.
- `frontend/e2e/a11y.spec.ts` — any assertion that the map is a descendant of `#main-surface`, plus tab-order order across header → map controls now that the header is fixed chrome. The `aria-busy` poll on `main#main-surface` stays valid (busy is still vacuously false on the map view here; the real busy retarget is O1).
- `frontend/e2e/a11y-bundle-d.spec.ts` — landmark / `#main-surface` couplings.
- `frontend/e2e/sheet-snap.spec.ts` — sheet snaps already measure `window.innerHeight`, so snap math is unaffected; re-confirm the sheet still floats over the now-fixed map and that `inert` on `#main-surface` still mutes the (now near-empty) `<main>`. The full-snap-sheet-vs-fixed-header z-order is a **known deferred follow-up** (mobile-overlay workstream — see Out of scope): assert only the **existing** sheet behavior here, do **not** add an assertion that the sheet occludes/inerts the header (it will not, once the header is `--z-chrome`), and do not change `inert` targeting in S2.
- `frontend/e2e/family-legend-collapse-visibility.spec.ts` — the #471 clip-guard asserts the legend chip rect sits inside the `#main-surface` rect. The map (and legend) no longer live inside `#main-surface`; re-target that rect assertion to the **viewport** (the inversion removes the clip parent by design). This is one of the substantive behavior re-baselines in S2.
- `frontend/e2e/error-states.spec.ts` and `frontend/e2e/axe.spec.ts` — re-point any `#main-surface`-scoped chrome scans. The App-level error path renders a pre-shell `StatusBlock` (no `AppHeader`, no `#main-surface`) and is unchanged by S2; confirm these specs key on `status-block__*` selectors (or `[role="status"]`), not on a `#main-surface` ancestor, and that nothing here newly assumes a fixed header on the error path.
- Any spec that opens the filters panel (search `frontend/e2e/` for `filters-panel` / `onOpenFilters` / the header filters trigger) — if one asserts the panel's position relative to the old sticky header, re-baseline it to the fixed-header clearance. Add the geometry assertion below if none exists.
- `frontend/e2e/prod-smoke.preview.spec.ts` couples to `main#main-surface` aria-busy; note this preview job runs **outside** the Mergify queue gate, so its breakage is silent — re-point its selectors on this branch even though it won't block the queue.

**Add:**
- An e2e assertion (extend `frontend/e2e/a11y.spec.ts` or a small new geometry check) that `.map-surface` `getBoundingClientRect()` ≈ the viewport (within a px tolerance, **no 16px gutters**) on a scoped landing at the desktop viewport, taken after `waitForAppReady` — the load-bearing **steady-state** proof the map went full-bleed. This e2e check asserts geometry only; it deliberately makes **no** claim about which call sized the canvas (a settled rect cannot distinguish a first-paint full-bleed from react-map-gl's own container observer from S3's corrective `map.resize()`).
- An e2e assertion that the map is **mounted on the unscoped scrim landing**: with no scope, `#map-layer` is present in the DOM and net `/api/observations` requests = 0 (existing `scopeActive` gate) — the falsifiable proof that the hoist preserved S1's always-mounted-under-scrim invariant (`#map-layer` is not gated on `scopeActive`).
- An e2e assertion that `.scope-control` `getBoundingClientRect()` anchors to the viewport (top ≈ `--space-md`, horizontally centered, **not** offset by the header height or collapsed onto `#main-surface`) on a scoped landing at 1440×900 — the falsifiable proof the mandatory `#map-layer` wrapper gives ScopeControl the right containing block (the original review gap).
- An e2e assertion that the **open filters panel clears the fixed header** on a scoped map landing at 390×844 and 1440×900 (panel top edge and close button below the header bottom) — the R15 second-surface proof.

**Unit:** `frontend/src/App.test.tsx` — assert the map block (`#map-layer` containing `MapSurface`/`ScopeControl`) renders as a **sibling of `<main>`, not a descendant** (DOM structure the existing `MapSurface` stub — a bare `<div data-testid="map-surface-stub" />`, `App.test.tsx:101-113` — can satisfy: query `#map-layer` and assert it is not contained within `#main-surface`), and that `.map-surface`/`#map-layer` is not nested inside `#main-surface`. Assert the new tab→region contract: `getByRole('tab', { name: 'Map view' })` carries `aria-controls="map-layer"` (not `"main-surface"`) and `#map-layer` is present. **No corrective-`map.resize()` unit assertion in S2** — the stub exposes no `getMap`/`resize` handle (the real `mapRef` is inside lazy `MapCanvas`, `MapCanvas.tsx:482`, which never mounts in jsdom), and the resize is S3's responsibility; the falsifiable proof S2 owns is the DOM-structure check above, not a resize spy. No DB mocks; no new `page.route` writes (the no-write grep must stay clean).

**Re-baseline (Unit) — `App.test.tsx:1169-1180` specifically.** The existing test *'an unscoped URL with a detail code still shows the chooser (detail is not a scope)'* (`App.test.tsx:1169-1180`) asserts on an unscoped render that `screen.queryByTestId('map-surface-stub')` is **`null`** (line 1178) — the *pre-S1* "chooser renders **in place of** the map" behavior. Post-S1 the chooser is an **inert scrim over a mounted, idle map**, so the map stub is **present** on an unscoped render — directly contradicting the always-mounted invariant S2's Unit section asserts. **Ownership of the flip:** the `toBeNull()` → present flip is **S1's** to land (S1 is the PR that converts unscoped from early-return to scrim-over-mounted-map, so it is the change that makes the stub appear); S1's re-baseline list must name `App.test.tsx:1169-1180` and replace line 1178's `expect(screen.queryByTestId('map-surface-stub')).toBeNull()` with an assertion that **both** the chooser scrim (`findByRole('region', { name: /Choose where to look at birds/i })`) **and** `map-surface-stub` / `#map-layer` are present (the detail-is-not-a-scope contract is now expressed as "scrim shown over the mounted map", not "map absent"). If S1 ships without flipping this test, S2 inherits a **red unit test on its own branch** the moment it relies on `#map-layer` being mounted unscoped — violating the "this PR alone leaves all four checks green" AC. **S2's branch-gate confirmation:** before relying on the always-mounted-unscoped behavior, confirm `App.test.tsx:1169-1180` already asserts the *scrim-over-mounted-map* shape on `main` (S1 landed it); if for any reason S1 left the `toBeNull()` expectation in place, S2 must flip it on its own branch (same assertion: scrim **and** `#map-layer`/stub present) rather than ship the contradiction. The same assertion must not be owned by both PRs — it is S1's by default, S2's only as a fallback if S1 missed it.

## Dependencies

Part of #761.

- **Depends on S1** (convert ScopeChooser to inert modal scrim over the persistent map + gate detail render on `scopeActive` + the POM readiness-gate / `#main-surface` selector-resilience that must merge before any shell PR). S1 also defines the named z-index token scale this PR's `--z-chrome` consumes, and establishes the always-mounted-map / `scopeActive`-fetch-gate model S2's hoist must preserve. **S1's re-baseline list must include `App.test.tsx:1169-1180`** — converting the unscoped render from "map stub absent (`toBeNull`)" to "scrim shown over a mounted map stub", because S1 is the change that makes the stub appear on the unscoped landing. If S1 lands without this flip, S2 inherits a red unit test on its branch (see Test plan → Re-baseline (Unit)); S2 confirms the flip is on `main` before relying on the always-mounted-unscoped invariant, and flips it on its own branch only as a fallback.
- The epic (#761) holds the authoritative numbered dependency graph; consult it for the full Phase 0→5 sequencing. S2 is the head of Phase 3's shell-inversion chain (S2 → {S3, S4, O1…O7}).

## Plan traceability

- Research report: `docs/analyses/2026-05-29-map-first-rearchitecture-761/report.md` — **§6 / WS1 (App-shell inversion, sub-issue 1.2)** + **WS2 (Full-bleed map container, sub-issue 2.1)**, **§10 sub-issue S2**, and **risk register R15** (floating header removes flow height → in-flow chrome surfaces slide under it: `#main-surface`'s live non-map view **and** the open `.filters-panel`). The §6 sheet-z-stack item ("z-index: 10/15/20 for peek/half/full … Guarantee sheet z-index ≥ map controls + popovers without a z-index war") is the **mobile-overlay workstream**, which S2 explicitly defers (see Out of scope). The `map.resize` on the flex→fixed transition is assigned to **S3** by the report's own sequencing (`report.md:251` "S3. … + `map.resize` on transition (WS2.2 + gap 8)") and is listed as unhandled gap 8 owned by WS2 (`report.md:222`) — S2 does not pull it in. Confirmed anchors: `.app` flex column (`styles.css:77-81`), `#main-surface` flex/padding (`styles.css:86-93`), `.map-surface` flex child (`styles.css:686-691`), `.app-header` sticky z-10 (`styles.css:164-174`), the in-`<main>` map block with sibling ScopeControl/MapSurface (`App.tsx:633-679`, 641 vs 649), `.filters-panel` `position: relative` sibling-of-`<main>` (`App.tsx:581-600`, `styles.css:1663-1665`), `.scope-control` absolute (`styles.css:1903-1920`), rail/sheet fixed siblings (`App.tsx:690-710`), sheet snap tiers (`styles.css:1132`/`1148`/`1303`/`1311`), the unscoped early-return that S1 converts to a scrim (`App.tsx:541-552`), the `scopeActive` fetch gate (`App.tsx:134`) and `mapVisible` view gate (`App.tsx:273`), the lazy-`MapCanvas` boundary + `mapRef` (`MapSurface.tsx:16-17`, `MapCanvas.tsx:482`), the `MapSurfaceProps` signature with no ref (`MapSurface.tsx:49`/`:180-202`), the pre-shell error early-return (`App.tsx:554-562` → `StatusBlock`, `status-block__*` classes per `StatusBlock.tsx:60-72`), and the in-map `.error-screen` (`styles.css:94-98`, consumed by `MapSurface.tsx:225` / `ErrorBoundary.tsx:39`).
- Epic workstream: #761 → **"1. App-shell inversion & routing"** and **"2. Full-bleed map container"**.

This PR changes visible layout under `frontend/**`, so the implementer must, before opening the PR:

1. Run `npm run dev --workspace @bird-watch/frontend` and drive the change **live via Playwright MCP** at all **five canonical viewports** — 390×844, 768×1024, 1024×768, 1440×900, 1920×1080 — × **both themes** (dark via `document.documentElement.setAttribute('data-theme','dark')` through `browser_evaluate`; do **not** use `prefers-color-scheme` emulation — the `[data-theme]` attribute convention overrides it).
2. Confirm the map is genuinely full-bleed (no gutters, header floating over it), the floating header does not occlude markers or overlap ScopeControl illegibly, **the ScopeControl bar sits at its `--space-md` inset from the viewport top (not under the header)**, **the open filters panel is fully visible and not clipped by the fixed header**, the detail rail/sheet still float correctly, and the live `FeedSurface` view doesn't slide under the header. Also confirm the **unscoped scrim landing** still shows the map mounted behind the scrim (idle, no `/api/observations` fired).
3. `browser_console_messages` must return **zero errors and zero warnings** at every viewport/theme (the single documented upstream basemap worker warning at zoom ≥ 10 is the only tolerated line and must be filtered).
4. Capture screenshots — 5 viewports × 2 themes per touched surface (≥10 captures) — into the PR via the **`pr-screenshots-via-user-attachments`** paste flow → `user-attachments/assets/<uuid>` URLs. Include at least one capture of the **open filters panel** on the map view to evidence the R15 clearance. **Never commit PNGs to the repo.** Verify the PR body has ≥10 attachment URLs: `gh pr view <N> --repo julianken/bird-sight-system --json body --jq '.body | [scan("user-attachments/assets/[a-f0-9-]++")] | length'`.
5. Per the "verify zoomed-in near borders" rule, also capture at z14-16 near a state border (e.g. AZ edge) in both themes to confirm the full-bleed reframe didn't introduce edge-clipping or chrome-occlusion regressions.
6. Obtain a **design-review PASS** from the `ui-design:ui-designer` agent dispatched via the Task tool at **`model: "opus"`** (explicit override — the agent's frontmatter is `model: inherit`), briefed with the PR URL, the design-intent reference (the #760 state-artboard mockup + this report section), all screenshot URLs, and the AC reference. **All five viewports must PASS** before bot review, critic review, and the Mergify queue.

julianken:
## Diagrams

DOM/stacking inversion — the map moves from a windowed flex child of the padded `<main>` to the viewport-filling root, with the chrome floating on top.

```mermaid
graph TB
    subgraph after["AFTER — #761 S2 (map-first root)"]
        A2[".app — flex column"]
        A2 --> H2["AppHeader<br/>position: fixed · --z-chrome 42<br/>(floats over the map)"]
        A2 --> ML["#map-layer (NEW wrapper)<br/>position: fixed · inset: 0 · --z-map 0"]
        ML --> SC2["ScopeControl<br/>absolute · top = header + space-md"]
        ML --> MS2[".map-surface absolute inset:0<br/>(fills the viewport)"]
        A2 --> MN2["&lt;main #main-surface&gt;<br/>empty · padding-top clears header<br/>(readiness gate + inert target)"]
        A2 --> RAIL2["detail rail / sheet<br/>--z-rail 43 (floats over header)"]
        A2 --> SCRIM2["chooser scrim --z-modal 50"]
    end

    subgraph before["BEFORE (windowed map)"]
        A1[".app — flex column"]
        A1 --> H1["AppHeader<br/>position: sticky · z-index 10"]
        A1 --> MN1["&lt;main #main-surface&gt;<br/>flex:1 · padding 16px (caps the map)"]
        MN1 --> SC1["ScopeControl (absolute)"]
        MN1 --> MS1[".map-surface flex:1 (windowed)"]
        A1 --> RAIL1["detail rail / sheet (fixed)"]
    end
```

## Summary

- **Why:** the map was *windowed inside the UI* (a padded, header-capped flex child), not the application itself. #761 (S2) inverts the shell so the map is the viewport-filling **root** and the chrome floats over it (Google-Maps-style) — the foundation the rest of the map-first epic builds on.
- **How:** DOM extraction, not conditional padding. The `{mapVisible && …}` block (ScopeControl + MapSurface) is hoisted out of `<main>` into a mandatory `<div id="map-layer">` (`position: fixed; inset: 0`) rendered as a **sibling** of `<main>`. The wrapper is load-bearing — ScopeControl is a sibling of MapSurface, so its `position: absolute` resolves against the fixed wrapper (≡ viewport) and keeps anchoring to the viewport top edge. `.app-header` becomes `position: fixed` and adopts the P1-defined-but-unapplied `--z-chrome` (42): above the map-assist overlays (40/41), below the detail rail (43).
- **Occlusion handled (R15 + the P1-deferred regression):** the now-floating header would intercept clicks on the top-anchored `.scope-control` (the exact regression P1 surfaced) — fixed by offsetting the control's top by `--header-height` so it clears the header band (state-scope whole-US-reset stays clickable). `#main-surface` gains a `padding-top`/`scroll-margin-top` header clearance for any future non-map surface; the open `.filters-panel` gains the same offset **plus** a positioned z-index in the map-assist band so its controls aren't buried behind the full-bleed map (interim — O4 makes it a floating sheet). The Map tab's `aria-controls` is retargeted `"main-surface"` → `"map-layer"`. Invariants preserved: `#map-layer` is gated only by `mapVisible` (NOT `scopeActive`), so the unscoped scrim landing still mounts the idle map and the `scopeActive` fetch gate holds `/api/observations` at **zero**; `<main>` keeps its `id`/`data-render-complete`/`mainRef`/`tabIndex`/`inert` target (inert retarget to `#map-layer` is O1).

### Full-viewport map root (scoped, AZ)

The headline S2 result: the map is the viewport-filling **root**, with the header + scope-control floating over it. All 5 canonical viewports × both themes.

| Viewport | Light | Dark |
|---|---|---|
| 390×844 (mobile) | scoped AZ 390×844 light | scoped AZ 390×844 dark |
| 768×1024 (tablet portrait) | scoped AZ 768×1024 light | scoped AZ 768×1024 dark |
| 1024×768 (tablet landscape) | scoped AZ 1024×768 light | scoped AZ 1024×768 dark |
| 1440×900 (desktop) | scoped AZ 1440×900 light | scoped AZ 1440×900 dark |
| 1920×1080 (wide) | scoped AZ 1920×1080 light | scoped AZ 1920×1080 dark |

### Unscoped landing (scrim over full-bleed map)

The chooser scrim over the full-bleed **idle** map (no scope yet → zero `/api/observations`). 1440×900, both themes.

| Light | Dark |
|---|---|
| unscoped landing 1440×900 light | unscoped landing 1440×900 dark |

### Floating-chrome interactions & border zoom

Mobile full-snap detail sheet — the collapse handle stays reachable under the floating header (the `body:has(.species-detail-sheet--full)` click-through). Border zoom (AZ) confirms no interior clipping at the artboard mask edge, light + dark.

| Mobile full-snap sheet (390, light) | AZ border zoom (1440, light) | AZ border zoom (1440, dark) |
|---|---|---|
| mobile full-snap sheet 390 light | AZ border zoom 1440 light | AZ border zoom 1440 dark |

_Live Playwright MCP verification PASSED — 0 console errors at all 5 viewports × both themes; scope-control non-occluded; mobile full-snap collapse-handle guard works; border zoom shows no interior clipping (only pre-existing OpenFreeMap basemap sprite warnings)._

- [x] Every new/renamed `className` in this PR has a matching CSS rule — `scripts/check-orphan-classnames.sh` → PASS. (This PR adds no new JSX `className`; it modifies existing selectors and adds the `#map-layer` id-selector rule. `orphan-classname-check`: PASS.)
- [ ] Multi-viewport design QA: PR body has ≥10 `user-attachments` URLs — PENDING (orchestrator captures the live drive at all 5 canonical viewports × 2 themes per #445).

## Test plan

- [x] `npm test --workspace @bird-watch/frontend` — green (65 files, **965** unit tests). New unit assertions: map renders in `#map-layer` as a **sibling** of `<main>` (not a `#main-surface` descendant); Map tab `aria-controls="map-layer"`; `#map-layer` mounts on the unscoped scrim landing with zero `/api/observations`.
- [x] New unit / integration tests added — `App.test.tsx` (S2 DOM-structure + tab-region + always-mounted-unscoped describe), `AppHeader.test.tsx` (`aria-controls="map-layer"`).
- [x] New Playwright e2e spec added — `e2e/map-root-geometry.spec.ts`: full-bleed `.map-surface` ≈ viewport (no 16px gutters); `.scope-control` viewport anchor clearing the fixed header; open filters-panel header clearance at 390×844 + 1440×900; `#map-layer` mounted idle + zero observations on the unscoped landing. Re-baselined the #471 family-legend clip-guard from `#main-surface` → the **viewport** (the map left `#main-surface` by design); routed attribution-modal/axe modal opens through the header affordance (the standalone `.attribution-trigger` now sits behind the full-bleed map). Full local e2e: **223 passed / 14 WebGL-skipped**, single-worker (CI runs SwiftShader for the skipped set).
- [x] `npm run build --workspace @bird-watch/frontend` — clean. `npm run lint` — clean. `npx knip` — clean. No-DB-writes grep — clean.
- [ ] (UI only) Playwright MCP smoke — PENDING: the orchestrator runs the live capture (this subagent did not use Playwright MCP per the dispatch). Reviewers repeat the drive + console check via `gh pr checkout` against the head SHA.

## Plan reference

Part of #761 (map-first re-architecture), sub-issue **S2** — head of Phase 3's shell-inversion chain (S2 → {S3, S4, O1…O7}). Research report: `docs/analyses/2026-05-29-map-first-rearchitecture-761/report.md` §6 / WS1 (App-shell inversion) + WS2 (Full-bleed map container), §10 sub-issue S2, risk register R15. Closes #775.

Out of scope (deferred to the named downstream sub-issue): corrective `map.resize()` on the flex→fixed transition + asymmetric `fitBounds` top-padding → **S3** (#773); `inert`/`aria-busy`/`aria-live` retarget onto `#map-layer` → **O1**; `.filters-panel` → floating sheet → **O4**; full-snap sheet z-order reconcile → **O5** (#783) — S2 only keeps the sheet's collapse handle reachable via a `body:has(.species-detail-sheet--full)` click-through on the header, no z-index war.

---

### DIFF
--- a/frontend/e2e/attribution-modal.spec.ts
+++ b/frontend/e2e/attribution-modal.spec.ts
@@ -68,8 +68,12 @@ test.describe('AttributionModal — open / close (desktop)', () => {
     const app = new AppPage(page);
     await app.goto('scope=us');
     await app.waitForAppReady();
-    const trigger = page.locator('button.attribution-trigger');
-    await trigger.click();
+    // #761 (S2): open via the AppHeader "Attribution" button (the live affordance,
+    // fixed chrome on --z-chrome). The standalone `.attribution-trigger` shim now
+    // sits in the `.app` flow BEHIND the full-viewport `#map-layer`, so a direct
+    // pointer click on it is intercepted by the map. The header fires the same
+    // onOpenAttribution → programmatic `.attribution-trigger.click()` shim.
+    await app.attributionTrigger.click();
     const dialog = page.locator('dialog.attribution-modal');
     await expect(dialog).toHaveAttribute('open', '');
     // The dialog has an h2 title.
@@ -80,7 +84,7 @@ test.describe('AttributionModal — open / close (desktop)', () => {
     const app = new AppPage(page);
     await app.goto('scope=us');
     await app.waitForAppReady();
-    await page.locator('button.attribution-trigger').click();
+    await app.attributionTrigger.click();
     const dialog = page.locator('dialog.attribution-modal');
     await expect(dialog.getByRole('heading', { level: 3, name: /bird sightings data/i })).toBeVisible();
     await expect(dialog.getByRole('heading', { level: 3, name: /family silhouettes/i })).toBeVisible();
@@ -91,7 +95,7 @@ test.describe('AttributionModal — open / close (desktop)', () => {
     const app = new AppPage(page);
     await app.goto('scope=us');
     await app.waitForAppReady();
-    await page.locator('button.attribution-trigger').click();
+    await app.attributionTrigger.click();
     const dialog = page.locator('dialog.attribution-modal');
     // Wait for the seeded Phylopic data to land (it arrives via the
     // /api/silhouettes fetch). At least one row with a creator must
@@ -118,7 +122,7 @@ test.describe('AttributionModal — open / close (desktop)', () => {
     const app = new AppPage(page);
     await app.goto('scope=us');
     await app.waitForAppReady();
-    await page.locator('button.attribution-trigger').click();
+    await app.attributionTrigger.click();
     const dialog = page.locator('dialog.attribution-modal');
     await expect(dialog.getByRole('heading', { level: 3, name: /bird sightings data/i })).toBeVisible();
     // Wait for any phylopic data to land so the assertion sees the full
@@ -135,40 +139,41 @@ test.describe('AttributionModal — open / close (desktop)', () => {
     }
   });
 
-  test('Escape closes the dialog and returns focus to the Credits trigger', async ({ page }) => {
+  test('Escape closes the dialog and returns focus to the opener', async ({ page }) => {
     const app = new AppPage(page);
     await app.goto('scope=us');
     await app.waitForAppReady();
-    const trigger = page.locator('button.attribution-trigger');
-    await trigger.click();
+    // #761 (S2): open via the header affordance (the standalone trigger is behind
+    // the full-bleed map). The modal restores focus to `document.activeElement` at
+    // open time — now the header "Attribution" button — so focus-return is asserted
+    // on that button (the live opener), not the occluded `.attribution-trigger`.
+    await app.attributionTrigger.click();
     const dialog = page.locator('dialog.attribution-modal');
     await expect(dialog).toHaveAttribute('open', '');
 
     await page.keyboard.press('Escape');
-    // After Escape, the dialog closes and focus returns to the trigger.
+    // After Escape, the dialog closes and focus returns to the opener.
     await expect(dialog).not.toHaveAttribute('open', '');
-    await expect(trigger).toBeFocused();
+    await expect(app.attributionTrigger).toBeFocused();
   });
 
-  test('close button closes the dialog and returns focus to the Credits trigger', async ({ page }) => {
+  test('close button closes the dialog and returns focus to the opener', async ({ page }) => {
     const app = new AppPage(page);
     await app.goto('scope=us');
     await app.waitForAppReady();
-    const trigger = page.locator('button.attribution-trigger');
-    await trigger.click();
+    await app.attributionTrigger.click();
     const dialog = page.locator('dialog.attribution-modal');
     const close = dialog.getByRole('button', { name: /close/i });
     await close.click();
     await expect(dialog).not.toHaveAttribute('open', '');
-    await expect(trigger).toBeFocused();
+    await expect(app.attributionTrigger).toBeFocused();
   });
 
   test('opening the dialog moves focus into the modal (close button is autofocused)', async ({ page }) => {
     const app = new AppPage(page);
     await app.goto('scope=us');
     await app.waitForAppReady();
-    const trigger = page.locator('button.attribution-trigger');
-    await trigger.click();
+    await app.attributionTrigger.click();
     const dialog = page.locator('dialog.attribution-modal');
     const close = dialog.getByRole('button', { name: /close/i });
     await expect(close).toBeFocused();
@@ -191,20 +196,21 @@ test.describe('AttributionModal — mobile viewport', () => {
     const app = new AppPage(page);
     await app.goto('scope=us');
     await app.waitForAppReady();
-    const trigger = page.locator('button.attribution-trigger');
-    await trigger.click();
+    // #761 (S2): open via the header affordance; focus returns to that opener
+    // (the standalone trigger is behind the full-bleed map — see desktop tests).
+    await app.attributionTrigger.click();
     const dialog = page.locator('dialog.attribution-modal');
     await expect(dialog).toHaveAttribute('open', '');
     await page.keyboard.press('Escape');
     await expect(dialog).not.toHaveAttribute('open', '');
-    await expect(trigger).toBeFocused();
+    await expect(app.attributionTrigger).toBeFocused();
   });
 
   test('Phylopic section renders at least one silhouette (mobile)', async ({ page }) => {
     const app = new AppPage(page);
     await app.goto('scope=us');
     await app.waitForAppReady();
-    await page.locator('button.attribution-trigger').click();
+    await app.attributionTrigger.click();
     const dialog = page.locator('dialog.attribution-modal');
     const rows = dialog.locator('[data-testid=attribution-phylopic-row]');
     await expect(rows.first()).toBeVisible({ timeout: 10_000 });
--- a/frontend/e2e/axe.spec.ts
+++ b/frontend/e2e/axe.spec.ts
@@ -394,9 +394,14 @@ test.describe('axe-core WCAG scans', () => {
     const app = new AppPage(page);
     await app.goto('view=map');
     await app.waitForAppReady();
-    // Phase 6: footer removed. Use .attribution-trigger (AttributionModal's
-    // own button) directly — no footer scoping needed.
-    await page.locator('button.attribution-trigger').click();
+    // #761 (S2): open via the AppHeader "Attribution" button — the documented
+    // real affordance, which fires onOpenAttribution → programmatic click on the
+    // hidden `.attribution-trigger` shim. The header is fixed chrome on
+    // `--z-chrome` (always clickable); the standalone `.attribution-trigger` now
+    // sits in the `.app` flow BEHIND the full-viewport `#map-layer`, so a direct
+    // POINTER click on it is intercepted by the maplibre attribution control. The
+    // header path is occlusion-immune (programmatic .click()).
+    await app.attributionTrigger.click();
     // Wait on the [open] attribute commit — observable contract that
     // showModal() has run, focus-delegation is settled, and the dialog
     // is in the top layer.
@@ -418,8 +423,10 @@ test.describe('axe-core WCAG scans', () => {
       const app = new AppPage(page);
       await app.goto('view=map');
       await app.waitForAppReady();
-      // Phase 6: footer removed. Use .attribution-trigger directly.
-      await page.locator('button.attribution-trigger').click();
+      // #761 (S2): open via the AppHeader "Attribution" button (occlusion-immune
+      // programmatic shim) — the standalone `.attribution-trigger` now sits behind
+      // the full-viewport `#map-layer`. See the desktop test for the rationale.
+      await app.attributionTrigger.click();
       await expect(page.locator('dialog.attribution-modal[open]')).toBeVisible();
       const results = await new AxeBuilder({ page }).withTags(WCAG_TAGS).analyze();
       if (results.violations.length) {
--- a/frontend/e2e/family-legend-collapse-visibility.spec.ts
+++ b/frontend/e2e/family-legend-collapse-visibility.spec.ts
@@ -3,7 +3,7 @@ import { AppPage } from './pages/app-page.js';
 import type { Observation } from '@bird-watch/shared-types';
 
 /**
- * Regression: collapsed family legend chip clipped by #main-surface overflow.
+ * Regression: collapsed family legend chip clipped by its scroll/overflow parent.
  *
  * PR #471 (W2) added `.map-context-strip { padding: var(--space-md)
  * var(--space-lg) }` (24 px vertical) + a 1px border-bottom. The combined
@@ -11,9 +11,14 @@ import type { Observation } from '@bird-watch/shared-types';
  * area, causing `#main-surface`'s overflow:auto scroll container to clip the
  * `position:absolute; bottom:12px` collapsed chip entirely.
  *
- * Fix (this PR): `#main-surface` becomes a flex column; `.map-surface` uses
- * `flex:1; min-height:0` instead of `height:100%`. The chip now sits inside
- * the flex-sized `.map-surface`, never below the clip boundary.
+ * #761 (S2) re-baseline: the map left `#main-surface` entirely — the shell
+ * inverted so the map is the viewport ROOT (`#map-layer { position: fixed;
+ * inset: 0 }`, a sibling of `<main>`). The collapsed chip now lives inside the
+ * viewport-filling `.map-surface`, NOT inside `#main-surface`. The clip parent
+ * the original #471 guard checked against is gone by design, so this guard now
+ * asserts the chip is fully inside the VIEWPORT (the new bounding box). The
+ * invariant the test protects is unchanged: the collapsed chip must be fully
+ * visible, never clipped by its container.
  *
  * This test is the regression guard #471 should have had.
  *
@@ -130,57 +135,56 @@ async function skipIfMapHookAbsent(
 }
 
 /**
- * Assert that the collapsed chip's bounding rect is fully inside
- * main#main-surface's bounding rect. Both edges (bottom and left) are
- * checked to within a 1px tolerance (fractional sub-pixel rounding).
+ * Assert that the collapsed chip's bounding rect is fully inside the VIEWPORT.
+ *
+ * #761 (S2): the map is now the viewport ROOT (`#map-layer { position: fixed;
+ * inset: 0 }`), so the chip's containing/clip parent is the viewport, not
+ * `#main-surface` (which the map left). Both edges (bottom and left) are checked
+ * to within a 1px tolerance (fractional sub-pixel rounding).
  */
-async function assertChipInsideMain(
+async function assertChipInsideViewport(
   page: import('@playwright/test').Page,
-): Promise<{ chipBottom: number; mainBottom: number; chipLeft: number; mainLeft: number }> {
+): Promise<{ chipBottom: number; viewportBottom: number; chipLeft: number; viewportLeft: number }> {
   const rects = await page.evaluate(() => {
-    // Browser context: select the readiness surface on the tag-AND-id-free
-    // `[data-render-complete]` attribute (the hook the map-first inversion
-    // (#761) carries forward), which resolves to the same single element as
-    // `#main-surface` today. A Playwright Locator can't cross into evaluate.
-    const main = document.querySelector<HTMLElement>('[data-render-complete]');
     // The collapsed chip is the toggle button with aria-expanded="false" inside .map-surface
     const chip = document.querySelector<HTMLElement>(
       '.map-surface button[aria-expanded="false"]',
     );
-    if (!main || !chip) return null;
-    const mainRect = main.getBoundingClientRect();
+    if (!chip) return null;
     const chipRect = chip.getBoundingClientRect();
     return {
       chipBottom: chipRect.bottom,
-      mainBottom: mainRect.bottom,
+      // #761 (S2): the map fills the viewport, so the bottom/left bound is the
+      // viewport edge (the map left #main-surface — the old clip parent is gone).
+      viewportBottom: window.innerHeight,
       chipLeft: chipRect.left,
-      mainLeft: mainRect.left,
+      viewportLeft: 0,
     };
   });
 
-  expect(rects, 'main#main-surface and/or collapsed chip button not found in DOM').not.toBeNull();
-  const { chipBottom, mainBottom, chipLeft, mainLeft } = rects!;
+  expect(rects, 'collapsed chip button not found in DOM').not.toBeNull();
+  const { chipBottom, viewportBottom, chipLeft, viewportLeft } = rects!;
 
-  // Chip bottom edge must be at or above main bottom edge (allow 1px for sub-pixel rounding)
+  // Chip bottom edge must be at or above the viewport bottom (allow 1px sub-pixel rounding)
   expect(
     chipBottom,
-    `Collapsed chip bottom (${chipBottom.toFixed(1)}) is BELOW main bottom (${mainBottom.toFixed(1)}) — chip is clipped`,
-  ).toBeLessThanOrEqual(mainBottom + 1);
+    `Collapsed chip bottom (${chipBottom.toFixed(1)}) is BELOW viewport bottom (${viewportBottom.toFixed(1)}) — chip is clipped`,
+  ).toBeLessThanOrEqual(viewportBottom + 1);
 
-  // Chip left edge must be at or right of main left edge
+  // Chip left edge must be at or right of the viewport left edge
   expect(
     chipLeft,
-    `Collapsed chip left (${chipLeft.toFixed(1)}) is LEFT OF main left (${mainLeft.toFixed(1)}) — chip is clipped`,
-  ).toBeGreaterThanOrEqual(mainLeft - 1);
+    `Collapsed chip left (${chipLeft.toFixed(1)}) is LEFT OF viewport left (${viewportLeft.toFixed(1)}) — chip is clipped`,
+  ).toBeGreaterThanOrEqual(viewportLeft - 1);
 
-  return { chipBottom, mainBottom, chipLeft, mainLeft };
+  return { chipBottom, viewportBottom, chipLeft, viewportLeft };
 }
 
 test.describe('Collapsed legend chip visibility — P0 regression from #471', () => {
   test.describe('desktop (1440×900)', () => {
     test.use({ viewport: { width: 1440, height: 900 } });
 
-    test('collapsed chip is fully inside main#main-surface bounds', async ({ page, apiStub }) => {
+    test('collapsed chip is fully inside the viewport bounds', async ({ page, apiStub }) => {
       await setupRoutes(page, apiStub);
       const app = new AppPage(page);
       // Clear localStorage so we get the default expanded state, then collapse
@@ -202,12 +206,12 @@ test.describe('Collapsed legend chip visibility — P0 regression from #471', ()
       await toggle.click();
       await expect(toggle).toHaveAttribute('aria-expanded', 'false');
 
-      // Assert chip is inside main bounds
-      const { chipBottom, mainBottom } = await assertChipInsideMain(page);
+      // Assert chip is inside the viewport bounds (#761 S2: the map fills the viewport)
+      const { chipBottom, viewportBottom } = await assertChipInsideViewport(page);
       // Provide readable context in the test output
       expect(
-        chipBottom <= mainBottom + 1,
-        `Desktop: chip bottom ${chipBottom.toFixed(1)} <= main bottom ${mainBottom.toFixed(1)}`,
+        chipBottom <= viewportBottom + 1,
+        `Desktop: chip bottom ${chipBottom.toFixed(1)} <= viewport bottom ${viewportBottom.toFixed(1)}`,
       ).toBe(true);
     });
   });
@@ -215,7 +219,7 @@ test.describe('Collapsed legend chip visibility — P0 regression from #471', ()
   test.describe('mobile (390×844)', () => {
     test.use({ viewport: { width: 390, height: 844 } });
 
-    test('collapsed chip is fully inside main#main-surface bounds', async ({ page, apiStub }) => {
+    test('collapsed chip is fully inside the viewport bounds', async ({ page, apiStub }) => {
       await setupRoutes(page, apiStub);
       const app = new AppPage(page);
       // Mobile defaults to collapsed — clear localStorage and let the
@@ -236,11 +240,11 @@ test.describe('Collapsed legend chip visibility — P0 regression from #471', ()
       const toggle = page.getByRole('button', { name: /bird families/i });
       await expect(toggle).toHaveAttribute('aria-expanded', 'false');
 
-      // Assert chip is inside main bounds
-      const { chipBottom, mainBottom } = await assertChipInsideMain(page);
+      // Assert chip is inside the viewport bounds (#761 S2: the map fills the viewport)
+      const { chipBottom, viewportBottom } = await assertChipInsideViewport(page);
       expect(
-        chipBottom <= mainBottom + 1,
-        `Mobile: chip bottom ${chipBottom.toFixed(1)} <= main bottom ${mainBottom.toFixed(1)}`,
+        chipBottom <= viewportBottom + 1,
+        `Mobile: chip bottom ${chipBottom.toFixed(1)} <= viewport bottom ${viewportBottom.toFixed(1)}`,
       ).toBe(true);
     });
   });
--- a/frontend/e2e/map-cell-popover.spec.ts
+++ b/frontend/e2e/map-cell-popover.spec.ts
@@ -168,6 +168,28 @@ test('@coarse tablet 768×1024: tap marker → cluster-list popover → tap spec
   // Wait for the map render to complete (canonical pattern).
   await page.locator('[data-testid="adaptive-grid-marker"]').first().waitFor({ state: 'visible' });
 
+  // #761 (S2): wait for the map to go IDLE before tapping. The cluster-list
+  // popover renders INSIDE the AdaptiveGridMarker; if the marker-reconcile churn
+  // (sourcedata → render → idle, triggered while the canvas is still settling)
+  // re-creates the marker DOM node AFTER the tap, the marker's local
+  // `isClusterListOpen` state resets and the popover is dismissed mid-assertion.
+  // Post-inversion the map mounts into the full-viewport `#map-layer` and the
+  // settle window can extend just past first-marker-visible at this viewport
+  // (the corrective `map.resize()` on the flex→fixed transition is S3's — #773 —
+  // and is intentionally NOT wired here; S2 asserts steady state). Waiting for a
+  // settled idle frame closes that race deterministically without map-resize
+  // plumbing. If `__birdMap` is unavailable (no WebGL), fall through — the
+  // describe-level skip guards already cover that path.
+  await page
+    .waitForFunction(() => {
+      const map = (window as { __birdMap?: { isMoving: () => boolean; loaded: () => boolean } }).__birdMap;
+      return !!map && map.loaded() && !map.isMoving();
+    }, undefined, { timeout: 10_000 })
+    .catch(() => { /* no map hook (no WebGL) — the existing skip guards apply */ });
+  // Settle one more rAF beyond `idle` so any trailing reconcile commit lands
+  // before we interact (the marker DOM node is stable after this point).
+  await page.waitForTimeout(300);
+
   // Tap a multi-leaf cluster marker.
   const marker = page.locator('[data-testid="adaptive-grid-marker"]').first();
   await marker.tap();
--- a/frontend/e2e/map-root-geometry.spec.ts
+++ b/frontend/e2e/map-root-geometry.spec.ts
@@ -48,6 +48,110 @@ async function setupRoutes(
   await apiStub.stubZipIndex();
 }
 
+/**
+ * Silhouettes stub for the attribution-clearance guard. The FamilyLegend only
+ * mounts when `silhouettes.length > 0` (FamilyLegend.tsx), so the legend-vs-
+ * attribution overlap test must register a non-empty silhouettes payload — the
+ * `tyrannidae` row matches AZ_OBS's `familyCode`, plus the required `_FALLBACK`
+ * row. Registered AFTER setupRoutes so this more-specific route wins (LIFO).
+ */
+async function stubSilhouettesForLegend(
+  page: import('@playwright/test').Page,
+): Promise<void> {
+  await page.route('**/api/silhouettes', async (route) => {
+    await route.fulfill({
+      status: 200,
+      contentType: 'application/json',
+      body: JSON.stringify([
+        {
+          familyCode: 'tyrannidae',
+          color: '#E84040',
+          colorDark: '#E84040',
+          svgData:
+            'M5 13 C5 9 9 8 13 9 L17 7 L17 10 L15 11 L15 14 L13 15 L8 15 L5 13 Z',
+          source: null,
+          license: null,
+          commonName: 'Tyrant Flycatchers',
+          creator: null,
+        },
+        {
+          familyCode: '_FALLBACK',
+          color: '#555555',
+          colorDark: '#555555',
+          svgData:
+            'M 6 12 C 6 9 8 7 11 7 C 13 7 14 8 15 9 L 18 8 L 18 10 L 16 11 L 16 14 L 14 16 L 9 16 L 6 14 Z',
+          source: null,
+          license: null,
+          commonName: 'Unknown family',
+          creator: null,
+        },
+      ]),
+    });
+  });
+}
+
+/**
+ * WebGL skip guard for the attribution-clearance guard. The MapLibre
+ * AttributionControl only attaches its DOM (`.maplibregl-ctrl-attrib`) after the
+ * map fires `load`; headless runs without a GPU never fire it, so the
+ * bounding-box overlap check would be vacuous. Skip cleanly — mirrors the
+ * `__birdMap` guard in family-legend-viewport.spec.ts. CI runs software WebGL
+ * (SwiftShader), so CI exercises the full assertion.
+ */
+async function skipIfAttributionAbsent(
+  page: import('@playwright/test').Page,
+  testRef: typeof test,
+): Promise<boolean> {
+  const present = await page
+    .locator('.maplibregl-ctrl-attrib')
+    .waitFor({ state: 'attached', timeout: 8_000 })
+    .then(() => true)
+    .catch(() => false);
+  if (!present) {
+    testRef.skip(
+      true,
+      '.maplibregl-ctrl-attrib not attached — maplibre `load` did not fire ' +
+        '(likely WebGL unavailable in headless run).',
+    );
+  }
+  return !present;
+}
+
+/**
+ * Read the attribution control + family legend bounding boxes and the size of
+ * their intersection. Returns null if either element is missing.
+ */
+async function measureAttributionVsLegend(
+  page: import('@playwright/test').Page,
+): Promise<{
+  attrib: { left: number; top: number; right: number; bottom: number };
+  legend: { left: number; top: number; right: number; bottom: number };
+  intersectX: number;
+  intersectY: number;
+} | null> {
+  return page.evaluate(() => {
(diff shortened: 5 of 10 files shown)


julianken:
Design review (S2 #775): **FAIL** — one viewport-scoped collision blocks production; everything else is a polished, correct inversion.

The inversion itself reads exactly as intended: full-bleed map as the base layer with chrome floating cleanly on top (Google-Maps idiom achieved). At 1440/1920 the thin floating header is well-proportioned and legible over both basemaps, the centered scope cluster floats without occluding the map, the legend has real breathing room bottom-left, and the unscoped scrim over the idle map is excellent. AZ artboard framing and the border-zoom mask clip are clean (no interior-label overhang). No theme-contrast issues — header/overlays read against both light and dark basemaps.

**The blocker — legend ⨯ basemap-attribution collision at short viewports (bottom-left corner):**

- Evidence: `scoped-az-390x844-light.png` + `-390x844-dark.png`, `scoped-az-768x1024-light.png` + `-dark.png`, `scoped-az-1024x768-light.png`. The 5-row `.family-legend` bottom-aligns flush to the viewport edge and its bottom-left corner overlaps the OSM/OpenFreeMap attribution line, partially obscuring the license text (`...Cornell Lab of Ornithology | © ... from OpenStreetMap`). Reads as a crowded, colliding corner — not two cleanly separated floating elements — at exactly the mobile viewport the epic prioritizes.
- Root cause (this PR's adjacency, not pre-existing): `.family-legend` is anchored `bottom: var(--space-md); left: var(--space-md)` (`frontend/src/styles.css:632-635`) and the maplibre attribution control has **no positioning override** in the tree, so it sits library-default bottom-left/bottom — the same corner. Pre-S2 the map was windowed in a padded `<main>`, so the legend sat in a shorter box and the attribution rode the map's own bottom; now that `#map-layer` fills 100vh and the legend anchors to the **viewport** bottom, at heights where 5 family rows are tall enough the two overlap. The #471 clip-guard re-baseline (`#main-surface` → viewport) covers legend-vs-edge clipping but not this new legend-vs-attribution adjacency. Theme-independent. Desktop is clear only because there's vertical slack and the attribution collapses to a compact bottom-right pill.
- Fix (one of, scoped): lift the legend above the attribution band — `.family-legend { bottom: calc(var(--space-md) + 22px); }` (clears the ~22px attribution row) — **or** move the maplibre attribution control to bottom-right at map init so the bottom-left corner is the legend's alone. Prefer the latter if downstream sheets/popovers also want the bottom-left corner. Add an e2e guard asserting `.family-legend` bottom edge ≤ attribution-control top edge at 390×844 and 768×1024.

Secondary (non-blocking, note for S3/O4): at 1440 the cell/cluster detail popover (`desktop-detail-popover-1440-light.png`) bottom edge sits right on top of the attribution pill — currently fine since it clears, but the same bottom-anchored crowding will recur as popovers grow; worth folding into the same attribution-corner cleanup. Re-request review after the legend fix lands.

julianken:
Design review (S2 #775) — re-review after fix: **PASS**

The bottom-left `.family-legend` / basemap-attribution collision (the sole blocker from the prior FAIL) is resolved across all three short viewports × both themes.

Verified on the corrected captures:

- **390×844 (light + dark):** Legend now sits clearly above the attribution band; the full three-line license string ("OpenFreeMap | Bird data: eBird (Cornell Lab of Ornithology) | © OpenStreetMap contributors | OpenFreeMap © OpenMapTiles Data from OpenStreetMap") is fully visible and unobscured.
- **768×1024 (light + dark):** Legend lifted above the single-line attribution; all license text readable.
- **1024×768 (light + dark):** Same — attribution fully visible below the lifted legend.

The lift introduced no new problem: the legend stays anchored bottom-left (not floating mid-screen), reads as intentional, and does not collide with the scope control, the species markers near Tucson, or the map edges. The tiered `--attribution-clearance` token (60/40/20px at 390/768/≥1024) plus the disjoint-boxes e2e guard match the live OVERLAP=0 re-measurement.

Desktop 1440/1920 were already PASS and are unchanged. All viewports now PASS.

julianken:
@Mergifyio queue

Issue #737, opened 2026-05-29
feat(frontend): in-state on-map ScopeControl (StateSelector + ZipInput + whole-US niche affordance) + CSS


julianken:
**Goal:** The **in-state** on-map Scope control — rendered only AFTER a scope is chosen (in a `?state=US-XX` state view or a `?scope=us` whole-US view), NOT on the landing. It floats over the map canvas and lets a user re-scope without going back to the chooser: a native `<select>` StateSelector + the existing `<ZipInput>` (from #739/D5) + a small **de-emphasized** exit affordance ("Change scope" → returns to the chooser by clearing scope; and/or "Whole US" → `?scope=us`). With full CSS (this issue carries both Task **C4** and Task **C8**). ⚠️ The pre-map landing **chooser is a SEPARATE component, #742** — do NOT build the chooser here, and do NOT rebuild `ZipInput` here (it lands in #739 — import and reuse it).

**Files:**
- `frontend/src/components/ScopeControl.tsx` (+ `ScopeControl.test.tsx`) — C4
- `frontend/src/styles.css` — C4 (component rules) + C8 (the orphan-classname/token-layer sweep across `ScopeControl` + `ScopeChooser` + the scope surfaces)
- `frontend/src/components/ds/ds-primitives.css` — C8's token-layer pass covers the primitive layer too (only if a primitive className is touched; otherwise leave untouched)

**Depends on:** #735 (C2 — `state`/`zip`/`scope` in `UrlState`/`DEFAULTS`; ScopeControl reads/writes scope through this), #739 (D5 — `<ZipInput>` + `ScopeResolution`/`zipResolutionToScope`/`ZIP_FLYTO_ZOOM`; ScopeControl renders ZipInput and forwards its `onResolve`), #742 (C2a — the `ScopeChooser` whose classNames C8's orphan sweep MUST also cover; running C8's sweep before #742 lands its classNames produces a false-clean result, so **C8 sequences after #742 merges**), #738 (C7 — C8's orphan sweep must also cover the `MapLede`/`FilterSentence` scope-narration classNames C7 ships). Builds in parallel with #736 (C3 camera) against the shared `ScopeResolution` contract.

[...]

PR #797, opened 2026-05-30
feat(frontend,map): asymmetric fitBounds top-padding + map.resize on layout transition  closes #773

### LINKED ISSUE #773: S3: Resolve TODO(#737) — asymmetric `fitBounds` top-padding (header + scope chrome clearance) + verify `map.resize()` / `isMobile` on the S2 flex→fixed transition

julianken:
## Diagrams

S3 resolves two defects that S2's full-bleed extraction surfaced: the framed camera now clears the floating chrome (asymmetric top inset), and a camera-neutral `ResizeObserver` corrects the canvas size on the flex→fixed containing-block swap.

```mermaid
graph TB
    subgraph fit["Asymmetric fitBounds padding (was uniform 48)"]
        T["FIT_BOUNDS_PADDING<br/>{ top: 152, bottom/left/right: 48 }<br/>top = --header-height(48) + --space-md(12)<br/>+ wrapped .scope-control height(~88) + margin"]
        T --> S1["imperative scope-camera effect<br/>map.fitBounds(activeBounds, …)"]
        T --> S2["mount-frame initialViewState<br/>fitBoundsOptions (deep-link path)"]
        S1 --> R["framed north edge sits BELOW<br/>the header + scope-control band<br/>(top-latitude markers reachable)"]
        S2 --> R
    end

    subgraph rz["Camera-neutral resize on the S2 flex→fixed transition"]
        B["containing block flips:<br/>.map-surface flex child → position:absolute<br/>inside #map-layer (position:fixed; inset:0)"]
        B --> O["ResizeObserver on data-testid=map-canvas wrapper"]
        O --> D["rAF-coalesced + idempotent + disconnect on cleanup"]
        D --> RS["map.resize() ONLY<br/>(no fitBounds / flyTo / refetch)<br/>→ canvas recomputed for the new box"]
        RS -. "NOT a camera move →<br/>cannot schedule /api/observations<br/>(S4 scope-gate invariant, R1)" .-> RS
    end
```

## Summary

- **Why (padding):** post-#761/S2 the map is the viewport-filling root and two overlays — the fixed `.app-header` + the top-anchored `.scope-control` — float over its top edge. The deferred `TODO(#737)`'s uniform `padding: 48` let that chrome occlude the framed region's northern strip (a state's top-latitude markers became visually unreachable). A single `FIT_BOUNDS_PADDING` object (top grown, other three edges still 48) now drives **both** fitBounds sites — the imperative scope-camera effect and the mount-frame `initialViewState.fitBoundsOptions` (so deep-linked scoped landings frame correctly, not just in-session picks). `top: 152` is **derived from the CSS tokens** that own the stack (`--header-height` 48 + `.scope-control` `--space-md` offset + its **wrapped** ≤480px height ~88px), so the framed north edge clears the bar even on the narrowest viewport.
- **Why (resize):** S2 swapped the map's *containing block* (padded `<main>` flex child → fixed `#map-layer` sibling); maplibre's built-in observer did not always re-read the canvas box for the new full-viewport size on first paint — the one-frame mis-sized-canvas root cause of the settle race S2 worked around test-side. A debounced `ResizeObserver` on the `data-testid="map-canvas"` wrapper now calls `map.resize()` **only** (never `fitBounds`/`flyTo`/refetch → preserves the S4 scope-gate invariant, report R1); rAF-coalesced, idempotent, disconnected on cleanup. The `isMobile` cluster-tier read is documented (now sees the true full-bleed width; at exactly 768px the tier is desktop — intended, with a 767→mobile / 768→desktop boundary unit test).

## Screenshots

Live Playwright MCP verification PASSED — asymmetric top-padding clears the chrome at all 5 viewports × both themes (north-edge clearance +39 to +113px), camera-neutral `map.resize` refills the canvas, 0 console errors (one pre-existing basemap `circle-11` warning).

**Scope-fit framing — markers clear the floating header (AZ)**

| Viewport | Light | Dark |
| --- | --- | --- |
| 390×844 | scoped-az 390x844 light | scoped-az 390x844 dark |
| 768×1024 | scoped-az 768x1024 light | scoped-az 768x1024 dark |
| 1024×768 | scoped-az 1024x768 light | scoped-az 1024x768 dark |
| 1440×900 | scoped-az 1440x900 light | scoped-az 1440x900 dark |
| 1920×1080 | scoped-az 1920x1080 light | scoped-az 1920x1080 dark |

**Resize + scope-pick framing**

| Resize (camera-neutral refill, 1440 light) | Scope-pick fit (1440 light) |
| --- | --- |
| resize after 1440 light | scope-pick fit 1440 light |

- [x] Every new/renamed `className` in this PR has a matching CSS rule in
      (or marked `N/A — class is consumed only by a primitive that ships its own CSS`)
      — `N/A — no new classNames` (camera/lifecycle change; `orphan-classname-check`: PASS).
- [x] Multi-viewport design QA: PR body has ≥10 `user-attachments` URLs
      (5 viewports × 2 themes per #445). Verify:
      `gh pr view <N> --repo julianken/bird-sight-system --json body --jq '.body | [scan("user-attachments/assets/[a-f0-9-]++")] | length'`
      returns ≥ 10 (or marked `N/A — not UI`)
      — DONE: 12 `user-attachments` URLs embedded (10 viewport×theme + 2 interaction).

## Test plan

- [x] `npm test --workspace @bird-watch/frontend` — green (**65 files, 968 unit tests**). Converted the existing scalar `padding:48` test to the asymmetric-object assertion (`top` strictly `>` `bottom`/`left`/`right`; those three `=== 48`); added: mount-frame deep-link padding (reads `initialViewState.fitBoundsOptions`), camera-neutral `map.resize()` (fires on a box change, asserts no `fitBounds`/`flyTo`), and the `isMobile` 767→mobile / 768→desktop boundary; extended the reduced-motion test to assert the asymmetric shape (`duration:0` still lands it).
- [x] New unit / integration tests added (behavior changed).
- [x] New Playwright e2e spec assertion added — `state-scope.spec.ts` deep-link case now asserts the framed **AZ north edge** (37.0°) projects **below** the `.scope-control` bottom band via the live `__birdMap` (WebGL-guarded; would FAIL under the old uniform 48). Stable 3/3 reruns.
- [x] `npm run lint` — clean. `npm run build --workspace @bird-watch/frontend` — clean. `npx knip` — clean (exit 0). `orphan-classname-check` — PASS. No-DB-writes grep — clean.
- [x] **Coarse-pointer workaround decision — RETAINED, not removed.** S3's resize fix corrects the canvas *mis-size* (the visual flash). But `map-cell-popover.spec.ts`'s coarse cluster-tap race is a **separate** phenomenon: the React adaptive-grid reconciler re-creates marker DOM nodes on trailing cluster-leaf/tile reconciles past first-marker-visible (independent of canvas sizing). Verified empirically against the seeded local DB: removing S2's whole settle block fails **3/3** reruns; dropping just the trailing `waitForTimeout(300)` fails **4/5**; with both retained it passes **3/3**. So the resize fix does **not** make the workaround removable — the spec comment is updated to record this finding (the extra `waitForTimeout` is dropped only where it is genuinely redundant — it is not here). Local e2e run single-worker against Postgres+PostGIS on :5433: state-scope, map-cold-load, map-symbol-layer, map-root-geometry, zip-scope, map, and coarse map-cell-popover all green; WebGL-only adaptive-grid specs skip (no GPU) as in S2.
- [x] (UI only) Playwright MCP smoke — DONE: the orchestrator runs the live capture (this subagent did not use Playwright MCP per the dispatch). Resize-flash verify + scope-fit framing at all 5 viewports × both themes happen there; this PR's branch (a) — `map.resize()` added — is taken because S2 already demonstrated the mis-sized-canvas fallout (`sameMarkerNode === false` ~600ms post-tap).

## Plan reference

Part of #761 (map-first re-architecture), sub-issue **S3**. Resolves the deferred `TODO(#737)` (asymmetric `fitBounds` top-padding) and **gap 8** (camera-neutral `map.resize()` on the flex→fixed transition). Research report: `docs/analyses/2026-05-29-map-first-rearchitecture-761/report.md` — §6 / WS2 (Full-bleed map container, Risks line 121 + sub-issue 2.2 + gap 8 line 222), §10 Phase 3 S3 (line 251). Depends on **S2** (#775 / PR #795), which deferred this corrective resize to S3. Does not block **S4** (`onViewportChange` scope-gate) — the resize is camera-neutral so it cannot schedule a bbox refetch (report R1).

---

### DIFF
--- a/frontend/e2e/map-cell-popover.spec.ts
+++ b/frontend/e2e/map-cell-popover.spec.ts
@@ -168,18 +168,22 @@ test('@coarse tablet 768×1024: tap marker → cluster-list popover → tap spec
   // Wait for the map render to complete (canonical pattern).
   await page.locator('[data-testid="adaptive-grid-marker"]').first().waitFor({ state: 'visible' });
 
-  // #761 (S2): wait for the map to go IDLE before tapping. The cluster-list
-  // popover renders INSIDE the AdaptiveGridMarker; if the marker-reconcile churn
-  // (sourcedata → render → idle, triggered while the canvas is still settling)
-  // re-creates the marker DOM node AFTER the tap, the marker's local
-  // `isClusterListOpen` state resets and the popover is dismissed mid-assertion.
-  // Post-inversion the map mounts into the full-viewport `#map-layer` and the
-  // settle window can extend just past first-marker-visible at this viewport
-  // (the corrective `map.resize()` on the flex→fixed transition is S3's — #773 —
-  // and is intentionally NOT wired here; S2 asserts steady state). Waiting for a
-  // settled idle frame closes that race deterministically without map-resize
-  // plumbing. If `__birdMap` is unavailable (no WebGL), fall through — the
-  // describe-level skip guards already cover that path.
+  // Wait for the map to go IDLE before tapping. The cluster-list popover renders
+  // INSIDE the AdaptiveGridMarker; if a trailing marker-reconcile (sourcedata →
+  // render → idle) re-creates the marker DOM node AFTER the tap, the marker's
+  // local `isClusterListOpen` state resets and the popover is dismissed
+  // mid-assertion.
+  //
+  // #761/S3 (#773) note — this wait is RETAINED, not removed. S3 lands the
+  // corrective camera-neutral `map.resize()` (a ResizeObserver on the map-canvas
+  // wrapper in MapCanvas.tsx) that fixes the full-bleed canvas MIS-SIZE on the
+  // flex→fixed transition. But this spec's race is a SEPARATE phenomenon: the
+  // React adaptive-grid reconciler re-creates marker nodes on trailing
+  // cluster-leaf / tile-driven reconciles that fire just past first-marker-
+  // visible — independent of canvas sizing. S3 verified removal empirically:
+  // dropping this block fails 3/3 reruns, and dropping just the trailing
+  // `waitForTimeout(300)` fails 4/5, so both are kept. If `__birdMap` is
+  // unavailable (no WebGL), fall through — the describe-level skip guards cover it.
   await page
     .waitForFunction(() => {
       const map = (window as { __birdMap?: { isMoving: () => boolean; loaded: () => boolean } }).__birdMap;
--- a/frontend/e2e/state-scope.spec.ts
+++ b/frontend/e2e/state-scope.spec.ts
@@ -259,6 +259,43 @@ test.describe('Scope chooser + state/whole-US scope (C9, #741)', () => {
     for (const url of obsRequests) {
       expect(new URL(url).searchParams.get('state')).toBe('US-AZ');
     }
+
+    // #737/S3 (#773): the deep-linked scoped landing frames AZ with the
+    // asymmetric `fitBounds` top-padding so the STATE envelope's north edge is
+    // pushed clear of the floating header + scope-control chrome band (markers
+    // near the top latitude are not hidden behind the bar). Assert via the live
+    // maplibre instance: project AZ's actual north latitude (37.004) to a screen Y
+    // and require it to sit BELOW the scope-control's bottom edge. With the old
+    // uniform 48px padding this Y was ~48 (under the ~113px chrome band → occluded
+    // → would fail). WebGL-guarded — falls through when `__birdMap` is absent
+    // (no-WebGL CI project), where the unit test + map-root-geometry chrome guards
+    // still cover the contract.
+    await page
+      .waitForFunction(() => {
+        const map = (window as { __birdMap?: { loaded: () => boolean; isMoving: () => boolean } }).__birdMap;
+        return !!map && map.loaded() && !map.isMoving();
+      }, undefined, { timeout: 10_000 })
+      .catch(() => { /* no WebGL hook — covered by unit + geometry guards */ });
+    const clearance = await page.evaluate(() => {
+      const map = (window as {
+        __birdMap?: { project: (lnglat: [number, number]) => { y: number } };
+      }).__birdMap;
+      const sc = document.querySelector<HTMLElement>('.scope-control');
+      if (!map || !sc) return null;
+      // AZ envelope (STATES_FIXTURE bbox [-114.82, 31.33, -109.04, 37.0]): lng
+      // center ≈ -111.93, north lat = 37.0 — the value App.tsx passes as the scope
+      // `bounds` north and the camera frames to.
+      const northY = map.project([-111.93, 37.0]).y;
+      return { northY, scopeBottom: sc.getBoundingClientRect().bottom };
+    });
+    if (clearance) {
+      // The framed state's north edge projects BELOW the chrome band — not
+      // occluded by the floating header + scope-control.
+      expect(
+        clearance.northY,
+        `framed AZ north edge (y=${clearance.northY.toFixed(1)}) clears scope-control bottom (y=${clearance.scopeBottom.toFixed(1)})`,
+      ).toBeGreaterThanOrEqual(clearance.scopeBottom);
+    }
   });
 
   test('chooser state <select> lists the name-sorted states from /api/states', async ({
--- a/frontend/src/components/map/MapCanvas.test.tsx
+++ b/frontend/src/components/map/MapCanvas.test.tsx
@@ -138,6 +138,10 @@ function makeFakeMap() {
     triggerRepaint: vi.fn(),
     getCanvas: vi.fn(() => canvas),
     getContainer: vi.fn(() => container),
+    // #737/S3 — corrective resize on the flex→fixed container transition. Spied
+    // so the ResizeObserver effect's camera-neutral resize() is assertable AND
+    // so the fitBounds/flyTo camera spies can be checked unchanged after a fire.
+    resize: vi.fn(),
     easeTo: vi.fn(),
     // Camera-contract spies (#736). fitBounds/flyTo are the scope-driven
     // imperative moves; setMaxBounds is asserted to be NEVER called (maxBounds
@@ -265,6 +269,40 @@ vi.mock('react-map-gl/maplibre', () => ({
 }));
 
 vi.mock('maplibre-gl/dist/maplibre-gl.css', () => ({}));
+
+/* ── Controllable ResizeObserver + rAF stubs (#737/S3) ──────────────────────
+   jsdom has no ResizeObserver. MapCanvas's corrective resize effect guards on
+   `typeof ResizeObserver === 'undefined'` and no-ops without it, so to exercise
+   the camera-neutral `map.resize()` we install a stub that records each observer
+   instance and lets a test fire its callback. requestAnimationFrame is stubbed to
+   run synchronously so the rAF-coalesced resize lands within `act()`. */
+const resizeObservers: Array<{ cb: ResizeObserverCallback; fire: () => void }> = [];
+class StubResizeObserver {
+  cb: ResizeObserverCallback;
+  constructor(cb: ResizeObserverCallback) {
+    this.cb = cb;
+    resizeObservers.push({
+      cb,
+      // Simulate a box-size change notification to this observer.
+      fire: () => cb([] as unknown as ResizeObserverEntry[], this as unknown as ResizeObserver),
+    });
+  }
+  observe() {}
+  unobserve() {}
+  disconnect() {}
+}
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+(globalThis as any).ResizeObserver = StubResizeObserver;
+// Run rAF callbacks synchronously so the resize effect's coalescing frame fires
+// inside the test's act() flush (return a non-zero id so the `frame !== 0` guard
+// and cancelAnimationFrame cleanup behave like the real API).
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+(globalThis as any).requestAnimationFrame = (cb: FrameRequestCallback): number => {
+  cb(0);
+  return 1;
+};
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+(globalThis as any).cancelAnimationFrame = () => {};
 
 class FakeImage {
   src = '';
@@ -1601,6 +1639,7 @@ describe('MapCanvas controllable camera (#736)', () => {
     deferMapLoad = false;
     deferredOnLoad = null;
     fakeMap = makeFakeMap();
+    resizeObservers.length = 0; // #737/S3 — reset captured RO instances per test.
     document.documentElement.removeAttribute('data-theme');
     // Default: nothing matches (no reduced motion, fine pointer path is fine).
     stubMatchMedia(null);
@@ -1611,7 +1650,7 @@ describe('MapCanvas controllable camera (#736)', () => {
     window.matchMedia = origMatchMedia;
   });
 
-  it('fitBounds(state) on scope change uses padding 48 + maxZoom 12 + essential:true + duration 600 (no reduced motion)', async () => {
+  it('fitBounds(state) on scope change uses asymmetric padding (top > bottom/left/right=48) + maxZoom 12 + essential:true + duration 600 (no reduced motion)', async () => {
     stubMatchMedia(null); // prefers-reduced-motion: reduce → false
     render(
       <MapCanvas observations={[]} bounds={AZ_BOUNDS} boundsKey="US-AZ" />,
@@ -1619,14 +1658,44 @@ describe('MapCanvas controllable camera (#736)', () => {
     await waitFor(() => expect(fakeMap.fitBounds).toHaveBeenCalled());
     const [bounds, opts] = fakeMap.fitBounds.mock.calls.at(-1);
     expect(bounds).toEqual(AZ_BOUNDS);
-    expect(opts.padding).toBe(48);
+    // #737/S3: padding is now an asymmetric object — a larger TOP inset clears
+    // the floating AppHeader + ScopeControl chrome that stacks over the
+    // full-bleed canvas top edge post-#761/S2; the other three edges keep 48.
+    expect(typeof opts.padding).toBe('object');
+    expect(opts.padding.top).toBeGreaterThan(opts.padding.bottom);
+    expect(opts.padding.top).toBeGreaterThan(opts.padding.left);
+    expect(opts.padding.top).toBeGreaterThan(opts.padding.right);
+    expect(opts.padding.bottom).toBe(48);
+    expect(opts.padding.left).toBe(48);
+    expect(opts.padding.right).toBe(48);
     expect(opts.maxZoom).toBe(12);
     expect(opts.essential).toBe(true);
     expect(opts.duration).toBe(600);
     expect(fakeMap.flyTo).not.toHaveBeenCalled();
   });
 
-  it('fitBounds duration is 0 (still essential:true) under prefers-reduced-motion (plan P3 gate)', async () => {
+  it('mount-frame deep-link uses the same asymmetric fitBounds padding (initialViewState.fitBoundsOptions)', async () => {
+    stubMatchMedia(null);
+    render(
+      <MapCanvas observations={[]} bounds={AZ_BOUNDS} boundsKey="US-AZ" />,
+    );
+    // Read the props the MockMap captured into data-props — the mount-frame
+    // `initialViewState` carries `fitBoundsOptions` so a deep-linked scoped
+    // landing (?state=US-AZ) frames with the same asymmetric padding, not the
+    // legacy scalar 48.
+    const mockMap = await screen.findByTestId('mock-map');
+    const props = JSON.parse(mockMap.getAttribute('data-props') ?? '{}');
+    const fitOpts = props.initialViewState?.fitBoundsOptions;
+    expect(fitOpts).toBeTruthy();
+    expect(typeof fitOpts.padding).toBe('object');
+    expect(fitOpts.padding.top).toBeGreaterThan(fitOpts.padding.bottom);
+    expect(fitOpts.padding.bottom).toBe(48);
+    expect(fitOpts.padding.left).toBe(48);
+    expect(fitOpts.padding.right).toBe(48);
+    expect(fitOpts.maxZoom).toBe(12);
+  });
+
+  it('fitBounds duration is 0 (still essential:true + asymmetric padding) under prefers-reduced-motion (plan P3 gate)', async () => {
     stubMatchMedia('(prefers-reduced-motion: reduce)');
     render(
       <MapCanvas observations={[]} bounds={AZ_BOUNDS} boundsKey="US-AZ" />,
@@ -1635,6 +1704,9 @@ describe('MapCanvas controllable camera (#736)', () => {
     const [, opts] = fakeMap.fitBounds.mock.calls.at(-1);
     expect(opts.duration).toBe(0);
     expect(opts.essential).toBe(true);
+    // #737/S3: reduced motion still lands the asymmetric camera instantly.
+    expect(typeof opts.padding).toBe('object');
+    expect(opts.padding.top).toBeGreaterThan(opts.padding.bottom);
   });
 
   it('prefers flyTo over fitBounds when both a state bounds AND a ZIP flyTo are present on the same cycle (finding f)', async () => {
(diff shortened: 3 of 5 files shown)


julianken:
@Mergifyio queue

PR #793, opened 2026-05-30
feat(frontend): remove the feed surface entirely (no remnants)  closes #777

julianken:
## Diagrams

```mermaid
flowchart TD
    URL["URL ?view=feed (stale bookmark)"] --> readUrl["readUrl() in url-state.ts"]
    readUrl -->|"'feed' no longer in View union / VALID_VIEW"| Else["else branch → DEFAULTS.view = 'map'"]
    Else --> Scoped{"scope present?"}
    Scoped -->|"scope=us / state=US-XX"| Map["Map surface renders"]
    Scoped -->|"unscoped"| Chooser["ScopeChooser renders"]

    subgraph Deleted["Deleted in this PR (no remnants)"]
        FS["FeedSurface.tsx + test"]
        FC["FeedCard.tsx + test"]
        FR["FeedRow.tsx + test"]
        OFR["ObservationFeedRow.tsx + test"]
        FT["utils/format-time.ts + test (feed-only)"]
        CSS[".feed-* CSS rules + .feed-row/.feed-lede tokens"]
        DEP["react-window + @types/react-window deps"]
    end

    FS -. "was the only consumer of" .-> FT
    FS -. "was the only consumer of" .-> DEP
```

The feed surface was reachable **only** via a stale `?view=feed` URL — no header tab, no in-app navigation. Removing `'feed'` from the `View` union + `VALID_VIEW` makes `readUrl()` fall through to `DEFAULTS.view = 'map'`. No redirect shim is added (none ever covered `feed` — the existing shim only handles `hotspots`/`species`).

## Summary

- **Why:** the feed was dead UX on a live code path (report §6 WS5 / §10 F1+V3). Every downstream shell-inversion PR (#761) would otherwise have to keep a phantom surface alive, and ~9 e2e specs asserted a branch no real user can reach.
- **One branch, not two:** the component deletion, the `View`-member removal, and the now-dead `utils/format-time` + `react-window` dependencies are interdependent — splitting them would leave knip red or strand a dead render branch (report R18). All four CI checks (`test`/`lint`/`build`/`e2e`) plus `knip` and `orphan-classname-check` are green on this single branch.
- **Behavior:** a stale `?view=feed` URL falls through to the map (chooser when unscoped) with **no surviving `'feed'` string** in `frontend/src` or `frontend/e2e`. `aria-busy` on `<main>` keeps the attribute and drops only the dead `&& state.view === 'feed'` conjunct (now exactly `aria-busy={observationsLoading}`).

### What was deleted / changed
[...]

PR #812, opened 2026-05-31
feat(frontend,perf): O8 — React.memo boundaries on FamilyLegend + ScopeControl (closes #784)

### LINKED ISSUE #784: O8: memoize relocated persistent overlays (React.memo) + render-count regression test for the always-mounted map
## Problem

Once the always-mounted map ships (epic #761), the assist surfaces become **persistent** App-root overlays instead of conditionally-mounted children of `<main>`. O2 hoists the skip-link + `FamilyLegend` to fixed App-root siblings; O3 relocates the context strip (`MapLede` + `FilterSentence` + freshness) to a floating card. After that move, every relocated overlay re-renders whenever **any** `App`-level state changes — including state that has nothing to do with the overlay — because none of them are memoized and at least one App value (`now`) fans out on a tab-return signal.

Concretely, in the current tree these are all **plain, un-memoized function components**:

- `frontend/src/components/FamilyLegend.tsx:102` — `export function FamilyLegend(`
- `frontend/src/components/ScopeControl.tsx:66` — `export function ScopeControl(`
- `frontend/src/components/MapLede.tsx:58` — `export function MapLede(`
- `frontend/src/components/ds/FilterSentence.tsx:81` — `export function FilterSentence(`

A `grep -rln "React.memo" src/` returns only `FeedRow.tsx` and `ObservationFeedRow.tsx` — **no overlay is memoized today.** While the map and its in-`<main>` assist surfaces live inside `<main>` this is cheap (they unmount with the view, so there is nothing to re-render in the background). Under map-first they never unmount, so an unrelated parent re-render walks the whole always-mounted stack.

The fan-out vector is `nowTick`. `App.tsx:471-480` bumps `nowTick` on every `visibilitychange` → `visible`:

```
const [nowTick, setNowTick] = useState(() => new Date());
useEffect(() => {
  function handleVisibility() {
    if (document.visibilityState === 'visible') {
      setNowTick(new Date());
    }
  }
  document.addEventListener('visibilitychange', handleVisibility);
  return () => document.removeEventListener('visibilitychange', handleVisibility);
}, []);
const now = nowTick;          // App.tsx:481
```

Every tab-return triggers a full `App` re-render (a new `nowTick` `Date` identity flows into the `[freshestObservationAt, now]` `useMemo` and through the render). **Crucially, freshness reaches overlays only as *primitive* props, not as an object.** `App.tsx:487-490` destructures the memo result —

```
const { state: freshnessState, label: freshnessLabel } = useMemo(
  () => deriveFreshness(freshestObservationAt, now),
  [freshestObservationAt, now],
);          // App.tsx:487-490
```

— and threads the **destructured primitives** (`freshness={freshnessState}` / `freshnessLabel={freshnessLabel}`) into `FeedSurface` (`App.tsx:626-627`) and `MapSurface` (`App.tsx:669-670`). `deriveFreshness` (`lib/freshness.ts:69`) returns a string-union `state` and a string `label`. The object literal it constructs is never passed by reference — only `freshnessState` (string) and `freshnessLabel` (string) cross the prop boundary. So on a **same-minute** `nowTick` bump, those two primitive props are byte-equal, and a shallow `React.memo` boundary on a freshness-bearing overlay short-circuits **by value** (`string ===`) with no extra memoization needed. No object-identity stabilization is required, because no freshness object identity ever escapes into props.

The two overlays this issue actually guards for the `nowTick` case — `FamilyLegend` and `ScopeControl` — take **no freshness props at all**: `ScopeControl`'s props are `scope` / `states` / `onPickState` / `onPickWholeUs` / `onExit` / `onResolve` (`ScopeControl.tsx:43` `ScopeControlProps`), and `FamilyLegend`'s props (`FamilyLegend.tsx:13` `FamilyLegendProps`) carry no `now`/freshness field. Their zero-re-render guarantee on a `nowTick` bump therefore comes **entirely from the `React.memo` boundary** (they receive only primitives + `useCallback`-stable handlers, none of which change on a `nowTick` bump), independent of any freshness work.

The event handlers passed to overlays are already `useCallback`-stable (`onViewportChange` App.tsx:323-area, `onFamilyToggle`, `onPickState`, `onPickWholeUs`, `onResolveZip`, `onExitScope`, `onSelectSpecies` App.tsx:497), so stable *handler* props are achievable — but with no `React.memo` boundary the children re-render anyway on the `nowTick` bump even though their own props are unchanged. The fix is the memo boundary, not a freshness-derivation change.

**A subtlety the memo boundary must account for: not every overlay receives only primitive/stable props today.** `MapLede` (`App.tsx`/`MapSurface.tsx:250-260`) and `ScopeControl` (`App.tsx:641-648`) take only primitives or `useCallback`-stable handlers, so a shallow `React.memo` is immediately load-bearing for them. **`FilterSentence` is the exception:** at its current call site it receives `filters={{ since, notable, speciesCode, familyCode }}` — a **fresh inline object literal every render** (`MapSurface.tsx:262`). Under default shallow `React.memo`, that prop's reference changes on every parent render, so `React.memo(FilterSentence)` would be a **no-op** and the "at most once per toggle / zero on a same-minute nowTick bump" guarantees would be unachievable for `FilterSentence` unless the relocated O3 parent memoizes the `filters` object. The precedent already exists in the codebase: `FeedSurface.tsx:292-297` builds an `activeFilters` value via `useMemo(() => ({ notable, since, speciesCode, familyCode }), [filters])` and passes *that* to `FilterSentence` (`FeedSurface.tsx:427`). The map-side relocation must do the same.

**Theme toggle is the worst single fan-out and is currently untested for render cost.** `ThemeToggle` flips `[data-theme]` on `<html>` (`frontend/src/components/ThemeToggle.tsx:36`). That one attribute write fans out two ways at once:

1. Every CSS-var-driven overlay in the always-mounted stack recomputes style; with no memo boundary the React tree above them also re-renders if the toggle is co-located with App state.
2. The basemap `MutationObserver` in `frontend/src/components/map/MapCanvas.tsx:1453-1473` observes `data-theme` and calls `map.setStyle(...)` (`MapCanvas.tsx:1466`), forcing a full basemap repaint:

```
const observer = new MutationObserver((mutations) => {
  for (const mutation of mutations) {
    if (mutation.type === 'attributes' && mutation.attributeName === 'data-theme') {
      const next = document.documentElement.getAttribute('data-theme') === 'dark' ? 'dark' : 'light';
      if (next === prevThemeRef.current) return;
      prevThemeRef.current = next;
      const style = next === 'dark' ? BASEMAP_DARK : BASEMAP_LIGHT;
      map.setStyle(style);   // MapCanvas.tsx:1466
    }
  }
});
```

The windowed architecture absorbed this by unmounting overlays on view change. Map-first removes that escape hatch, so a single theme toggle now fans a React re-render across the persistent overlay stack **and** a `setStyle` repaint at the same time (research report §9 gap 5; WS8.3). Nothing guards the render count of that path today. Note the two halves live in **different test scopes**: the React-side render count is observable from `App.test.tsx` (the overlays are App-root siblings); the `setStyle`-fires-exactly-once dedup lives in `MapCanvas` and is observable only from a `MapCanvas`-level test — see the Test plan.

This is the perf half of the always-mounted-map workstream. It does not move DOM, change behavior, or touch styling — it adds `React.memo` boundaries, makes the `FilterSentence` `filters` prop referentially stable so its boundary is load-bearing, and locks the result with a render-count regression test.

## Scope

**Memoize the relocated persistent overlays (`React.memo`).** Wrap each always-mounted overlay so an unrelated `App` re-render does not propagate through it when its own props are referentially unchanged:

- `frontend/src/components/FamilyLegend.tsx` — wrap the `FamilyLegend` export in `React.memo`.
- `frontend/src/components/MapLede.tsx` — wrap `MapLede` in `React.memo`.
- `frontend/src/components/ds/FilterSentence.tsx` — wrap `FilterSentence` in `React.memo`.
- `frontend/src/components/ScopeControl.tsx` — wrap `ScopeControl` in `React.memo`.

Keep each module's existing named export name and signature unchanged (memoize internally and re-export under the same name, e.g. `export const FamilyLegend = memo(FamilyLegendImpl)`), so no import sites or tests change shape. Use default `React.memo` shallow comparison — do **not** add custom `areEqual` comparators (they hide future prop bugs). For the shallow boundary to be load-bearing, every prop must be either a primitive or a referentially-stable reference; that is already true for `MapLede` and `ScopeControl` (primitives + `useCallback`-stable handlers), and is made true for `FilterSentence` by the `filters`-prop memoization below. For the freshness-bearing overlays (`MapLede`/`MapSurface` context strip), the freshness props that cross the boundary are the **primitives** `freshnessState`/`freshnessLabel` (App.tsx:626/627/669/670), which a shallow `React.memo` compares by value — so a same-minute `nowTick` bump that yields a value-equal `(state, label)` pair short-circuits with no additional work.

**Make the `FilterSentence` `filters` prop referentially stable at the relocated O3 call site (`frontend/src/components/MapSurface.tsx` / the O3 floating-card parent).** Today the map side passes `filters={{ since, notable, speciesCode, familyCode }}` as an inline object literal (`MapSurface.tsx:262`), which defeats a shallow memo. When O3 relocates the context strip, the relocated parent must build the `filters` value via `useMemo(() => ({ since, notable, speciesCode, familyCode }), [since, notable, speciesCode, familyCode])` and pass the memoized value — mirroring the existing `FeedSurface.tsx:292-297` `activeFilters` pattern — so the `React.memo(FilterSentence)` boundary actually short-circuits on a `nowTick` bump. If O3's relocation has not yet introduced a stable `filters` reference for the map side when this issue is picked up, that memoization is part of this issue's diff (it is the load-bearing half of the FilterSentence boundary, not a separate concern).

**No `nowTick` freshness-derivation change is needed (`frontend/src/App.tsx`).** Freshness already reaches every overlay as the **primitive** props `freshnessState` (string union) and `freshnessLabel` (string), destructured from the memo at `App.tsx:487-490` and threaded at `App.tsx:626/627/669/670` — the `{ state, label }` object that `deriveFreshness` (`lib/freshness.ts:69`) returns is never passed by reference. So a same-minute `nowTick` bump produces **value-equal primitive props**, which a shallow `React.memo` boundary short-circuits without any extra memoization. Do **not** add an object-memoization step keyed on `[state, label]`: it would memoize an object that never crosses a prop boundary, and would be dead weight. The overlays this issue guards for the `nowTick` case (`FamilyLegend`, `ScopeControl`) take **no** freshness props at all, so behind their new `React.memo` boundary they see zero new props on a `nowTick` bump and must not re-render — that guarantee comes entirely from the memo boundary, not from any freshness work.

**Add a render-count regression test.** Extend the existing `mapSurfaceRef.renderCount` instrumentation pattern in `frontend/src/App.test.tsx` (the `vi.mock('./components/MapSurface.js')` stub increments a counter at `App.test.tsx:112`) to instrument the relocated App-root overlays, then assert:

1. A `nowTick` bump (`visibilitychange` → `visible`) within the same minute (so the resolved freshness label is unchanged) produces **zero additional renders** of the non-freshness overlays (`FamilyLegend`, `ScopeControl`).
2. A theme toggle (`data-theme` flip on `<html>`) does not produce an unbounded render cascade — assert each **App-root-instrumented** overlay re-renders **at most once** per toggle (and any freshness-bearing overlay only when its props actually change). The companion guarantee that `MapCanvas`'s `setStyle` path fires **exactly once** per flip is asserted in a `MapCanvas`-level test, not `App.test.tsx` (the `MapSurface`/`MapCanvas` subtree is mocked away there) — see the Test plan.

See the Test plan for the hard constraint that instrumentation can only observe overlays O2/O3 have actually hoisted to App-root siblings; any overlay still nested inside the `MapSurface` stub on this branch is out of scope for the `App.test.tsx` assertion and keeps its render-count guard in its own component test.

## Out of scope

- **Hoisting the overlays out of `<main>` / making them fixed App-root siblings** — that is O2 (skip-link + `FamilyLegend`) and O3 (context strip → floating card). This issue assumes those relocations have landed (it is a hard dependency) and only adds the memo boundaries + the `filters`-prop stabilization + the test. Do not move DOM here.
- **Any change to the `nowTick`/`deriveFreshness` derivation in `App.tsx`.** The destructured-primitive wiring at `App.tsx:487-490`/`626`/`627`/`669`/`670` is already correct for this issue's purposes — see the Scope note. Do not refactor it.
- **`ObservationPopover` (`MapCanvas.tsx:1974`).** It renders **inside** `MapCanvas`, is driven by `MapCanvas`-local selection state (`selectedObs`), and is never in the `App` `nowTick` render path; neither O2 (skip-link + `FamilyLegend`) nor O3 (context strip) relocates it to App root. Memoizing it — and asserting "zero App-`nowTick` re-renders" of it — would be vacuous/unverifiable from `App.test.tsx`, since it never lived in the path `nowTick` cascades through. If a future task genuinely makes it an App-root overlay, it gets its own issue citing that relocation; any render-count optimization of the popover today is a separate `MapCanvas`-local concern guarded by `ObservationPopover.test.tsx` / a `MapCanvas`-level test, not this one.
- **The named z-index token scale** (P1 / WS3.1) and any overlay stacking changes — separate issue.
- **Scope-gating `onViewportChange`** (S4 / WS8.2 / R1) — the unscoped-landing zero-`/api/observations` guard is its own issue.
- **maplibre scope-gated prefetch** (O9 / WS8.1) and the dist chunk-size baseline (P3 / WS8.4) — separate perf issues.
- **Basemap `setStyle`-while-inert correctness and the `basemap-dark-flip.spec.ts` pixel-coordinate re-baseline** (V2 / gap 4) — this issue only guards the *render-count* of the theme path (and, in a `MapCanvas`-level test, the `setStyle`-fires-exactly-once dedup), not the dark-flip pixel ACs or the inert-canvas `setStyle` behavior.
- **`map.resize` on the flex→fixed transition** (S3 / gap 8) — separate.
- Any change to data fetching, caching, URL state, or visible layout/styling.

## Acceptance criteria

- [ ] `FamilyLegend`, `MapLede`, `FilterSentence`, and `ScopeControl` are each wrapped in `React.memo` with their existing export names and prop signatures unchanged (no import-site or call-site edits required for the wrapping itself).
- [ ] The relocated map-side `FilterSentence` `filters` prop is supplied as a `useMemo`-stable object (keyed on `[since, notable, speciesCode, familyCode]`), mirroring `FeedSurface`'s `activeFilters` (`FeedSurface.tsx:292-297`), so the `React.memo(FilterSentence)` boundary is load-bearing rather than a no-op.
- [ ] The `nowTick` visibilitychange bump (`App.tsx:471-480`) no longer cascades a re-render through the overlays that do not consume freshness: on a same-minute tab return, `FamilyLegend` and `ScopeControl` (which take **no** freshness props) record **zero** additional renders, behind their new `React.memo` boundaries. (No freshness-derivation change is made — freshness reaches overlays only as the primitive props `freshnessState`/`freshnessLabel` at `App.tsx:626/627/669/670`, which are value-equal on a same-minute bump and so are already short-circuited by shallow `React.memo`; see Scope.)
- [ ] **At the App level:** a theme toggle (`[data-theme]` flip on `<html>`) re-renders each App-root-instrumented overlay **at most once**; console stays clean. This is asserted in `frontend/src/App.test.tsx`.
- [ ] **At the `MapCanvas` level:** a `[data-theme]` flip triggers the `MapCanvas` `setStyle` path **exactly once** per flip (no React-side double-fire), guarded by the existing `prevThemeRef` dedup (`MapCanvas.tsx:1463-1466`). This is asserted in a `MapCanvas`-level test (e.g. `MapCanvas.test.tsx`), **not** `App.test.tsx` — the `MapSurface`/`MapCanvas` subtree is mocked to a bare stub there, so the real `MutationObserver` + `setStyle` never run in `App.test.tsx`.
- [ ] A render-count regression test exists in `frontend/src/App.test.tsx` covering both the `nowTick`-bump case and the App-level theme-toggle case, using the existing `renderCount` instrumentation pattern, scoped to whichever target overlays O2/O3 have actually hoisted to App-root siblings on this branch (see Test plan).
- [ ] No visible behavior, layout, or styling change at any of the 5 canonical viewports × 2 themes; no DOM moves.
- [ ] **This PR alone leaves all four CI checks green — `test`, `lint`, `build`, `e2e` — plus `knip`.** No "leave CI red, fix in the next PR." `knip` must stay clean on the same branch (it is a required Mergify-gating check even though the workflow runs `continue-on-error: true`); `React.memo` wrapping does not introduce unused exports, but verify no helper added for the test is left dangling.

## Test plan

**Instrumentation precondition (read first).** `App.test.tsx` replaces the entire `MapSurface` subtree with a stub via `vi.mock('./components/MapSurface.js')` (`App.test.tsx:101-115`) that returns `<div data-testid="map-surface-stub" />`. **Anything that renders inside `MapSurface`/`MapCanvas` today — `MapLede`, `FilterSentence`, `ObservationPopover`, and the entire `MapCanvas` `MutationObserver` + `map.setStyle` theme path — is therefore invisible to `App.test.tsx` and cannot be render-counted (or `setStyle`-counted) there until O2/O3 actually hoist it out to an App-root sibling.** (Confirmed: `App.test.tsx` contains zero references to `setStyle`, `MutationObserver`, `data-theme`, or `BASEMAP`.) `ScopeControl` is already a direct App child (`App.tsx:641`) and so is observable from `App.test.tsx` today (though still inside `<main>` pre-O2/O3). Concretely:

- Add a dedicated `vi.mock` counter (mirroring the `mapSurfaceRef.renderCount` pattern at `App.test.tsx:112`) **only for overlays that O2/O3 have moved to App-root siblings of `MapSurface`** on the branch this issue picks up. The residual `MapSurface` stub keeps its existing `renderCount` semantics untouched.
- If a target overlay (e.g. `MapLede`, `FilterSentence`) is **still a `MapSurface`/`MapCanvas` descendant** on this branch, it is **out of scope for the `App.test.tsx` render-count assertion** — its render-count guard belongs to its own component test (`MapLede.test.tsx`, `FilterSentence.test.tsx`), and this issue only adds the `React.memo` wrapping for it, not an App-level render-count assertion.
- This means the exact set of overlays the App-level test asserts on is determined by what O2/O3 landed; `FamilyLegend` (O2) and `ScopeControl` (already an App child; O2/O3 finalize its placement) are the minimum, and the context-strip pieces join only once O3 has un-nested them from the `MapSurface` stub.

**Unit (vitest + React Testing Library) — `frontend/src/App.test.tsx`:**
- **`nowTick` decoupling test:** render the scoped map view, reset the in-scope overlay render counters, dispatch a `visibilitychange` with `document.visibilityState === 'visible'` (`Object.defineProperty` + `document.dispatchEvent(new Event('visibilitychange'))`) within the same minute (so the resolved freshness label is byte-identical), and assert the non-freshness overlays (`FamilyLegend`, `ScopeControl`, plus any hoisted context-strip piece that does not take freshness) recorded zero additional renders. (No object-identity assertion on the freshness derivation — freshness crosses the boundary only as value-equal primitives on a same-minute bump.)
- **App-level theme-toggle render-count test:** reset counters, flip `document.documentElement.setAttribute('data-theme', 'dark')`, and assert each App-root-instrumented overlay re-rendered **at most once**. Do **not** assert on `setStyle` here — the real `MapCanvas` is mocked away, so any `setStyle` assertion in `App.test.tsx` would spy on a stub and prove nothing about the real dedup.
- Keep the existing `App.test.tsx` suite green. The `mapSurfaceRef.renderCount` resets at `App.test.tsx:150/227/270/333/407/479/533/583/650/746/800/924` are reused; do not perturb their expectations. The new overlay counters are **separate** refs reset in their own test bodies — do not fold them into `mapSurfaceRef`.

**`MapCanvas`-level unit test — `frontend/src/components/map/MapCanvas.test.tsx` (or the existing `MapCanvas` test file):**
- **`setStyle`-exactly-once test:** with a real (or test-double) maplibre map whose `setStyle` is a spy, flip `document.documentElement.setAttribute('data-theme', 'dark')` and assert the `MutationObserver` callback (`MapCanvas.tsx:1453-1466`) invokes `map.setStyle` **exactly once**; flip back and forth to confirm the `prevThemeRef` dedup (`MapCanvas.tsx:1463-1466`) suppresses no-op re-fires. This is the home for the "exactly once / no React-side double-fire" guarantee — it cannot live in `App.test.tsx`, which mocks `MapCanvas` away.

**Component unit tests (no re-baseline expected):** `FamilyLegend.test.tsx`, `MapLede.test.tsx`, `FilterSentence.test.tsx`, and `ScopeControl.test.tsx` must still pass unchanged — `React.memo` is transparent to render-by-name and prop-driven assertions. If any test imports the component as a value and asserts on its `.name`/identity, adjust the wrapping (`memo` preserves rendering but changes the component object) rather than the test's behavioral intent.

**E2E:** No new e2e spec and no e2e re-baseline are required — this is a render-internal optimization with no DOM/behavior delta. Run the full `e2e` suite to confirm zero regressions; in particular `basemap-dark-flip.spec.ts` (theme path) and `family-legend.spec.ts` must stay green untouched. If any spec goes red, the memo wrapping or the `filters` memoization changed observable behavior and must be corrected here (not deferred).

## Dependencies

Part of #761.

Depends on (epic-relative, see the epic's authoritative numbered graph): **O2** (hoist skip-link + `FamilyLegend` to persistent App-root, `position: fixed`) and **O3** (relocate context strip — `MapLede` + `FilterSentence` + freshness — to a floating card). Both must land first because this issue memoizes the overlays *after* they have become always-mounted App-root siblings; memoizing them while they still mount/unmount with the view would be a no-op, and the `App.test.tsx` render-count assertions cannot observe overlays still nested inside the `MapSurface` mock (see Test plan). Transitively gated by S2 (shell inversion), which O2/O3 themselves depend on.

## Plan traceability

- Research report: `docs/analyses/2026-05-29-map-first-rearchitecture-761/report.md` — §6 **WS8 — Performance of always-mounted map** ("memoize relocated persistent overlays (React.memo + decouple `nowTick`)", sub-issue **8.3**); §9 completeness gap **5** ("Theme switching as a re-render/repaint vector across the persistent overlay stack — add theme-toggle to the render-count test"); §10 sequencing entry **O8** (depends on O2, O3).
- Epic workstream: #761 → **Workstream 8 (Performance of an always-mounted map)** — "Maintain memoization discipline as overlays move to top level … so unrelated `App` state changes don't cascade re-renders through every persistent overlay."

julianken:
## Diagrams

```mermaid
graph TD
    A[App re-render — nowTick bump] --> B[AppHeader re-render]
    A --> C{FamilyLegend\nReact.memo}
    B --> D{ScopeControl\nReact.memo}
    C -- "props unchanged\n(memo short-circuits)" --> E[NO re-render ✓]
    D -- "props unchanged\n(memo short-circuits)" --> F[NO re-render ✓]
    A --> G[MapCanvas MutationObserver]
    G -- "prevThemeRef dedup\n(setStyle fires ONCE)" --> H[basemap repaint ✓]
```

Both persistent overlays were plain un-memoized function components before this PR. Under the map-first architecture (epic #761) they are always-mounted App-root/AppHeader siblings — a `nowTick` visibilitychange bump caused a full re-render walk through both even when their props were unchanged.

## Summary

- Wrap `FamilyLegend` (App.tsx:1229) and `ScopeControl` (AppHeader.tsx:211) in `React.memo` using the `memo(FamilyLegendImpl)` / `React.memo(ScopeControlImpl)` pattern, preserving named exports and setting `displayName` for devtools.
- No prop stabilization needed at either call site: all props are primitives or `useCallback`-stable references (confirmed at each call site), so the default shallow comparison short-circuits on any unrelated App re-render.
- `MapLede` and `FilterSentence` intentionally left alone — they are no longer live overlays (removed in #793/#805); only `DsPreview.tsx` renders them.

## Screenshots

N/A — not visible UI (perf: React.memo boundaries + render-count/setStyle tests; no DOM/style change)

      (or marked `N/A — class is consumed only by a primitive that ships its own CSS`)
- [x] Multi-viewport design QA: N/A — not UI

## Test plan

- [x] `npm test --workspace @bird-watch/frontend` — 1062 passed (67 files)
- [x] New render-count regression tests added in `App.test.tsx` (4 tests):
  - `nowTick bump → ZERO additional renders` of `FamilyLegend` (memo holds)
  - `nowTick bump → ZERO additional renders` of `ScopeControl` (memo holds)
  - `theme toggle → ≤1 extra render` of `ScopeControl` (memo holds)
  - `theme toggle → ≤1 extra render` of `FamilyLegend` (memo holds)
- [x] New MapCanvas setStyle tests added in `MapCanvas.test.tsx` (2 tests):
  - `light→dark flip fires map.setStyle EXACTLY ONCE` (prevThemeRef dedup)
  - `redundant same-value mutation fires map.setStyle ZERO times`
- [x] `npx playwright test --workers=1` — 246 passed, 15 skipped (perf gate excluded)
- [x] `npx knip` — no new dead code (pre-existing config hint only)
- [x] `bash scripts/check-orphan-classnames.sh` — PASS
- [x] (UI only) Playwright MCP smoke — N/A (perf-only, no DOM/style change)

## Plan reference

Part of epic #761 map-first overlay series (O8). Out of dedicated plan file — tracked in issue #784.

---

### DIFF
--- a/frontend/src/App.test.tsx
+++ b/frontend/src/App.test.tsx
@@ -2,6 +2,7 @@ import { describe, it, expect, vi, afterEach, beforeEach } from 'vitest';
 import { render, screen, waitFor, act, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import type { LngLatBounds } from 'maplibre-gl';
+import type React from 'react';
 
 // Phase 4 / #663: useIsCompact calls window.matchMedia. JSDOM does not
 // implement it — polyfill with a stub that returns non-compact (wide
@@ -27,6 +28,7 @@ const {
   mockUrlState,
   mapSurfaceRef,
   mockPrefetchMapCanvas,
+  overlayRenderCounts,
 } = vi.hoisted(() => ({
   mockGetHotspots: vi.fn(),
   mockGetObservations: vi.fn(),
@@ -78,6 +80,12 @@ const {
       | { center: [number, number]; zoom: number; key: string }
       | undefined,
     renderCount: 0,
+  },
+  // O8 (#784): render-count tracking for the two memoized App-root overlays.
+  // Incremented by the stubs below; asserted in the "O8 memo boundary" suite.
+  overlayRenderCounts: {
+    familyLegend: 0,
+    scopeControl: 0,
   },
 }));
 
@@ -134,6 +142,38 @@ vi.mock('./components/MapSurface.js', () => ({
 vi.mock('./prefetch.js', () => ({
   prefetchMapCanvas: mockPrefetchMapCanvas,
 }));
+
+// O8 (#784): wrap FamilyLegend + ScopeControl with render-counting HOCs.
+// Each mock imports the REAL implementation (so all prop-passing, event
+// emission, and DOM content remain correct for existing tests) and wraps it in
+// an additional React.memo layer whose render body increments overlayRenderCounts
+// before delegating to the real component. This keeps the full ScopeControl ZIP
+// / exit / state-select behavior intact for #740 / O9 integration tests while
+// also making the render-count assertions load-bearing for O8's memo guard:
+// a nowTick bump that causes either wrapper to re-render will increment the
+// counter, failing the assertion in the O8 suite.
+vi.mock('./components/FamilyLegend.js', async () => {
+  const real = await vi.importActual<typeof import('./components/FamilyLegend.js')>('./components/FamilyLegend.js');
+  const { memo, createElement } = await import('react');
+  const RealFamilyLegend = real.FamilyLegend;
+  const WrappedFamilyLegend = memo(function FamilyLegendCounting(props: Parameters<typeof RealFamilyLegend>[0]) {
+    overlayRenderCounts.familyLegend += 1;
+    return createElement(RealFamilyLegend as React.ComponentType<typeof props>, props);
+  });
+  WrappedFamilyLegend.displayName = 'FamilyLegend';
+  return { ...real, FamilyLegend: WrappedFamilyLegend };
+});
+vi.mock('./components/ScopeControl.js', async () => {
+  const real = await vi.importActual<typeof import('./components/ScopeControl.js')>('./components/ScopeControl.js');
+  const { memo, createElement } = await import('react');
+  const RealScopeControl = real.ScopeControl;
+  const WrappedScopeControl = memo(function ScopeControlCounting(props: Parameters<typeof RealScopeControl>[0]) {
+    overlayRenderCounts.scopeControl += 1;
+    return createElement(RealScopeControl as React.ComponentType<typeof props>, props);
+  });
+  WrappedScopeControl.displayName = 'ScopeControl';
+  return { ...real, ScopeControl: WrappedScopeControl };
+});
 
 // Stub the ApiClient constructor so useBirdData / useStates receive a
 // controllable mock.
@@ -1952,5 +1992,161 @@ describe('O2 (#770): skip-link + FamilyLegend hoisted to App-root siblings', () 
     const mapLayer = container.querySelector('#map-layer');
     if (!skipLink || !mapLayer) return;
     expect(mapLayer.contains(skipLink)).toBe(false);
+  });
+});
+
+describe('O8 (#784): React.memo render-count regression — FamilyLegend + ScopeControl', () => {
+  /**
+   * Asserts that the two memoized App-root overlays (FamilyLegend, ScopeControl)
+   * do NOT re-render when unrelated App-level state changes.
+   *
+   * Fan-out vector: `nowTick` bumps on every `visibilitychange` → `visible`.
+   * On a same-minute bump the resolved freshness label is unchanged, so
+   * FamilyLegend and ScopeControl (which take no freshness props) must stay
+   * at 0 additional renders — their memo boundary short-circuits.
+   *
+   * Theme toggle: AppHeader re-renders (freshnessLabel/region propagate through
+   * it), but ScopeControl's own props are unchanged → at most 1 render per
+   * theme flip is allowed for ScopeControl (the initial render only; subsequent
+   * nowTick or irrelevant-state bumps must not produce more renders).
+   *
+   * Stubs are memo-wrapped (see vi.mock above) so the test is load-bearing:
+   * removing React.memo from either stub would cause the nowTick-bump assertion
+   * to observe a render-count increment and fail.
+   */
+  const SCOPED_US_STATE = {
+    since: '14d' as const,
+    notable: false,
+    speciesCode: null as string | null,
+    familyCode: null as string | null,
+    view: 'map' as const,
+    scope: { kind: 'us' as const } as
+      | { kind: 'unscoped' }
+      | { kind: 'us' }
+      | { kind: 'state'; stateCode: string },
+  };
+
+  beforeEach(() => {
+    __resetSilhouettesCache();
+    __resetStatesCache();
+    mockGetHotspots.mockResolvedValue([]);
+    mockGetObservations.mockResolvedValue({ data: [], meta: { freshestObservationAt: null } });
+    // Return a non-empty silhouettes list so the FamilyLegend gate passes
+    // (mapVisible && scopeActive && silhouettes.length > 0) and the stub renders.
+    mockGetSilhouettes.mockResolvedValue([{
+      familyCode: 'tyrannidae',
+      color: '#C77A2E',
+      colorDark: '#C77A2E',
+      svgData: 'M0 0L1 1Z',
+      svgUrl: null,
+      source: 'placeholder',
+      license: 'CC0',
+      commonName: 'Tyrant Flycatchers',
+      creator: null,
+    }]);
+    mockGetStates.mockResolvedValue([]);
+    mapSurfaceRef.renderCount = 0;
+    overlayRenderCounts.familyLegend = 0;
+    overlayRenderCounts.scopeControl = 0;
+    mockUrlState.state = SCOPED_US_STATE;
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+    document.documentElement.removeAttribute('data-theme');
+  });
+
+  it('nowTick bump (visibilitychange→visible) produces ZERO additional renders of FamilyLegend', async () => {
+    render(<App />);
+
+    // Wait for data load: App resolves silhouettes and FamilyLegend's HOC wrapper
+    // is called at least once (mapVisible && scopeActive both true in SCOPED_US_STATE).
+    // The HOC wraps the real component — we wait on the render count, not a DOM stub.
+    await waitFor(() => {
+      expect(overlayRenderCounts.familyLegend).toBeGreaterThan(0);
+    });
+
+    const countAfterMount = overlayRenderCounts.familyLegend;
+
+    // Simulate a nowTick bump — the same signal App wires on visibilitychange.
+    // jsdom does not fire real events on documentElement.hidden, so dispatch
+    // visibilitychange with visibilityState='visible' manually.
+    Object.defineProperty(document, 'visibilityState', {
+      value: 'visible',
+      writable: true,
+      configurable: true,
+    });
+    await act(async () => {
+      document.dispatchEvent(new Event('visibilitychange'));
+    });
+
+    // React.memo short-circuits: FamilyLegend's props are unchanged → 0 extra renders.
+    expect(overlayRenderCounts.familyLegend).toBe(countAfterMount);
+  });
+
+  it('nowTick bump (visibilitychange→visible) produces ZERO additional renders of ScopeControl', async () => {
+    render(<App />);
+
+    // ScopeControl renders inside AppHeader on a scoped view. The HOC wrapper
+    // is called as soon as App mounts with scopeActive=true.
+    await waitFor(() => {
+      expect(overlayRenderCounts.scopeControl).toBeGreaterThan(0);
+    });
+
+    const countAfterMount = overlayRenderCounts.scopeControl;
+
+    Object.defineProperty(document, 'visibilityState', {
+      value: 'visible',
+      writable: true,
+      configurable: true,
+    });
+    await act(async () => {
+      document.dispatchEvent(new Event('visibilitychange'));
+    });
+
+    // React.memo short-circuits: ScopeControl's props are unchanged → 0 extra renders.
+    expect(overlayRenderCounts.scopeControl).toBe(countAfterMount);
+  });
+
+  it('theme toggle does not produce an unbounded render cascade (ScopeControl: at most 1 extra render per toggle)', async () => {
+    document.documentElement.setAttribute('data-theme', 'light');
+    render(<App />);
+
+    await waitFor(() => {
+      expect(overlayRenderCounts.scopeControl).toBeGreaterThan(0);
+    });
+
+    const countAfterMount = overlayRenderCounts.scopeControl;
+    expect(countAfterMount).toBeGreaterThan(0);
+
+    // Flip [data-theme]: triggers AppHeader to re-render (freshness reads theme
+    // indirectly via AppHeader's ledeText/region props), but ScopeControl's
+    // direct props (scope/states/callbacks) are unchanged → memo should hold.
+    await act(async () => {
+      document.documentElement.setAttribute('data-theme', 'dark');
+    });
+
+    // ScopeControl's memo boundary holds on a theme flip (its props don't change).
+    // Assert at most 1 additional render — any more indicates the memo is broken.
+    expect(overlayRenderCounts.scopeControl - countAfterMount).toBeLessThanOrEqual(1);
+  });
+
+  it('theme toggle does not produce an unbounded render cascade (FamilyLegend: at most 1 extra render per toggle)', async () => {
+    document.documentElement.setAttribute('data-theme', 'light');
+    render(<App />);
+
+    await waitFor(() => {
+      expect(overlayRenderCounts.familyLegend).toBeGreaterThan(0);
+    });
+
+    const countAfterMount = overlayRenderCounts.familyLegend;
+    expect(countAfterMount).toBeGreaterThan(0);
+
+    await act(async () => {
+      document.documentElement.setAttribute('data-theme', 'dark');
+    });
+
+    // FamilyLegend's memo boundary holds on a theme flip (its props don't change).
+    expect(overlayRenderCounts.familyLegend - countAfterMount).toBeLessThanOrEqual(1);
   });
 });
--- a/frontend/src/App.test.tsx
+++ b/frontend/src/App.test.tsx
@@ -82,7 +82,7 @@ const {
     renderCount: 0,
   },
   // O8 (#784): render-count tracking for the two memoized App-root overlays.
-  // Incremented by the stubs below; asserted in the "O8 memo boundary" suite.
+  // Incremented by the counting wrappers in vi.mock below; asserted in the O8 suite.
   overlayRenderCounts: {
     familyLegend: 0,
     scopeControl: 0,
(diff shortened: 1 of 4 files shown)


julianken:
@Mergifyio queue

PR #841, opened 2026-06-02
fix(ingestor): fan out recent ingest per-state (#840)

julianken:
## Diagrams

```mermaid
sequenceDiagram
    participant Ingestor
    participant eBird
    participant Postgres

    loop for each of 49 CONUS_STATE_CODES (paced ≥1000ms between rounds)
        Ingestor->>eBird: GET /data/obs/US-XX/recent
        Ingestor->>eBird: GET /data/obs/US-XX/recent/notable
        Note over Ingestor: per-state notable intersection (keyset is state-local)
        alt 429 (rate limited)
            eBird-->>Ingestor: 429 + Retry-After
            Note over Ingestor: "back off (Retry-After else exponential)<br/>3 consecutive 429s ⇒ circuit-break ⇒ abort run as failure"
        else 500 / timeout
            eBird-->>Ingestor: error
            Note over Ingestor: isolate — record + continue (other 48 states unaffected)
        end
    end
    Ingestor->>Ingestor: aggregate + dedup (subId, speciesCode); #484 species-meta invariant
    Ingestor->>Postgres: upsert observations (silhouette + is_notable stamp)
    Note over Ingestor: status ladder — success=0 failed · partial=1..5 failed · failure=>5 failed OR repeated 429
```

## Summary

- **Why:** the daily `recent` ingest made ONE nationwide `/data/obs/US/recent` call, which eBird structurally dedups to one-observation-per-species region-wide (879 records / 879 distinct species for `US`). High-activity regions win every tie, so non-AZ states were starved — Missouri showed 15 species/7d in our API vs 214 in eBird. This fans `recent` out per-state over `CONUS_STATE_CODES` (49 codes, incl. DC, excl. AK/HI) so each state's own `/recent` + `/recent/notable` is requested and the per-state richness reaches the DB.
- **Pacing + cost:** 49 states × 2 calls ≈ 98 eBird calls/run vs 1 today. Injectable `paceMs` defaults to ≥1000ms (precedent `cli.ts:323`) with the first-call-skip pattern (`run-backfill.ts:64`); ~98 calls × 1000ms ≈ 98s sits well under the 900s Cloud Run job timeout, so no infra change is needed (flagged as out-of-scope per the issue). Every-30-min cron ⇒ ~4,700 eBird calls/day from one key — the pacing floor + 429 circuit-break keep that polite.
[...]

Issue #840, opened 2026-06-02
fix(ingestor): recent ingest must fan out per-state — national /recent dedups to 1-obs-per-species, starving non-AZ states


julianken:
## Problem

The daily `recent` ingest under-covers most states. Live data confirms a geographic coverage gradient — our 7-day distinct-species count vs eBird ground truth:

| State | ours (7d) | eBird (7d) | coverage |
|---|---|---|---|
| US-AZ | 315 | 329 | 96% |
| US-NY | 224 | 286 | 78% |
| US-TX | 187 | 402 | 47% |
| US-NE | 56 | 256 | 22% |
| US-KS | 29 | 229 | 13% |
| US-IA | 27 | 206 | 13% |
| US-OK | 25 | 212 | 12% |
| US-MO | 15 | 214 | 7% |

Central/plains states are nearly empty in the recent window; Arizona (the project's original single-state home) is near-complete. This is user-visible: zoomed out, central states show no recent data; zoomed in, counts are anomalously low (e.g. Missouri "15 species / 7 days" in June breeding season).

## Root cause

`runIngest` (`services/ingestor/src/run-ingest.ts:38`) makes **one nationwide call**: `client.fetchRecent('US', { back: 14 })` → `GET /data/obs/US/recent` (`services/ingestor/src/ebird/client.ts:31-42`). The `recent` kind is dispatched with `regionCode: 'US'` from `cli.ts:239`.

eBird's region-`recent` endpoint **returns at most one observation per species for the entire region** (the single most-recent sighting per species). Verified directly: `GET /data/obs/US/recent?back=14` returns exactly **879 records / 879 distinct species** — one row per species, nationwide. A species present in 40 states appears **once**, attributed to whichever state filed the freshest report. High-activity regions win every tie; sparse states get almost nothing. `maxResults=10_000` is NOT the limiter — the one-per-species dedup is structural to the endpoint.

Per-state, the same window is far richer (eBird `/data/obs/US-MO/recent?back=7` = 214 species), so the data exists upstream — we just aren't requesting it per-state.

**Why 14d currently looks OK but 7d is broken:** the 14d view is served largely from older per-state *backfill* data. Those per-state backfill schedulers (`bird-ingestor-backfill-<state>`, ~49 of them) are currently **PAUSED**, so nothing replenishes per-state recent data except the national `/recent` call — which structurally can't. As backfilled data ages past the retention window without per-state replenishment, the gap spreads to 14d too.

[...]

PR #844, opened 2026-06-02
fix(db-client): UNNEST-based upsertObservations to avoid uint16 param overflow (#843)

### LINKED ISSUE #843: fix(db-client): batch upsertObservations — single INSERT exceeds Postgres 65535-param limit at fan-out scale (blocks #840 in prod)
## Problem

The per-state `recent` fan-out (#840 / PR #841) is **failing in production** — the manually-triggered post-deploy ingest (`bird-ingestor-cz6jg`, 2026-06-02) finished with `status: failure` and this error:

```
bird_ingest_run_completed  status=failure  kind=recent
firstError: "bind message has 54695 parameter formats but 0 parameters"
```

The transaction rolled back, so **no data landed — the central-states coverage bug (#840) is still live in prod** (re-measured post-ingest: MO=15, KS=29, OK=25 species/7d, unchanged from baseline; eBird ground truth ~210–230).

## Root cause

`upsertObservations` (`packages/db-client/src/observations.ts:18-52`) builds **one single `INSERT … VALUES` statement with a `$N` placeholder per field** — 9 columns × N rows. The Postgres wire protocol encodes the Bind message's parameter count as a **uint16 (max 65,535)**. node-postgres does not guard this — past 65,535 the count **silently overflows mod 65536**, desyncing the parameter-format header from the actual value list. That is exactly the observed error: **13,359 rows × 9 = 120,231 params; 120,231 mod 65,536 = 54,695** → `"54695 parameter formats but 0 parameters"` (the format count overflowed, the value count desynced to 0). (Credit: this mechanism was derived in bot review — the literal error number reverse-solves to ~13.3k aggregated rows.)

**Critical implication for the fix:** 7,281 rows (65,535÷9) is a **corruption threshold, NOT a safe ceiling to stay just under** — exceeding it doesn't cleanly reject, it silently overflows/corrupts. Chunk well below it (the `MAX_ROWS_PER_INSERT = 5000` below → 45,000 params, comfortable headroom), or use UNNEST so the ceiling is structurally unreachable.

The old national `/recent` call returned only ~879 rows, so this never triggered. The #840 per-state fan-out aggregates **tens of thousands** of rows across 49 states (the failing run had ~13.3k) → the single-statement build overflows the uint16 param count and the bind fails. This is a **latent un-batched-insert bug** (the single-INSERT pattern predates #840, introduced in #9) that #840 exposed; #840's fan-out logic is correct, but it can't land data until the upsert is batched.

**The `stampPlaceholders` UPDATE at `observations.ts:64-80` has the SAME unbounded `($N,$N)` build (2 params/row → overflows at 32,767 rows) and MUST be fixed too — not "audited," fixed.** If chunking: chunk the stamp UPDATE on the *same batch boundaries* as the INSERT, inside the *same wrapping transaction*. If UNNEST: convert it too.

## Fix — chunk the upsert (or switch to UNNEST)

**File:** `packages/db-client/src/observations.ts`, `upsertObservations`.

Two valid approaches — **UNNEST is preferred** (it makes the param ceiling structurally unreachable: one array param per column = 9 params total regardless of row count, so there is no threshold to get wrong). Chunking is acceptable if UNNEST's type-casts prove awkward. Either way keep the existing `ON CONFLICT … is_notable OR` semantics exactly:

1. **Chunk the rows** — split `inputs` into batches of a safe size (≤ 7000 rows at 9 cols, leaving headroom; suggest a named const like `MAX_ROWS_PER_INSERT = 5000`), run one INSERT per chunk **inside a single transaction** (BEGIN/COMMIT so the whole ingest is still atomic — a mid-batch failure must not leave a half-written state), sum the affected counts. Apply the same chunking to the `stampPlaceholders` UPDATE (it has the same unbounded build — see root cause).
2. **UNNEST-based insert** — `INSERT … SELECT * FROM unnest($1::text[], $2::numeric[], …)` passes one array param per COLUMN (9 params total, regardless of row count), sidestepping the limit entirely. Cleaner, but verify column types/casts match the schema and the `ON CONFLICT` clause still works.

Preserve exactly: the `ON CONFLICT (sub_id, species_code) DO UPDATE` with `is_notable = observations.is_notable OR EXCLUDED.is_notable`, `ingested_at = now()`, and the batch-scoped stamp UPDATE's O(batch) (not O(table)) behavior.

## Verification
- Unit test (real testcontainer Postgres, no mocks): upsert a row count **> 7,281** (e.g. 14,000 synthetic observations — above the failing run's ~13.3k) in one `upsertObservations` call — currently overflows/throws, must succeed after the fix and return the correct affected count. NOTE: the existing 6K-row test at `observations.test.ts:228` deliberately bypasses the real per-row upsert path — this new regression test MUST exercise the real `upsertObservations(pool, inputs)` path (the slow one), or it won't catch the bug.
- Cross-chunk `is_notable` OR-coalesce assertion: if the SAME (sub_id, species_code) appears in two different chunks (one notable, one not), the final row must be `is_notable=true` — prove chunk boundaries don't break the OR-coalesce.
- Test that chunk boundaries don't drop/double-count rows (e.g. 10,001 rows → exactly 10,001 distinct upserted).
- Existing `observations.test.ts` upsert/notable/ON-CONFLICT tests still green.
- `npm run build`/`test`/`lint` green across `packages/db-client` + `services/ingestor`.
- **Post-deploy validation (the real proof):** after merge+deploy, trigger `gcloud run jobs execute bird-ingestor`, confirm `status: success`, then verify `GET /api/observations?state=US-MO&since=7d` distinct species jumps from 15 toward eBird's ~214. (Can't verify in CI; note in PR.)

## Out of scope
- The #840 fan-out logic itself (correct; already merged).
- Read-api / frontend.

julianken:
## Diagrams

The bug and the fix, side by side — the param count that crosses the wire per upsert:

```mermaid
flowchart TB
    subgraph before["BEFORE — single INSERT … VALUES ($1..$N), 9 params/row"]
        A1["#840 fan-out aggregates ~13.3k rows"] --> A2["build 13,359 × 9 = 120,231 bind params"]
        A2 --> A3{"> 65,535 uint16 ceiling?"}
        A3 -->|yes| A4["node-postgres overflows mod 65536<br/>120,231 mod 65,536 = 54,695"]
        A4 --> A5["bind desyncs: 'bind message has 54695<br/>parameter formats but 0 parameters'"]
        A5 --> A6["tx ROLLBACK — no data lands<br/>MO/KS/OK coverage fix stays dark"]
    end
    subgraph after["AFTER — UNNEST, one array param per COLUMN"]
        B1["any row count"] --> B2["INSERT … SELECT * FROM unnest($1::text[] … $9::bool[])<br/>= 9 params total, always"]
        B2 --> B3["stamp UPDATE … FROM unnest($1::text[], $2::text[])<br/>= 2 params, batch-scoped O(batch)"]
        B3 --> B4["BEGIN → insert → stamp → COMMIT on one client<br/>ceiling structurally unreachable"]
    end
```

## Summary

- **Why it broke:** `upsertObservations` built one `INSERT … VALUES` with a `$N` placeholder per field (9 cols/row). Postgres encodes a Bind message's parameter count as a **uint16 (max 65,535)** and node-postgres does not guard it — past 65,535 the count silently overflows `mod 65536` and the bind desyncs. The merged #840 per-state `recent` fan-out aggregates tens of thousands of rows into ONE upsert call (~13.3k in the failing prod run → 120,231 params → `120,231 mod 65,536 = 54,695`), so the post-deploy ingest failed with `bind message has 54695 parameter formats but 0 parameters` and rolled back — the central-states coverage fix never landed. **7,281 rows (65,535÷9) was a corruption threshold, not a safe ceiling.**
- **The fix (UNNEST — preferred):** rewrote the INSERT as `INSERT … SELECT * FROM unnest($1::text[], …, $9::bool[])` and the stamp UPDATE's `(VALUES …)` set as `unnest($1::text[], $2::text[])` — **one array param per column**, so the total stays at 9 (insert) / 2 (stamp) params regardless of row count and the ceiling is structurally unreachable. Casts match the `observations` schema exactly (TEXT / float8 / timestamptz / int / bool; `loc_name` + `how_many` nullable). The **stamp UPDATE was fixed too** (it had the same unbounded 2-params/row build, overflowing at ~32,767 rows) and keeps its batch-scoped O(batch) behavior.
- **Atomicity:** both statements run inside one `BEGIN/COMMIT` on a single pooled client (acquire via `pool.connect()`, `ROLLBACK` + rethrow on error, `release()` in `finally`) so a mid-upsert failure cannot half-write a national map. `ON CONFLICT (sub_id, species_code) DO UPDATE … is_notable = observations.is_notable OR EXCLUDED.is_notable, ingested_at = now()` preserved verbatim.

## Screenshots

N/A — not UI (db-client-only change; no `frontend/**` files touched).

## Test plan

- [x] `npm run build` across shared-types + db-client + ingestor — clean
- [x] `npm run lint` — green (db-client/ingestor have no lint script; `--if-present` skips, matching CI; the lint-job token guard scans only `frontend/src/`, untouched)
- [x] New regression tests in `observations.test.ts` drive the **real** `upsertObservations(pool, inputs)` path (NOT the `generate_series` bypass the existing 6K LIMIT-cap test uses):
  - **14,000 rows** in one call — confirmed failing first (`bind message has 60464 parameter formats but 0 parameters`, = `14,000 × 9 mod 65,536`), green after the fix, returns the correct affected count, and all 14k rows are silhouette-stamped (proves the stamp UPDATE ran across the whole batch).
  - **14,001 rows** — exactly 14,001 distinct upserted, no drop/double-count.
  - **is_notable OR-coalesce at fan-out scale** — same `(sub_id, species_code)` arriving non-notable then notable across two >7,281-row upsert calls ends `is_notable=true`; a later non-notable upsert does NOT clobber it (OR, not assignment). (In-batch dedup of a repeated key remains the caller's job — the #840 `run-ingest.ts` `byKey` map — since a single statement, old or new, rejects two rows sharing the conflict key; UNNEST does not change that contract.)
- [x] Full `@bird-watch/db-client` suite — 14 files, **118 passed (1 todo)**; existing upsert/notable/ON-CONFLICT/stamp-scope tests stay green
- [x] Full `@bird-watch/ingestor` suite — 23 files, **222 passed**. (The connection-path change initially turned the `test` gate **RED**: the pre-existing `run-backfill.test.ts` failure-injection test wrapped only `pool.query` in a Proxy, so the new `pool.connect()` → `client.query` path bypassed the synthetic INSERT failure and `summary.status` stayed `'success'` → `1 failed | 22 passed` test files. **Fixed in this branch** — `test(ingestor): mock pool.connect path in run-backfill failure-injection test`: the Proxy now also intercepts `connect`, returning a Proxy over the real `PoolClient` whose `query` applies the same observations-INSERT reject rule, so the failure fires on the new transaction shape and the `status='failure'` assertion stays meaningful. Verified by falsification against the new db-client build — the pre-fix Proxy reproduces the exact CI error `expected 'success' to be 'failure'`, the new Proxy makes it pass.)
- [ ] New Playwright e2e spec — N/A (no user-visible behavior change)
- [ ] **Post-deploy validation (the real proof, confirmable only after merge+deploy+ingest):** trigger `gcloud run jobs execute bird-ingestor`, confirm `status: success`, then verify `GET /api/observations?state=US-MO&since=7d` distinct species jumps from **15 toward eBird's ~214**. Cannot run in CI.

## Plan reference

Out of plan — production hotfix for #843 (latent un-batched-insert bug, introduced in #9, exposed by the merged #840 / #841 per-state fan-out). Closes #843.

---

### DIFF
--- a/packages/db-client/src/observations.test.ts
+++ b/packages/db-client/src/observations.test.ts
@@ -142,6 +142,111 @@ describe('upsertObservations', () => {
       expect(r.silhouette_id).toBeNull();
     }
   });
+
+  // #843 — Postgres bind-message param overflow regression guard.
+  //
+  // The old single-`INSERT … VALUES ($1..$N)` build emitted one $N placeholder
+  // per field (9 cols/row). The Postgres wire protocol encodes the Bind
+  // parameter count as a uint16 (max 65,535); node-postgres does NOT guard it,
+  // so once total params cross 65,535 the count silently overflows mod 65536
+  // and the bind desyncs ("bind message has N parameter formats but 0
+  // parameters"). 65,535 ÷ 9 = 7,281 rows is a CORRUPTION THRESHOLD, not a safe
+  // ceiling. The #840 per-state fan-out aggregates tens of thousands of rows
+  // into ONE upsert call (the failing prod run had ~13.3k), which is exactly
+  // what this exercises. These tests MUST drive the REAL upsertObservations
+  // path (the per-row JS build + correlated stamp UPDATE), NOT the
+  // generate_series bypass the LIMIT-cap test above uses — that bypass skips the
+  // build site that overflows and would never catch this bug. The path is slow
+  // for 14k rows; that is inherent to exercising the real code.
+  function makeSynthetic(n: number, opts?: { startAt?: number }): ObservationInput[] {
+    const startAt = opts?.startAt ?? 0;
+    const out: ObservationInput[] = [];
+    for (let i = startAt; i < startAt + n; i++) {
+      out.push({
+        subId: `S-bulk-${i}`,
+        speciesCode: 'vermfly',
+        comName: 'Vermilion Flycatcher',
+        lat: 31.72 + (i % 1000) * 0.0001,
+        lng: -110.88 - (i % 1000) * 0.0001,
+        obsDt: '2026-04-15T08:00:00Z',
+        locId: `L-bulk-${i}`,
+        locName: i % 7 === 0 ? null : `Loc ${i}`,
+        howMany: i % 5 === 0 ? null : (i % 4) + 1,
+        isNotable: false,
+      });
+    }
+    return out;
+  }
+
+  it('upserts 14,000 rows in one call without overflowing the uint16 bind-param limit (#843)', async () => {
+    const inputs = makeSynthetic(14_000);
+    const count = await upsertObservations(db.pool, inputs);
+    expect(count).toBe(14_000);
+
+    const { rows } = await db.pool.query<{ n: string }>(
+      "SELECT count(*)::text AS n FROM observations WHERE sub_id LIKE 'S-bulk-%'"
+    );
+    expect(Number(rows[0]!.n)).toBe(14_000);
+
+    // The stamp UPDATE (also previously unbounded, 2 params/row) must have run
+    // across the whole batch — every row gets its silhouette_id.
+    const { rows: stamped } = await db.pool.query<{ n: string }>(
+      "SELECT count(*)::text AS n FROM observations WHERE sub_id LIKE 'S-bulk-%' AND silhouette_id = 'tyrannidae'"
+    );
+    expect(Number(stamped[0]!.n)).toBe(14_000);
+  }, 120_000);
+
+  it('upserts 14,001 distinct rows with no drop or double-count across boundaries (#843)', async () => {
+    const inputs = makeSynthetic(14_001);
+    const count = await upsertObservations(db.pool, inputs);
+    expect(count).toBe(14_001);
+
+    const { rows } = await db.pool.query<{ total: string; distinct: string }>(
+      `SELECT count(*)::text AS total,
+              count(DISTINCT (sub_id, species_code))::text AS distinct
+       FROM observations WHERE sub_id LIKE 'S-bulk-%'`
+    );
+    expect(Number(rows[0]!.total)).toBe(14_001);
+    expect(Number(rows[0]!.distinct)).toBe(14_001);
+  }, 120_000);
+
+  it('OR-coalesces is_notable across large upsert calls at fan-out scale (#843)', async () => {
+    // The `is_notable = observations.is_notable OR EXCLUDED.is_notable`
+    // ON-CONFLICT clause must survive the UNNEST rewrite at >7,281-row scale.
+    // Callers (the #840 fan-out) pass a per-call deduplicated batch — Postgres
+    // rejects two rows sharing the conflict key inside ONE statement ("ON
+    // CONFLICT DO UPDATE command cannot affect row a second time"), exactly as
+    // the old single-INSERT form did, so the in-batch dedup is the caller's job
+    // (run-ingest.ts byKey map). What this guards is the CROSS-CALL coalesce: a
+    // key that arrives non-notable in one big upsert and notable in the next
+    // must end notable, with both calls clearing the uint16 ceiling.
+    const conflictKey = { subId: 'S-notable-x', speciesCode: 'annhum' as const };
+    const notNotable: ObservationInput = {
+      ...conflictKey, comName: "Anna's Hummingbird",
+      lat: 32.3, lng: -110.99, obsDt: '2026-04-15T08:00:00Z',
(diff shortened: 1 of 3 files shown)


julianken:
@Mergifyio queue

PR #822, opened 2026-06-01
read-api crash-hardening: 512Mi memory + pg timeouts + DB-probing /health (#821)

julianken:
## Diagrams

Three independent guards on the read-api OOM-kill failure path. The root cause: a low-zoom national `/api/observations` aggregated query buffers a large rowset in-heap, peaking 257–352 MiB against the 256 Mi container cap; the OOM kill severs the pg connection, which surfaces to the user as the full-screen "Couldn't load bird data" 503.

```mermaid
flowchart TD
    A[Low-zoom national /api/observations] --> B{read-api container<br/>memory headroom?}
    B -- "256Mi cap (before)" --> C[OOM kill mid-request]
    C --> D[pg connection severed]
    D --> E[user-facing 503<br/>Couldn't load bird data]

    G1["Guard 1 (infra)<br/>cap → 512Mi"] -.restores headroom.-> B
    G2["Guard 2 (db-client)<br/>statement_timeout 15s"] -.runaway query fails fast<br/>SQLSTATE 57014.-> A
    G3["Guard 3 (read-api)<br/>/health probes SELECT 1"] -.detects severed conn,<br/>returns 503 not false-200.-> D
```

## Summary

- **Why:** The user-facing 503 is a recurrence risk: the read-api OOM-kills when a national aggregated query buffers a large rowset against a near-zero-headroom 256 Mi cap (ops-log-forensics report §3). The `pg` "connection terminated" log lines are second-order to the OOM, not an independent DB fault. This bundles the three verified P0/P1 crash-hardening quick-wins from #821.
- **Guard 1 (P0, stop-the-crash):** doubles the read-api Cloud Run memory cap 256Mi → 512Mi, restoring headroom. admin-api is deliberately untouched (not on the user-facing path).
- **Guards 2 & 3 (defense-in-depth):** a 15s `statement_timeout` makes a runaway query fail fast with a clean `pg` error (SQLSTATE 57014) instead of holding its connection until OOM; `/health` now round-trips `SELECT 1` so an uptime probe reports 503 (not a false 200) when the connection is gone.

## Screenshots

N/A — not UI. The diff is infra (`infra/terraform/`) + backend (`packages/db-client/`, `services/read-api/`); no `frontend/**` files touched.

## Test plan

[...]

PR #846, opened 2026-06-02
fix(db-client): SET LOCAL statement_timeout=0 in upsert txn + fetchNotable maxResults (#845)

### LINKED ISSUE #845: fix(db-client): SET LOCAL statement_timeout=0 in upsert txn — 15s cap cancels the national fan-out upsert (final blocker for #840 coverage)
## Problem

The per-state `recent` ingest (#840/#841 + UNNEST upsert #843/#844, both merged) **still fails in prod and lands no data** — Missouri is still 15 species/7d vs eBird's ~214. A full write-path scale audit (across pg timeouts, job runtime, memory, eBird/DB caps + 2 adversarial verifications) found this is the **single remaining will-block**:

`upsertObservations` runs its UNNEST INSERT + stamp UPDATE on a pooled client that inherits the **`statement_timeout: 15_000`** default from `packages/db-client/src/pool.ts:35` (added by #822 to protect the *read-api* from runaway queries). The ~13,300-row national upsert exceeds 15s on `db-g1-small`, Postgres fires SQLSTATE **57014 `canceling statement due to statement timeout`**, the transaction rolls back, zero rows commit.

**Confirmed reproduced** across 5 post-`0803fa8` executions (`p9fv7`, `f2gnf`, `tj2r4`, `qglfw`, `cqglm`) — all `firstError: "canceling statement due to statement timeout"`, `status: failure`, no rows landed.

The audit explicitly **ruled out** other scale blockers (so this should be the last one): species_meta invariant won't trip (full global taxonomy ~17.8k rows loaded since #530), no OOM (~110MB peak vs 512Mi limit), 900s job timeout has 600s+ headroom, handler.ts already fixed in #841.

## Fix — `SET LOCAL statement_timeout = 0` inside the upsert transaction (MUST)

**File:** `packages/db-client/src/observations.ts`, `upsertObservations`.

Immediately after `await client.query('BEGIN')`, add:
```js
await client.query('SET LOCAL statement_timeout = 0');
```
- `SET LOCAL` is **transaction-scoped** — it applies only to the INSERT + stamp UPDATE in this transaction and reverts on COMMIT/ROLLBACK.
- This is deliberately **narrower than** raising the pool-level `statement_timeout` for the ingestor: the other ingestor queries (`startIngestRun`, `findMissingSpeciesMeta`, `finishIngestRun`) correctly KEEP the 15s session default, and the read-api pool (separate process, never calls `upsertObservations`) is untouched. `createPool`'s 15s default stays — new callers keep protection.
- The Cloud Run job's 900s timeout remains the outer kill switch (the audit measured ~115-137s expected total, comfortably inside it).
- **Do NOT** instead disable statement_timeout at the ingestor pool level (would strip the guard from the fast ingestor queries that should keep it), and **do NOT** chunk the upsert in this PR (chunking also fixes it but adds per-chunk transaction/partial-success/retry complexity — file it as a separate follow-on architectural issue; this one-line fix gets data flowing now).

### `fetchNotable` missing maxResults (data quality)
**File:** `services/ingestor/src/ebird/client.ts`, `fetchNotable` (~L45-49). It sets `back` + `detail=simple` but NOT `maxResults`, so eBird defaults to 100 results. In busy states at peak migration (TX/FL/CA, 200-500 notable sightings/14d), notable observations past the top-100 get `is_notable=false` silently — the row lands but the flag is wrong. Add `url.searchParams.set('maxResults', '10000')` mirroring `fetchRecent` (client.ts:37).

### Stale pacing comment (correctness)
**File:** `services/ingestor/src/cli.ts` (~line 243), near the `recent` `runIngest` call. The "~98 calls × 1s ≈ 98s" comment conflates 98 HTTP calls with the ~48 per-state inter-round pauses and is inaccurate — fix to reflect ~48 inter-round pauses × 1s ≈ 48s pacing, with each state firing recent+notable concurrently per round. Comment-only.

## Out of scope (deferred follow-ups — file separately, do NOT do here)
- **M2:** downgrade the species_meta invariant from `throw`-and-abort to filter-unknown-codes-and-warn (matters at the ~Oct 2026 eBird annual taxonomy release, not now).
- **M3:** AbortController wall-clock guard on the fetch phase (defends against a 49-state timeout storm exceeding 900s before clean shutdown).
- **Chunked/per-state upsert** as an architectural improvement (reduces per-statement duration + heap peak).
- `fetchRecent` 10k server-side cap (eBird hard ceiling; only bites TX/FL/CA at extreme peak; document, don't fix here).

## Verification
- Unit test (real testcontainer Postgres, no mocks): upsert a large batch (≥14,000 rows) AND set a deliberately tiny session `statement_timeout` on the test pool (e.g. 100ms), then assert `upsertObservations` SUCCEEDS — proving the `SET LOCAL 0` overrides the session cap for the transaction. Without the fix this test fails with 57014.
- Confirm the `SET LOCAL` does NOT leak: after the upsert completes, acquire a NEW connection from the pool (or run a query that lands on a different/reset session) and assert its `statement_timeout` is back to the session default — the test must check a connection state AFTER the transaction closed, or it proves nothing about the revert.
- `fetchNotable` test asserts the URL includes `maxResults=10000`.
- `npm run build`/`test`/`lint` green across `packages/db-client` + `services/ingestor`.
- **Post-deploy proof (the real validation, noted for the operator — can't run in CI):** after merge+deploy, `gcloud run jobs execute bird-ingestor` finishes `status: success`, then `GET /api/observations?state=US-MO&since=7d` distinct species jumps from 15 toward eBird's ~214.

julianken:
## Diagrams

```mermaid
sequenceDiagram
    participant Job as Cloud Run Job (900s cap)
    participant Ingestor
    participant eBird
    participant Postgres

    Job->>Ingestor: execute (recent fan-out)
    loop per state (49) — recent+notable concurrent, ~48 inter-round 1s pauses
        Ingestor->>eBird: GET /data/obs/US-XX/recent (maxResults=10000)
        Ingestor->>eBird: GET /data/obs/US-XX/recent/notable (maxResults=10000 ← #845)
    end
    Ingestor->>Postgres: BEGIN
    Note over Ingestor,Postgres: SET LOCAL statement_timeout = 0  ← #845<br/>(txn-scoped; reverts on COMMIT/ROLLBACK)
    Ingestor->>Postgres: UNNEST INSERT ~13.3k rows + stamp UPDATE (>15s)
    Ingestor->>Postgres: COMMIT (statement_timeout back to session 15s)
```

## Summary

- **Why:** the per-state `recent` fan-out (#840) + UNNEST upsert (#844) still lands zero data in prod (MO 15 species/7d vs eBird ~214). The ~13.3k-row national upsert runs on a pooled client that inherits `pool.ts`'s 15s `statement_timeout` (#822, for the read-api), exceeds it on `db-g1-small`, and Postgres cancels it (SQLSTATE 57014) → whole txn rolls back → nothing commits. This is the **final** scale blocker (the audit ruled out species_meta invariant, OOM, and job timeout).
- **Fix (MUST):** `SET LOCAL statement_timeout = 0` immediately after `BEGIN` in `upsertObservations`. Transaction-scoped, so it exempts only the INSERT + stamp UPDATE and auto-reverts on COMMIT/ROLLBACK — every other ingestor query (`startIngestRun`, `findMissingSpeciesMeta`, `finishIngestRun`) and the separate read-api pool keep their 15s guard. No pool.ts change, no chunking.
- **Also (SHOULD):** `fetchNotable` now requests `maxResults=10000` (it inherited eBird's default-100 cap, silently leaving `is_notable` false past the top-100 in busy states); and the stale `~98 calls × 1s` pacing comment in `cli.ts` is corrected to the actual ~48 inter-round pauses.

## Screenshots

N/A — not UI

## Test plan

- [x] `npm run typecheck && npm run test` — green (db-client 120 passed/1 todo; ingestor 223 passed)
- [x] New unit / integration tests added (if behavior changed) — real testcontainer Postgres, no mocks: (1) a 14k-row upsert on a pool with a deliberately tiny 50ms session `statement_timeout` SUCCEEDS (confirmed RED first with 57014, then green); (2) **no-leak** — a FRESH connection acquired *after* the txn closes shows `SHOW statement_timeout` = `50ms`, NOT `0`; (3) `fetchNotable` request URL asserts `maxResults=10000`
- [ ] New Playwright e2e spec added (if user-visible behavior changed) — N/A, no frontend change
- [x] `npm run build` — clean production build (db-client + ingestor)
- [ ] (UI only) Playwright MCP smoke — N/A, db-client + ingestor only, no `frontend/**`

## Plan reference

Out of plan — fixes #845, the final scale blocker for the #840 per-state coverage fix. Closes #845.

---

### DIFF
--- a/packages/db-client/src/observations.test.ts
+++ b/packages/db-client/src/observations.test.ts
@@ -1,4 +1,5 @@
 import { describe, it, expect, beforeAll, beforeEach, afterAll } from 'vitest';
+import pg from 'pg';
 import { startTestDb, type TestDb } from './test-helpers.js';
 import {
   upsertObservations, getObservations, getObservationsAggregated,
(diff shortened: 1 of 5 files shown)


julianken:
@Mergifyio queue

PR #860, opened 2026-06-03
fix(#859): aggregate species into low-zoom buckets — real names + working links at every zoom

julianken:
## Diagrams

Re-architecture from **compute-on-read** (lazy per-cell fetch — buggy) to **compute-on-write** (species aggregated into the bucket server-side). Every low-zoom popover now reads real species it already holds.

```mermaid
flowchart LR
  subgraph Server["read-api / db-client"]
    SQL["getObservationsAggregated\nGROUP BY grid cell, family\ntop-8 species/family by count"]
    DICT["GET /api/species\ncode → {comName, familyCode}\n(immutable-cached ~9KB)"]
  end
  subgraph Wire["AggregatedBucket"]
    B["{lat,lng,count,speciesCount,\n families:[{code,count,speciesCount,\n   species:[{code,count}] top-8]}]}"]
  end
  subgraph Frontend
    REND["buckets → supercluster\n(cluster by count)"]
    POP["Cell / Cluster / point popover\nreads bucket.families[].species\nname ← dictionary · link ← real code"]
    DETAIL["species detail panel\n(real code → /api/species/:code)"]
  end
  SQL --> B --> REND --> POP --> DETAIL
  DICT -. names .-> POP
```

Deleted entirely (not patched): synthetic-observation fabrication, the per-click lazy fetch, `isSyntheticCode`, and the centroid-bbox geometry — the sources of the bug class.

## Summary

- **Why:** at low/national zoom the server aggregated to *family codes only*; the frontend fabricated synthetic "observations" (`comName` = the Latin family code, `speciesCode` = a dead `agg-…`), so popovers showed a Latin family name with a dead detail link, and a lazy per-cell fetch produced empty cells from centroid-vs-bucket geometry. Closes the core of #859. Full verified RCA: [#859 investigation](https://github.com/julianken/bird-sight-system/issues/859).
- **What:** the bucket now carries its own species (top-8 per family + exact `speciesCount` for "+N more"), resolved to names via a one-time `/api/species` dictionary. Real species + working detail links render uniformly at every zoom — no synthetic data, no geometry, no per-click fetch. Popovers portal to `document.body` (structural z-index fix for the marker-`transform` stacking trap). Per-family lists cap at top-8 with an active **"+N more" drill-in**; cluster families default-collapsed.
- **Cost:** national grid payload ~47 KB → **~85 KB gzip** (measured at full national cardinality), well under the considered 130–160 KB target; the family legend and lede counts stay exact (cap never leaks into them).

## Screenshots

[...]

Issue #859, opened 2026-06-03
Low-zoom (aggregated) cell/cluster popovers show a Latin family name and a dead species link instead of the real common name + working detail link


julianken:
## Symptom

At low/national zoom (map zoom < 6, the aggregated render mode), clicking a "species" row in a grid-cell or cluster popover shows a **Latin family name** (e.g. `accipitridae`) instead of the common name, and the row is **not** a clickable link to the species-detail panel. At zoom >= 6 (close range) the identical interaction correctly shows the common name and renders a working species link. The behavior must be **identical at every zoom**: click → real common name → working species-detail link, with no difference between national and close zoom.

This is not a caching or WAF artifact — it is confirmed at every layer of the stack. The aggregated grid payload simply never carries species identity, and the frontend fabricates fake species rows to fill the gap.

## Root cause (layered, all verified)

1. **Backend SQL carries only family codes per cell.** `packages/db-client/src/observations.ts`, `getObservationsAggregated` (function at line 310; SQL at lines 369–380). At zoom < 6 it groups observations into a coarse grid and emits `array_remove(array_agg(DISTINCT sm.family_code), NULL) AS families` (line 375). Each cell carries family codes and a count only — **no species code, no species name, ever**.

2. **Wire confirms it.** Live prod `GET /api/observations?...&zoom=3`: every bucket is `{lng, lat, count, speciesCount, families:[...family codes...]}`. No species identity on the wire.

3. **Frontend fabricates fake species.** `frontend/src/data/use-bird-data.ts`, `expandBucketsToSyntheticObservations` (lines 35–64) expands each bucket into synthetic `Observation` rows where `comName = family` (the family code — **this is the "Latin name" the user sees**, line 50) and `speciesCode = "agg-..."` (intentionally non-resolvable, lines 47–49).

4. **Dead-link gate.** `frontend/src/data/use-bird-data.ts` `isSyntheticCode` (lines 32–33) plus `frontend/src/data/use-species-detail.ts` line 65: any `agg-` code short-circuits the detail fetch, so no detail and no clickable link is rendered. The popover renderers — `frontend/src/components/map/ClusterListPopover.tsx` line 218 and `frontend/src/components/map/CellPopover.tsx` line 119 — render the synthetic `{s.count}x {s.comName}` (i.e. the family code) as plain text rather than a link.

[...]

PR #861, opened 2026-06-03
fix: national low-zoom 503 — optimize aggregated species query (prod incident)

julianken:
## Diagrams

National low-zoom 503 RCA + the query-plan change:

```mermaid
flowchart TB
    subgraph req["Failing prod request: gridMultiplier=2, since=14d, whole-US, no bbox/state"]
      direction LR
      A["~550k observations<br/>(~467k in 14d window)"]
    end

    subgraph before["BEFORE (#859) — ~147s, 503"]
      B1["base: 467k rows<br/>(geom→bucket, family join)"]
      B2["per_species → ranked → per_family → families<br/>(per-bucket jsonb)  ~1.7s ✅"]
      B3["FINAL: count(*) over ALL base<br/>GROUP BY (lng,lat, <b>f.families jsonb</b>)"]
      B4["jsonb blob pulled into sort key,<br/>duplicated across 467k rows<br/>→ 3.6 GB external-merge sort 💥"]
      B1 --> B2 --> B3 --> B4
      B4 -->|"> 15s statement_timeout"| B5["57014 cancel → HTTP 503"]
    end

    subgraph after["AFTER (#862) — ~2.3s ✅"]
      C1["base (unchanged)"]
      C2["families CTE (unchanged jsonb)"]
      C3["bucket_totals: count(*) + count(DISTINCT sp)<br/>GROUP BY (lng,lat) only — <b>no jsonb in key</b>"]
      C4["LEFT JOIN families onto bucket_totals<br/>(no giant sort)"]
      C1 --> C2 --> C4
      C1 --> C3 --> C4
      C4 --> C5["byte-identical result, well under 15s"]
    end

    A --> before
    A --> after
```

## Summary

- **Why it 503'd.** The coarsest national aggregated request exceeded the read-api pool's 15s `statement_timeout` (`packages/db-client/src/pool.ts`). `EXPLAIN (ANALYZE, BUFFERS)` at prod scale (~550k obs, ~467k in the 14d window, ~5.6k coarse buckets) showed the #859 final SELECT computed the bucket totals over **all** of `base` while `GROUP BY (lng_bucket, lat_bucket, f.families)`. Because `f.families` is a large per-bucket jsonb blob, the planner pulled it into the GROUP BY sort key and duplicated it across every one of the ~467k national base rows, forcing a **3.6 GB external-merge sort to disk → ~147s**. The per-family jsonb CTE chain itself was fast (~1.7s); the sort was the whole cost.
- **The fix is a pure query restructure — no index, no migration.** Split the bucket totals into their own `bucket_totals` CTE keyed only on `(lng_bucket, lat_bucket)` (no jsonb in the sort key), then `LEFT JOIN` the prebuilt families jsonb onto it. The multi-GB sort disappears. The output is **byte-identical** — `f.families` was always functionally determined by the bucket key, so grouping by it never changed which rows came back, only the cost. The 39 existing #859 nesting/ordering/NULL-family tests pass unchanged, and OLD-vs-NEW was verified multiset-equal at prod scale.
- **Before/after national timings at prod scale (~550k rows):** ~147s (parallel) / ~100s (serial) → **~2.3–3.0s**. Well under the <3s target and the 5s ceiling; ~5x under the 15s timeout even on the slower serial plan.
- **Adds the perf guard #859 was missing** (it measured payload bytes but never query time): a real-PostGIS testcontainer test seeds ~550k CONUS observations and asserts the national query finishes under 8s **and** is byte-identical on re-run. It FAILS on the pre-#862 query (~100s) and PASSES on the fix — pinning the regression. Follow-up to #860/#859.

## Screenshots

N/A — not UI (db-client query + integration test only; no `frontend/**` changes).

## Test plan

- [x] `npm run test --workspace @bird-watch/db-client` — green (15 files, 129 passed / 1 todo), incl. the new perf guard and the migrations-down-chain test
- [x] `npm run test --workspace @bird-watch/read-api` — green (170 passed)
- [x] New integration test added: `observations-aggregated-perf.test.ts` (prod-scale perf guard + byte-identical re-run). Confirmed it FAILS on `origin/main`'s query (`expected 100223 to be less than 8000`) and PASSES on the fix (~2.3–3.0s)
- [x] `npm run build` — clean
- [x] `npm run lint` — clean
- [x] `npx knip` — no new findings (exit 0; only the pre-existing `docs/plans/...prototype` hint, `knip.ts` untouched)
- [x] N/A — not UI, no Playwright MCP smoke

## Plan reference

Out of plan — production incident hotfix. Follow-up to #860 / #859 (the national species-aggregation work that introduced the per-family jsonb nesting); this PR fixes the query-time regression that #859's payload-only measurement missed.

---

### DIFF
--- a/packages/db-client/src/observations-aggregated-perf.test.ts
+++ b/packages/db-client/src/observations-aggregated-perf.test.ts
@@ -1,0 +1,293 @@
+/**
+ * PROD-SCALE PERF GUARD for the national low-zoom aggregated query (#862).
+ *
+ * Prod incident: the coarsest national aggregated request (gridMultiplier=2,
+ * since=14d, whole-US, no bbox/state) returned HTTP 503 because the #859
+ * `getObservationsAggregated` query exceeded the 15s `statement_timeout`. The
+ * prior #859 work measured PAYLOAD size but never QUERY TIME — this test pins
+ * the query-time regression that the payload tests could not see.
+ *
+ * What this guards:
+ *   1. PERF — at ~550k observations across CONUS the national query must finish
+ *      well under the timeout. The pre-#862 query body spilled a multi-GB
+ *      external-merge sort (it grouped the bucket totals by the per-bucket
+ *      `families` jsonb, duplicating the blob across every base row) and ran
+ *      ~147s. The #862 fix splits the totals into their own `bucket_totals`
+ *      CTE keyed only on (lng_bucket, lat_bucket) → ~2.3s. The threshold below
+ *      FAILS on the pre-#862 query and PASSES on the fix.
+ *   2. CORRECTNESS — the fix must not silently change results. We snapshot the
+ *      full national result on first run (sorted, with the families jsonb
+ *      canonicalised) and assert run-to-run equality, so any future change to
+ *      the CTE chain that alters buckets/counts/top-8/ordering trips the test.
+ *
+ * Determinism note: the bucket key `round(ST_X(geom)*mult)/mult` is
+ * float-valued, and Postgres' PARALLEL HashAggregate over a float group key
+ * produces run-to-run variation in which bucket a boundary-adjacent row lands
+ * in (verified during the #862 investigation — it affects the pre-#862 query
+ * identically, so it is not a correctness regression in the fix). To make the
+ * correctness snapshot reproducible we disable parallel gather on this test DB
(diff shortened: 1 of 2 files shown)


julianken:
@Mergifyio queue

Issue #711, opened 2026-05-22
feat(ingestor): post-ingestion CF cache-warm for popular bboxes


julianken:
## Summary

After each 30-min `/recent` ingest cycle invalidates the freshness of `/api/observations` responses, real-user pans into popular metros hit Cloud SQL fresh. Add a new `cache-warm` CLI kind to the bird-ingestor that pre-fetches a curated list of ~77 popular bbox URLs ~2 min after each ingest tick, hydrating Cloudflare's edge cache so the next user pan into the same area is a HIT instead of a MISS.

Expected effect: CF cache HIT rate on popular bboxes jumps from cold-on-first-user-pan to warm-by-pre-fetch. Cloud SQL CPU under HN-class spike concentrates on long-tail bboxes only (rural pans), which are infrequent.

## Background

Today (2026-05-22) we:
- Made `api.bird-maps.com` proxied through Cloudflare (#708)
- Added a Cache Rule respecting origin `s-maxage=300` (#706 + manual apply)
- Confirmed CF cache works end-to-end (5/5 unique bboxes show MISS → HIT pattern)

The cache works correctly but each bbox is cold on its first viewer. A small pre-warm step turns "popular metros take 6-12s for the first viewer after each ingest cycle" into "popular metros are <100ms from the first viewer."

### New CLI kind `cache-warm`

`services/ingestor/src/cli.ts` gets a new branch (mirror the structure of `recent`, `hotspots`, `prune`, `backfill`):

```ts
} else if (kind === 'cache-warm') {
  summary = await deps.runCacheWarm({ baseUrl: 'https://api.bird-maps.com' });
}
```

`services/ingestor/src/run-cache-warm.ts` (new file):
- Iterates a hard-coded list of bbox URLs (see below for the list)
- Issues `GET` requests sequentially (concurrency=1) with `AbortSignal.timeout(15_000)`, sleeping 200ms between requests
- Records per-URL: status code, cf-cache-status response header, duration ms
- Emits structured log `bird_ingest_cache_warmed` with `{ kind: 'cache-warm', total, missCount, hitCount, expiredCount, dynamicCount, errorCount, p50ms, p95ms }` at end

This is a pure HTTP-side helper — no DB connection needed, no eBird API calls. The `cache-warm` kind in `cli.ts` should skip the `pool = new Pool(...)` initialization to keep startup fast.

### URL list — 77 entries

**CONUS aggregated (z=3, z=4):** the two queries every user hits on initial page load
[...]

PR #867, opened 2026-06-04
perf(frontend,ingestor): quantize the /api/observations fetch bbox to a shared grid so the CDN cache (and warmer) actually hit (#866)

### LINKED ISSUE #866: perf(frontend): quantize the /api/observations fetch bbox to a shared grid so the CDN cache (and the cache-warmer) actually hit
## Problem

The CDN cache on `/api/observations` is effectively inert in production. Live Cloudflare telemetry (zone `bird-maps.com`, 30 days to 2026-06-04):

- **Zone-wide cache HIT ratio: 0.67%** (1,462 / 217,026 requests) — 99.3% of traffic reaches origin.
- **`/api/*` HIT ratio: ~5%** on the best recent day (216 / 4,307), ~60% `none/dynamic`, 27% `expired`.
- Cloud Run + Cloud SQL absorb essentially all `/api` load; the edge shields ~1–5%.

**Root cause (code-verified, structural — not a config typo):**

- The Cloudflare cache rule keys on the **raw request URI** — `infra/terraform/cache-rule.tf` `cache_key` block sets only `ignore_query_strings_order = true`, no value normalization.
- The frontend serializes the **full-precision float bbox** straight from the viewport into the query:
  - `frontend/src/api/client.ts:33` → `url.searchParams.set('bbox', f.bbox.join(','))`
  - `frontend/src/components/map/MapCanvas.tsx:1567` → `cb(map.getBounds(), Math.floor(map.getZoom()))` (zoom floored; **bbox is raw `getBounds()`**).
- Every pan/zoom therefore produces a unique cache key → **guaranteed MISS by construction**.

**The cache-warmer is orphaned.** `services/ingestor/src/run-cache-warm.ts` (#711, shipped) runs every 30 min (`2,32 * * * *`) and GETs 77 fixed URLs built from `.toFixed(2)` metro centroids ± fixed half-widths (`ZOOM_HALFW`, lines 90-94). Those keys are a **disjoint keyspace** from real requests. The only shared key is the initial paint — `DEFAULT_BBOX_CONUS = [-125, 24, -66, 50]` @ `zoom=3` (`frontend/src/App.tsx:295`) is value-identical to warm URL #1 — and even that misses on desktop, which opens at `zoom=4` against a different warmer bbox. That single key is approximately the entire 0.67% hit rate.

## Goal

Collapse cache-key cardinality by snapping the **fetch** bbox to a shared, deterministic grid consumed by **both** the frontend request path and the warmer, so that (1) nearby viewports resolve to one key (**organic hits**) and (2) warmed keys become a subset of requestable keys (**the warmer pays off**).

Target: `/api` aggregated-mode hit ratio from ~5% toward 40–70% on popular keys (organic + warmed); see the cold-window note under Gotchas for why the *initial* post-ship measurement may land lower until `s-maxage` is revisited.

**This is not a cost-reduction change** — the bill floor is the always-on Cloud SQL instance, not per-request compute. The win is origin-load reduction, latency headroom, and activating an already-built subsystem. Do not frame it as a cost win.

## Scope decision (v1 = aggregated mode only)

Snapping applies **only when `zoom < 6`** (the aggregated path, `app.ts:240`). At `zoom >= 6` (per-observation mode) the raw bbox passes through **unchanged**.

Rationale — this resolves a hard correctness conflict and concentrates the payoff:
- The per-axis bbox cap (`services/read-api/src/validate.ts:191-237`: `maxLngSpan 45`, `maxLatSpan 25`) is enforced **only at `zoom >= 6`**, and real views already sit at that boundary (`App.tsx:61-63`). **Outward** (superset) snapping of a bbox already at 45° would push it past the cap → a 400 "bbox too large" where the raw request passed. Outward-snapping and the cap are mutually unsatisfiable at the boundary, so per-observation snapping is **out of v1 scope** (tracked as a follow-up; would require clamp-to-cap logic, smaller steps, and its own tests).
- The aggregated path is where the cache win is largest anyway: CONUS initial load + multi-state metros, coarse payloads, and the warmer's highest-value entries (CONUS z3/z4 + metro z5).

## Approach

1. **Canonical snap function.** Add a pure `snapFetchBbox(bbox, zoom) → bbox` (and its string serializer) to a new **`packages/geo`** workspace (importable by both `frontend/` and `services/ingestor/`; pure + unit-tested). For `zoom >= 6` it is a passthrough (returns the input bbox unchanged in v1). For `zoom < 6` it rounds each axis **outward** (floor W/S, ceil E/N) to `SNAP_STEP_DEG[tier]`, so the snapped bbox is always a **superset** of the input — we never under-fetch.

2. **Grid step (numeric, lattice-aligned to the server buckets).** The server bucket size is `1 / gridMultiplier` (`app.ts:241`). The request snap step is a small integer multiple of that bucket size, chosen so payloads stay bounded and the step is exactly representable in 2 decimals:

   | zoom tier | server `gridMultiplier` | server bucket | **`SNAP_STEP_DEG`** | notes |
   |---|---|---|---|---|
   | `z ≤ 3` | 2 | 0.5° | **1.0°** | CONUS; default bbox is integer-aligned → snaps to itself |
   | `z = 4` | 4 | 0.25° | **0.5°** | multi-state; 2× bucket |
   | `z = 5` | 8 | 0.125° | **0.25°** | metro; 2× bucket; round-trips through `.toFixed(2)` |

   All steps are multiples of 0.25° → exact under `.toFixed(2)`. (These are the proposed values; the implementer may tune within the constraint *step = integer × bucket size AND representable in 2 decimals*.)

3. **Canonical serializer (the actual collision lever).** `snapFetchBbox` emits one canonical decimal string via **`.toFixed(2)`** (matching the warmer's existing format, `run-cache-warm.ts:121-122`). **Both** `client.ts` and `run-cache-warm.ts` produce the query value through this single function, so the serialized values are identical. Because `cache-rule.tf:40` already sets `ignore_query_strings_order = true`, the required invariant is **same param-set + same values, order-independent** — not literal whole-URL byte-identity.

4. **Frontend consumes it — in `client.ts` `getObservations` (fetch-time), NOT in `App.tsx` state.** Snap the bbox as the request is built. This is deliberate: the #847 scope-change reseed sets `debouncedBbox` from the **raw** `scopeBounds` envelope (`App.tsx:497`), so snapping at the state layer would leave the *first fetch of every scope change* unsnapped — a guaranteed warmer miss on the single most important transition. Snapping at fetch-time covers the reseed path and the normal idle path uniformly. The map keeps rendering the true viewport; only the fetch bbox is snapped, and because it is a superset MapLibre clips off-screen data → no visible change.

5. **Warmer consumes it.** Route the warmer's **aggregated** entries (CONUS z3/z4 + metro z5) through the same `snapFetchBbox`, so those warmed URLs are value-identical to what a client viewing that anchor requests; fix the desktop `zoom=4` CONUS default so the most common first paint warms in both layouts. The warmer's `z6/z7` per-observation entries are retained but remain disjoint until the per-observation follow-up (a known, bounded v1 limitation). Keep the rate-limit safety margin (concurrency=1, 200 ms) intact.

## Files

- **`packages/geo`** — new workspace: `snapFetchBbox` + serializer + tests (consumed by frontend **and** ingestor). Export from `packages/geo/src/index.ts` (**consumer-compiled, source-only**) so no `dist` build is needed and the hand-ordered root `package.json` `build` script needs no new entry. (If `geo` instead ships a `tsc` `dist/` build like `shared-types`, it must be inserted **before** its consumers in that explicit build list or the `build` CI gate goes red.) Wire into root `package.json` workspaces + `tsconfig`.
- `frontend/src/api/client.ts` — consume `snapFetchBbox` in `getObservations` (fetch-time; **not** `App.tsx` state — see Approach §4).
- `services/ingestor/src/run-cache-warm.ts` — `buildCacheWarmUrls()` consumes `snapFetchBbox` for the aggregated tiers; fix the desktop-z4 CONUS default.
- Tests that encode the old shape — **update, do not delete**: `services/ingestor/src/run-cache-warm.test.ts` (the `total === 77` assertion ~line 15 **and** the exact-URL assertions ~lines 34-45) **and** `services/ingestor/src/cli.test.ts` (the second `total === 77` site, ~lines 537-591).
- No changes to `cache-rule.tf`, the SQL aggregation query, `gridMultiplier`, or the DB.

## Acceptance criteria

- **Cardinality-collapse test (the real lever):** ≥100 distinct raw-float viewports sampled from a realistic `z=5` pan distribution over a 1°×1° region collapse to **≤ 25** distinct snapped query-value strings — the lattice bound `(1° / 0.25° + 1)² = 25` — i.e. an order-of-magnitude reduction from the ~100 raw keys. Plus a **sub-cell invariance** case: viewports whose four edges all fall within one `SNAP_STEP_DEG` interval map to a **single** key. Analogous bounds at `z=3` (1.0° step) / `z=4` (0.5° step). (An absolute `≤2` is wrong — a 1° region spans ~4 cells/axis at a 0.25° step; the gate must be the lattice bound, not a fixed small constant.)
- **Warmer/client agreement test:** for a representative metro anchor at `z=5`, the **full query param-set** the warmer emits equals what the frontend builds for a viewport centered there (default scope) — `bbox` (snapped value), `zoom`, **and** `since=14d`, with no filter params. Assert the whole param-set, not just bbox: `client.ts:29` emits `since` only when truthy (the default `state.since='14d'` is truthy, so it is emitted), and any missing/extra param forks the key.
- **Superset property test:** `snapFetchBbox(b, z)` output ⊇ `b` for randomized inputs at `z < 6`; **passthrough** test confirming `z >= 6` returns the input bbox unchanged.
- **No visual regression:** driven live through Playwright MCP at the 5 canonical viewports × 2 themes (`CLAUDE.md` Testing) — edge observations still render, zero console errors/warnings.
- `read-api` + frontend unit suites green; the three warmer/cli contract assertions above updated and passing.
- **Measurement plan:** record `/api` aggregated-mode HIT ratio (`cf-cache-status` sampling / Cloudflare GraphQL `httpRequestsAdaptiveGroups`, 1-day windows on Free plan) before merge and ~48 h after, **sampled at matching clock hours** (traffic is diurnal); report before/after in the PR.

## Displayed counts are already viewport-clipped (no work needed)

The superset fetch does **not** inflate the lede / family-legend "in view" totals. Those counts derive from `filterBucketsByBounds(buckets, viewportBounds)` (`App.tsx:572-576`, `frontend/src/lib/viewport-filter.ts`), where `viewportBounds` is the **raw** map bounds — never the snapped fetch bbox. Extra buckets returned by the snapped superset fall outside `viewportBounds` and are filtered out before counting. Snapping therefore changes only the *fetched* set (for cache reuse), not what the user sees counted. No count-clipping work is required — add a one-line comment at the snap site pointing at this existing clip so a future reader doesn't re-derive the concern.

## Out of scope / non-goals

- **Per-observation (`zoom >= 6`) snapping** — deferred (needs clamp-to-cap; see Scope decision). The warmer's z6/z7 entries stay disjoint until then.
- No SQL / `gridMultiplier` / DB changes.
- No edge-Worker or Cloudflare cache-key Transform — Free-plan cache keys can't arithmetically round a value, a Worker adds cost/complexity, and edge normalization would neither reduce origin DB load nor activate the warmer. Request-side snapping does both.
- Filtered views (`notable`/`species`/`family`/`state`) won't hit the warmer but still gain organic snapping; broadening warmer coverage to filters is a follow-up.

## Gotchas

- **Round outward (superset).** Rounding inward would under-fetch and drop edge observations.
- Zoom is already floored at `MapCanvas.tsx:1567`; keep integer zoom in the key.
- Snapped value format must round-trip the step exactly — keep steps as 0.25° multiples so `.toFixed(2)` is lossless on both call sites.
- Mode boundary at `zoom 6` (`app.ts:240`): snapping is a no-op at/above it in v1; the warmer straddles it (z5 aggregated → snapped; z6/z7 per-obs → unchanged).
- **Hit-rate expectation (cold-window arithmetic).** With `s-maxage=300` + `stale-while-revalidate=600` and the 30-min warmer cadence, a *warmed* key is HIT-serving (~5 min fresh + ~10 min stale) for ~15 of every 30 min → the warmer's contribution has a ~50% duty cycle. **Organic** (user-to-user) hits are not cadence-bound — any key re-requested within its 15-min fresh+SWR window stays hot. Expect the first measurement to land *below* the 40–70% aspiration until `s-maxage` is raised toward the ingest cadence — a follow-up in `services/read-api/src/cache-headers.ts`, **out of scope here** (measure the snapping baseline first).

## Evidence / references

- Telemetry pulled 2026-06-03: Cloudflare GraphQL 30-day zone HIT 0.67%; a live `cf-cache-status` demo on a fixed-bbox URL returned `MISS` then `HIT (age:2)` on repeat — the edge cache works; the low rate is purely float-bbox cardinality.
- Code: `client.ts:33`, `MapCanvas.tsx:1567`, `cache-rule.tf` `cache_key`, `run-cache-warm.ts:90-127`, `app.ts:240-241`, `App.tsx:61-63,295`, `validate.ts:191-237`.
- Context: warmer = #711 (closed/shipped); aggregation model = #859/#860.

julianken:
## Diagrams

```mermaid
flowchart LR
    subgraph Frontend
      VP["raw float viewport<br/>getBounds()"]
    end
    subgraph geo["@bird-watch/geo (new, source-only)"]
      SNAP["snapFetchBbox(bbox, zoom)"]
      SER["serializeBbox → .toFixed(2)"]
    end
    subgraph Warmer["ingestor cache-warmer"]
      WB["buildCacheWarmUrls()"]
    end

    VP -->|"zoom < 6: floor W/S, ceil E/N to grid<br/>zoom ≥ 6: passthrough"| SNAP
    WB -->|"CONUS z3/z4 + metro z5"| SNAP
    SNAP --> SER
    SER -->|"identical canonical bbox value"| KEY["one shared CDN cache key"]
    KEY --> CF["Cloudflare edge HIT<br/>(organic + warmed)"]

    note["z3 step 1.0° · z4 step 0.5° · z5 step 0.25°<br/>each = integer × server bucket, exact under .toFixed(2)"]
```

```mermaid
sequenceDiagram
    participant Client as Frontend (client.ts)
    participant Geo as @bird-watch/geo
    participant CDN as Cloudflare
    participant Origin as Read API + Cloud SQL

    Note over Client,Origin: BEFORE — full-precision float bbox ⇒ unique key every pan ⇒ MISS
    Client->>Geo: snapFetchBbox([-118.241,33.998,-107.237,40.051], 5)
    Geo-->>Client: [-118.25,33.75,-107.00,40.25] → "-118.25,33.75,-107.00,40.25"
    Client->>CDN: GET /api/observations?since=14d&bbox=<snapped>&zoom=5
    CDN->>Origin: MISS (cold) — then warmer + neighbours reuse the SAME key
    CDN-->>Client: HIT on repeat / warmed key
```

## Summary

- **Why:** the CDN cache on `/api/observations` was inert (zone-wide HIT 0.67%) because the frontend serialized the full-precision float viewport bbox straight into the query string — every pan/zoom minted a unique Cloudflare cache key, a guaranteed MISS by construction. The 77-URL cache-warmer (#711) keyed on a disjoint keyspace, so it paid off for ~one key.
- **What:** a new source-only `@bird-watch/geo` workspace exposes `snapFetchBbox(bbox, zoom)` (rounds the *fetch* bbox OUTWARD to a shared coarse grid for the aggregated path `z<6`; passthrough at `z>=6`) and a canonical `.toFixed(2)` serializer. **Both** the frontend (`client.ts`, fetch-time) and the warmer (`run-cache-warm.ts`, aggregated tiers) emit the bbox through this one function, so nearby viewports collapse to one cache key and warmed keys become a subset of requestable ones. The warmer's CONUS entries now derive from `DEFAULT_BBOX_CONUS` snapped at both z3 and z4, so the initial paint warms in both mobile and desktop layouts.
- **Measured collapse (new unit test):** 150 jittered z5 viewports over a 1°×1° region → **20** distinct snapped keys (≤ the `(1/0.25+1)²=25` lattice bound); z4 → 9; z3 → 4. Superset, passthrough, sub-cell-invariance, and warmer/client param-set agreement are all unit-proven. Snapping is OUTWARD (superset), so MapLibre clips off-screen data and there is no visual change; displayed "in view" counts are unaffected (they derive from `filterBucketsByBounds` against the RAW viewport bounds).

## Screenshots

Driven live via Playwright MCP at `?scope=us` (CONUS aggregated). Map renders, edge/coastal observations appear, legend + lede counts unchanged from `main`, snapped canonical bbox confirmed on the wire (`-125.00,24.00,-66.00,50.00&zoom=3`). **Zero console errors** at every viewport. (The `[adaptive-grid] getClusterLeaves rejected` warnings are pre-existing — proven byte-identical on `main` at the same view — and unrelated to this change, which only alters the `?bbox=` query value.)

**390×844 — iPhone 14 Pro (mobile)**

| Light | Dark |
|---|---|
|  |  |

**768×1024 — iPad portrait (tablet)**

| Light | Dark |
|---|---|
|  |  |

**1024×768 — iPad landscape / small laptop**

| Light | Dark |
|---|---|
|  |  |

**1440×900 — Desktop standard**

| Light | Dark |
|---|---|
|  |  |

**1920×1080 — Wide desktop**

| Light | Dark |
|---|---|
|  |  |

## Test plan

- [x] `npm test` (all workspaces) — green (geo 15, frontend 1097, ingestor 224, read-api + db-client/shared-types green)
- [x] New unit tests added: geo cardinality-collapse (z3/z4/z5) + sub-cell invariance + superset + passthrough; client snap/passthrough; warmer snapped-URL + warmer/client param-set agreement
- [x] Updated the three contract assertions that encoded the old shape (`run-cache-warm.test.ts` CONUS + metro exact-URLs; `total === 77` stays 77)
- [x] `npm run build` — clean production build (full root chain)
- [x] `npx knip` — exit 0 (geo exports all consumed)
- [x] Built the ingestor Docker image and ran the warmer inside it — confirms `@bird-watch/geo` source-only resolves at runtime (node:24 strips-and-runs the `.ts`); correct snapped URLs, count=77
- [x] Playwright MCP smoke — `npm run dev` against the live API, drove `?scope=us` at all 5 canonical viewports × 2 themes. Map renders, edge observations appear, legend/lede counts unchanged, the snapped canonical bbox (`-125.00,24.00,-66.00,50.00&zoom=3`) is confirmed on the wire. **Zero console errors.** The `[adaptive-grid] getClusterLeaves rejected` warnings are pre-existing (proven byte-identical on `main` at the same view) and unrelated to this change.

## Plan reference

Out of plan — implements issue #866 (agent-ready, bot-approved). Closes #866.

---

### DIFF
--- a/frontend/e2e/state-scope.spec.ts
+++ b/frontend/e2e/state-scope.spec.ts
@@ -1,5 +1,6 @@
 import { test, expect, STATES_FIXTURE } from './fixtures.js';
 import { AppPage } from './pages/app-page.js';
+import { snapFetchBbox, type Bbox } from '@bird-watch/geo';
 
 /**
  * C9 (#741) — chooser-first scope IA: chooser landing, state-select, `?scope=us`
@@ -118,24 +119,36 @@ test.describe('Scope chooser + state/whole-US scope (C9, #741)', () => {
   }
 
   /**
-   * #847 — does `bbox` tightly match the scope `envelope` (each edge within
-   * `tol` degrees)? This is the DETERMINISTIC, WebGL-independent differentiator
+   * The aggregated seed zoom the #847 reseed writes alongside the scope
+   * envelope (App.tsx `AGGREGATED_SEED_ZOOM`). The reseed fetch is therefore an
+   * aggregated (zoom < 6) request, so #866's fetch-time snapping applies to it.
+   */
+  const AGGREGATED_SEED_ZOOM = 3;
+
+  /**
+   * #847 — does `bbox` match the scope `envelope` (each edge within `tol`
+   * degrees)? This is the DETERMINISTIC, WebGL-independent differentiator
    * between the bug and the fix: in this no-WebGL CI the map never settles a
    * state-tight viewport into `debouncedBbox`, so on UNPATCHED `main` the
    * post-switch fetch carries the cold-mount CONUS seed (`-125,24,-66,50`),
-   * which `intersects` (but does NOT tightly match) the new state's envelope.
+   * which `intersects` (but does NOT match) the new state's envelope.
    * The render-phase reseed sets `debouncedBbox` to the new scope's exact
    * envelope, so AFTER the fix the post-switch bbox matches within tol — RED on
-   * main (CONUS seed), GREEN after fix (tight envelope). The bug's live recovery
-   * path (a real map idle refining the bbox) is absent here, so the FETCH
-   * payload — not the eventual lede — is the falsifiable signal.
+   * main (CONUS seed), GREEN after fix.
+   *
+   * #866 — `client.ts` now snaps the FETCH bbox OUTWARD to the shared grid at
+   * zoom < 6, so the reseed fetch carries `snapFetchBbox(envelope, seedZoom)`,
+   * NOT the raw envelope. Compare against the SNAPPED envelope so the assertion
+   * tracks the real on-the-wire contract. (The CONUS seed snapped at z3 is still
+   * `-125,24,-66,50` — integer-aligned — so the RED-on-main property holds.)
    */
   function bboxMatchesEnvelope(
     bbox: [number, number, number, number],
     envelope: [number, number, number, number],
     tol = 0.5,
   ): boolean {
-    return bbox.every((v, i) => Math.abs(v - envelope[i]) <= tol);
+    const snapped = snapFetchBbox(envelope as Bbox, AGGREGATED_SEED_ZOOM);
+    return bbox.every((v, i) => Math.abs(v - snapped[i]) <= tol);
   }
 
   /** Pull the bbox off the LAST /api/observations request carrying `state`. */
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -12,6 +12,7 @@
     "test:e2e": "playwright test"
   },
   "dependencies": {
+    "@bird-watch/geo": "*",
     "@bird-watch/shared-types": "*",
     "@microsoft/clarity": "^1.0.2",
     "@turf/buffer": "^7.3.5",
--- a/frontend/src/api/client.test.ts
+++ b/frontend/src/api/client.test.ts
@@ -25,11 +25,48 @@ describe('ApiClient', () => {
     const envelope = JSON.stringify({ data: [], meta: { freshestObservationAt: null } });
     vi.spyOn(global, 'fetch').mockResolvedValue(new Response(envelope, { status: 200 }));
     const client = new ApiClient({ baseUrl: '' });
+    // No zoom ⇒ per-observation mode ⇒ no snapping, but the canonical
+    // serializer still emits .toFixed(2) values (#866). The integer-aligned
+    // CONUS default serializes to its 2-decimal form.
     await client.getObservations({ bbox: [-125, 24, -66, 50] });
     const call = (fetch as unknown as { mock: { calls: [string, unknown][] } }).mock.calls[0]!;
     const url = call[0];
     // URLSearchParams encodes "," as "%2C"; either form is acceptable.
-    expect(url).toMatch(/bbox=-125(?:%2C|,)24(?:%2C|,)-66(?:%2C|,)50/);
+    expect(url).toMatch(/bbox=-125\.00(?:%2C|,)24\.00(?:%2C|,)-66\.00(?:%2C|,)50\.00/);
+  });
+
+  it('snaps the fetch bbox to the shared cache grid at zoom < 6 (#866)', async () => {
+    // A jittered z5 metro viewport snaps OUTWARD to the 0.25° grid and
+    // serializes via the canonical .toFixed(2) form — the cache-key lever.
+    // Snapping happens at FETCH time (not App.tsx state) so it covers both the
+    // #847 scope-reseed path and the idle path. Displayed counts stay correct
+    // because they derive from filterBucketsByBounds(buckets, viewportBounds)
+    // against the RAW map bounds, not this snapped fetch bbox.
+    const envelope = JSON.stringify({ data: [], meta: { freshestObservationAt: null } });
+    vi.spyOn(global, 'fetch').mockResolvedValue(new Response(envelope, { status: 200 }));
+    const client = new ApiClient({ baseUrl: '' });
+    await client.getObservations({ bbox: [-118.241, 33.998, -107.237, 40.051], zoom: 5 });
+    const call = (fetch as unknown as { mock: { calls: [string, unknown][] } }).mock.calls[0]!;
+    const url = call[0];
+    // raw → snap(z5) → [-118.25, 33.75, -107.00, 40.25] → .toFixed(2)
+    expect(url).toMatch(
+      /bbox=-118\.25(?:%2C|,)33\.75(?:%2C|,)-107\.00(?:%2C|,)40\.25/,
+    );
+  });
+
+  it('passes the bbox through unchanged at zoom >= 6 (per-observation mode, #866)', async () => {
+    const envelope = JSON.stringify({
+      mode: 'observations', data: [], meta: { freshestObservationAt: null },
+    });
+    vi.spyOn(global, 'fetch').mockResolvedValue(new Response(envelope, { status: 200 }));
+    const client = new ApiClient({ baseUrl: '' });
+    await client.getObservations({ bbox: [-118.241, 33.998, -107.237, 40.051], zoom: 7 });
+    const call = (fetch as unknown as { mock: { calls: [string, unknown][] } }).mock.calls[0]!;
+    const url = call[0];
+    // Passthrough: only the canonical serializer applies, no grid snap.
+    expect(url).toMatch(
+      /bbox=-118\.24(?:%2C|,)34\.00(?:%2C|,)-107\.24(?:%2C|,)40\.05/,
+    );
   });
 
   it('maps stateCode to ?state= on /api/observations (#735)', async () => {
--- a/frontend/src/api/client.ts
+++ b/frontend/src/api/client.ts
@@ -2,6 +2,7 @@ import type {
   Hotspot, Observation, ObservationsResponse, SpeciesMeta, ObservationFilters,
   FamilySilhouette, StateSummary, SpeciesDictEntry,
 } from '@bird-watch/shared-types';
+import { snapFetchBboxParam, serializeBbox } from '@bird-watch/geo';
 
 export interface ApiClientOptions {
   baseUrl?: string;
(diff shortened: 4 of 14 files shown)


julianken:
@Mergifyio queue

PR #876, opened 2026-06-05
perf(read-api,frontend): collapse state-scope cache key + dedup in-flight obs fetches (#873 #874)

julianken:
## Diagrams

Two composing levers on the state-scope `/api/observations` path. **#873** collapses the cache key (and tightens the origin scan) by sending the state's *fixed* envelope instead of the center-varying canonical viewport box, in the aggregated path only. **#874** dedups/cancels in-flight fetches at the single client boundary.

```mermaid
flowchart TB
    subgraph A["#873 — state-scope cache key (zoom < 6 ONLY)"]
        pan["3 different CA pans<br/>(viewport centers differ)"] -->|"BEFORE: center-varying<br/>canonicalFetchBbox(viewport)"| miss["3 distinct CONUS-centered keys<br/>→ 100% CF MISS<br/>→ 12–14s CONUS scan-then-clip"]
        pan -->|"AFTER: snapFetchBbox(StateSummary.bbox)"| hit["ONE key per (state, zoom, filters)<br/>-124.50,32.50,-114.00,42.25<br/>→ CF HIT 2nd load + warmable<br/>→ state-tight origin query"]
    end

    subgraph B["#874 — in-flight dedup (client.ts getObservations)"]
        rapid["rapid scope/pan<br/>CA→NV→TX"] -->|"new fetch supersedes"| abort["AbortController.abort() prior<br/>(use-bird-data ignores AbortError)"]
        dup["concurrent byte-identical key"] -->|"coalesce"| shared["return shared promise<br/>(ONE network call)"]
    end
```

```mermaid
flowchart LR
    states["/api/states<br/>StateSummary.bbox"] --> appmemo["App.tsx scopeStateBbox<br/>(useMemo on states + scope)"]
    appmemo -->|"ObservationFilters.stateBbox"| hook["useBirdData<br/>(read via stateBboxRef at fetch-time,<br/>NOT a trigger dep → no extra fetch)"]
    hook --> client["client.ts getObservations<br/>zoom<6 + state + stateBbox<br/>→ snapFetchBboxParam(stateBbox, zoom)"]
    client -->|"?bbox=<fixed snapped envelope>&state=US-XX"| edge["CF edge: one key per (state, zoom)"]
```

## Summary

[...]

PR #871, opened 2026-06-05
perf: tune cache-warming freshness — s-maxage 1800→2400 + retire the legacy CONUS seed

julianken:
## Diagrams

Two independent levers on the `/api/observations` cache path, both surfaced by the #869 prod validation. Lever A widens the edge TTL so a warmed object survives the full warm cycle; Lever B seeds the frontend's first request on the *same* canonical key the warmer warms.

```mermaid
flowchart TB
    subgraph A["Lever A — read-api TTL (cadence boundary)"]
        warm["cache-warmer<br/>every 30 min"] -->|writes object| edge["CF edge object"]
        edge -->|"s-maxage=1800 == cadence<br/>→ EXPIRED at cycle end"| old["late-cycle cold load:<br/>stale-serve (MISS-ish)"]
        edge -->|"s-maxage=2400 (40 min)<br/>~10-min margin"| new["late-cycle cold load:<br/>clean HIT ✅"]
    end

    subgraph B["Lever B — frontend seed (canonical key)"]
        legacy["legacy DEFAULT_BBOX_CONUS<br/>[-125,24,-66,50]"] -->|"canonicalFetchBbox(_,3)"| cold["-129,20,-65,52<br/>(cold key → MISS)"]
        seed["INITIAL_BBOX_SEED = CONUS_BOUNDS<br/>[-130,20,-65,52] (from @bird-watch/geo)"] -->|"canonicalFetchBbox(_,3)"| warmkey["-130,20,-65,52<br/>(warmed key → HIT) ✅"]
    end
```

## Summary

- **Why (Lever A):** `s-maxage=1800` exactly equals the 30-min warm cadence, so a warmed object hits its TTL boundary right as the cycle ends — a real cold-load late in a cycle lands on a just-expired object and stale-serves (`EXPIRED`) instead of a clean HIT (what the #869 desktop probe caught on the *correct* `-130` key: a freshness-timing artifact, not a key mismatch). `2400s (40 min) > 1800s` gives a ~10-min margin so the object stays fresh through the whole cycle until the next warm refreshes it. Edge-staleness ceiling rises ~30→~40 min (0.2% of the 14-day window) — accepted per the #868/#870 staleness decision.
[...]

PR #877, opened 2026-06-05
fix(frontend): clear leaked markers + stale header count on scope transition (#872, #875)

### LINKED ISSUE #872: fix(frontend): clear leaked DOM species-markers + stale header count on scope transition
> Revision 2 — corrected against `julianken-bot`'s review (verified vs `origin/main`): fixed line anchors, retargeted the header-count fix, and demoted the dangerous dep-array variant.

## Problem
On a **state→state** scope transition, the *previous* state's species-icon DOM markers stay rendered ~0.3–1.2s after the camera arrives (drifting **outside the new state's outline**), then all markers vanish ~1s (blank gap), then the new markers paint ~3s in. The header sightings-count also shows the **previous** state's number until the new data lands. Reproduced 4/4 (CA→NV, NV→TX, TX→AZ, AZ→NY) on prod 2026-06-05.

**The backend is innocent** — every `/api/observations?...&state=US-XX` response was correctly polygon-clipped to in-state buckets (GCP read-api logs: "no cross-state leakage at origin"). The leak is purely client-side DOM.

## Root cause
The adaptive-grid reconciler effect at **`frontend/src/components/map/MapCanvas.tsx:2136`** has dep array `[silhouettes, silhouettesById, silhouettesVersion, mapReady, aggregated, buckets, dictionary]` — **`observations` is absent**. The DOM markers render from `groups.map(...)` (~`MapCanvas.tsx:2693`, keyed on geometry-derived `g.key`) and are set **only** inside the async reconciler (`setGroups`@2029, `setSilhouetteOffsets`@2058), which commits on the **next map `idle`** after its per-cluster `getClusterLeaves` chain resolves. The `<Source>` (~2611-2614) updates synchronously via react-map-gl `setData`, but the DOM markers lag until that async pass — so old-scope markers stay mounted. (Note: `frontend/src/data/use-bird-data.ts:145-167` keeps the *previous* scope's `observations` until the new fetch resolves, so during the ~3s gap **both** source and markers show stale data; the marker/source asymmetry only manifests at data-arrival — the blank gap is partly the fetch latency itself, not solely reconciler lag.)

The stale **header count** is the same lag: the visible header is `App.tsx` `ledeText` useMemo (**`App.tsx:986-1036`**), whose existing loading guard at **`App.tsx:1003`** only fires when `observationCount === 0` — so it misses the state→state case where the *stale prior* count is nonzero.

## Fix sketch
**Primary (safe):** invalidate DOM-marker state the instant `observations` changes — either (1) an `observations`-keyed effect that synchronously `setGroups([])` / `setSilhouetteOffsets(new Map())` / resets `prevHiddenSubIdsRef` before the new reconcile, or (2) add a **stable obs-version counter** (NOT raw `observations`) to the reconciler dep array. **Do NOT add raw `observations` to the dep array** — unlike `buckets`/`dictionary` (which have stable empty-identity defaults `EMPTY_BUCKETS`/`EMPTY_DICT`, `MapCanvas.tsx:641-645`, whose comment warns a fresh `[]` per render "spins an infinite re-register loop"), `observations` is a fresh array on every `useBirdData` resolution → reopens that loop. Keep the existing `cacheGeneration` race-guard. **Header:** retarget to `App.tsx` `ledeText` (986-1036) — add a placeholder branch keyed on `observationsLoading` alone, so it shows while loading even when the stale `observationCount` is nonzero.

## Acceptance criteria
- State→state transition: **zero** previous-state DOM markers remain after the new `observations` is set — assert in the existing `MapCanvas.test.tsx` fakeMap harness (it already asserts on `[data-testid="adaptive-grid-marker"]` counts, ~791-828; a "rerender with new observations → old-scope markers gone" assertion drops straight in).
- Header placeholder renders while `observationsLoading` is true **even when the stale `observationCount` is nonzero** (the case `App.tsx:1003` currently misses).
- chrome-devtools repro of CA→NV→TX shows no out-of-outline markers post-data-set; zero console errors.

## Severity
**P1/P2 boundary** — most user-visible correctness defect (markers sit outside the selected state on every state switch), but self-heals in ~3s, app not yet publicized → schedule, not hotfix.

## References
chrome-devtools QA + GCP log correlation, 2026-06-05. Source `setData` behavior verified vs @vis.gl/react-maplibre docs.

### LINKED ISSUE #875: chore(frontend): silence getClusterLeaves 'No cluster with the specified id' flood after setData re-index
> Revision 2 — corrected against `julianken-bot`'s review (verified vs `origin/main`): fixed line refs, corrected the popover mechanism, struck the non-working fix option, and pinned the match semantics.

## Problem
On every aggregated-data load at the state-overview zoom (~z5), the console emits **15–41** `[adaptive-grid] getClusterLeaves rejected agg:1:5:XXXX:YY [object Error]` warnings. The `[object Error]` hides `No cluster with the specified id: NNNN`. **User impact is cosmetic** (P3) — but a dirty console is a Tier-1 finding under the repo's UI-review gate, so it will trip future frontend-PR reviews.

## Root cause
The reconciler reads `cluster_id`s **fresh each pass** from `map.queryRenderedFeatures(undefined, {layers:['clusters-hit']})` (~`MapCanvas.tsx:1783`); when `setData` (new `observations`) re-clusters in the worker, the idle-tick reconciler can race that re-cluster and pass `cluster_id`s from the **prior** supercluster generation into `getClusterLeaves` (~call site `MapCanvas.tsx:1855`), which rejects. The existing `warnedRejections` de-dupe (~`MapCanvas.tsx:1921`) only collapses repeats of the *same* key, so a re-index minting many fresh ids produces a burst. (Verified live: a current id resolved, a logged stale id `5642` rejected with "No cluster with the specified id".)

**Why popovers still work (corrected):** popovers call the **same** `getClusterLeaves(id,64,0)` — there is **no** embedded-data fallback. They work because a click lands *after* the map idles and the index settles, so the rendered `cluster_id`s match the live index at click time — a different timing regime from the idle-tick reconciler racing the worker re-cluster. (If `getClusterLeaves` *did* reject at click time, the click hits `catch {}` and silently no-ops.)

## Fix sketch
**(a)** Swallow the expected post-reindex rejection: match it with `String(err?.message).includes('No cluster with the specified id')` (**NOT `===`** — the supercluster message has a trailing period and an appended id: `"No cluster with the specified id: NNNN"`), and swallow it **regardless of render mode** — the message is the benign discriminator, and QA confirmed the flood fires at BOTH `z<6` aggregated (`agg:…`) AND `z≥6` per-observation (`obs:…`). **Keep warning on any OTHER rejection message** so a genuinely broken cluster still surfaces. (The original agg-only scoping was too narrow — the per-observation flood would otherwise persist.) **(b)** Or validate the id against `map.querySourceFeatures('observations')` before calling `getClusterLeaves`. **Strike the earlier option (c)** ("clear the leaf cache on `observations` change") — the rejecting ids are freshly queried from `queryRenderedFeatures`, not pulled from `leafCache`, so clearing the cache does nothing; and the reconciler dep array (~`MapCanvas.tsx:2136`) deliberately omits `observations`, so there's no "invalidate on observations change" hook today.

## Acceptance criteria
- A state-overview load (fresh and post-transition) produces **zero** `getClusterLeaves rejected` console warnings — assert with a deterministic test extending `MapCanvas.test.tsx` (harness exists: "reconciler swallows getClusterLeaves rejections" ~L688, warn-once assertion ~L755): a `No cluster with the specified id: 5642` rejection yields **zero** `console.warn`, while a different rejection (`'boom'`) **still warns once**. Add cases for BOTH a `agg:` (aggregated) and an `obs:` (per-observation) stale-id rejection → both yield 0 warns; an unrelated rejection still warns once.
- Species popovers still open with correct content at z5 and z9 (no regression in the click path).

## Severity
**P3** (console hygiene; popovers functional). Largely resolves alongside the scope-transition reconciler fix (#872).

## References
chrome-devtools QA 2026-06-05. maplibre-gl ^5.24.0 clustering API verified (no drift). Note: `dictGen` is a dictionary-resolution flag (0/1), not a supercluster generation counter.

julianken:
## Diagrams

```mermaid
sequenceDiagram
    participant User
    participant App as App.tsx (ledeText)
    participant MC as MapCanvas
    participant Hook as useBirdData
    participant Src as GeoJSON Source
    participant Rec as adaptive-grid reconciler

    User->>Hook: pick new state (CA → NV)
    Hook->>Hook: observationsLoading = true<br/>(KEEPS prior observations until resolve)
    Hook-->>App: observationsLoading=true, count still NONZERO
    Note over App: BEFORE: :1003 guard only fires at count===0<br/>→ leaks stale "4080 sightings"
    App->>App: #872 FIX: observationsLoading branch<br/>→ "Updating…" (count-free)
    Hook-->>MC: observations = new array identity
    MC->>MC: #872 FIX: render-phase clear<br/>setGroups([]) / setSilhouetteOffsets(new Map())
    Note over MC: BEFORE: async markers lag → leak<br/>outside new outline ~0.3–1.2s
    MC->>Src: setData (synchronous re-cluster)
    Rec->>Rec: idle tick: stale cluster_id → getClusterLeaves rejects
    Note over Rec: #875 FIX (BOTH modes): swallow<br/>"No cluster with the specified id: NNNN"
    Rec-->>MC: new-scope markers paint
```

## Summary

- **#872** — On a state→state scope change, the reconciler effect deliberately omits `observations` from its dep array (the `EMPTY_BUCKETS`/`EMPTY_DICT` infinite-re-register hazard), so its async DOM markers lag the synchronously-updated GeoJSON `<Source>` and the prior state's markers leak outside the new outline. Fix: a synchronous `observations`-keyed clear (render-phase store-previous-prop/compare-in-render, mirroring the existing `silhouettesVersion` ref) drops `groups`/`silhouetteOffsets`/`prevHiddenSubIdsRef` the instant `observations` identity changes — so stale markers never paint. Raw `observations` stays out of the reconciler deps; the `cacheGeneration` race-guard is untouched.
- **#872 (header)** — The `App.tsx` `ledeText` cold-load guard only fired at `observationCount===0`, missing the state→state case where the carried-over prior count is nonzero. Added an `observationsLoading`-keyed `"Updating…"` placeholder branch so any in-flight refetch shows count-free copy, never the stale number.
- **#875** — After `setData` re-indexes the supercluster, the idle reconciler can pass prior-generation `cluster_id`s into `getClusterLeaves`, which rejects with `"No cluster with the specified id: NNNN"` (15–41 warnings per state-overview load). Swallow that exact message (`includes`, not `===` — trailing period + appended id) **in BOTH render modes** — QA confirmed the same benign flood fires at `z<6` aggregated (`agg:…`) AND `z≥6` per-observation (`obs:…`). The message is the benign discriminator, so the swallow keys off it alone (no render-mode gate); any OTHER rejection message still warns once via `warnedRejections`. Cache eviction stays unconditional. Closes #872 and closes #875.

> **Update (rebased onto #876 + widened #875):** rebased onto `main` after PR-A (#876) merged — clean, App.tsx regions disjoint (PR-A's `stateBbox`/`useBirdData` filter threading vs this PR's `ledeText` placeholder). Per the revised #875, the swallow is now widened from aggregated-only to **both render modes** (the per-observation `obs:` flood the earlier "Known concern" flagged is now resolved). New test asserts an `obs:` stale-id rejection yields 0 warns alongside the existing `agg:` case; an unrelated `boom` rejection still warns once.

> **Update (review fix — `boundsKey` gating):** the render-phase marker-clear now keys on **`boundsKey`** (the canonical scope-change signal, derived from `state.scope` in `App.tsx`) rather than `observations` array identity. A `julianken-bot` review caught that the observations-identity gate fired on *every* fetch — so a same-scope pan/zoom at `z≥6` (which refetches) would blank the adaptive-grid markers (~0.3–1.2s flicker) on a common interaction. `boundsKey` changes only on real scope transitions and is stable under pan, confining the clear to the case #872 targets. New flicker-guard test asserts a fresh `observations` array under an *unchanged* `boundsKey` does NOT clear markers; the #872 scope-transition test was updated to drive a real `boundsKey` change. `prevObservationsRef` removed.

## Screenshots

Driven live via Playwright MCP against `npm run dev` (proxied to the live API) at the 5 canonical viewports × 2 themes. The fix changes **transient** behavior + console hygiene only (loading placeholder + no marker leak during the refetch window + suppressed getClusterLeaves flood); the settled map is visually identical with or without the fix — there is no pixel diff, mirroring PR #871. The load-bearing evidence is the live transition + console verification below, not static frames.

**Live verification (this fix):**
- **State transition (CA → NV → TX), desktop 1440×900:** header shows **`Updating…`** (count-free) for the full in-flight window — NOT the stale prior-state count — then resolves to the new state's count; no stale out-of-outline markers persist after the new data sets (#872 confirmed). Sampled `Updating…` continuously from t≈120 ms through t≈2.5 s.
- **Zoom to z≥6 (per-observation mode) + pan/zoom churn, CA:** API served `zoom=6,7,8,9` requests (per-obs regime); console stayed **clean of any `[adaptive-grid] getClusterLeaves rejected` warning** — the #875 per-obs flood is gone (the exact regression this PR widens the swallow to cover).
- **Whole-session console (Playwright MCP, `all=true`):** **zero `getClusterLeaves rejected` warnings** across CA load + CA→NV→TX + z3→z9 zoom. The only residual warning is one benign maplibre-gl worker message (`Expected value to be of type number, but found null`, a `blob:` tile-expression warning — pre-existing, library-level, unrelated). A transient `503 database unavailable` from the live Texas API request also surfaced (infra blip, not this change).

**Screenshots (paste-upload) status — N/A / pending reviewer ruling.** The 5×2 viewport×theme captures were driven in the Playwright-MCP sandbox, which is not reachable by the chrome-devtools paste-upload browser used by the `pr-screenshots-via-user-attachments` skill (and that browser cannot reach the local dev server) — the known serial-ship limitation. Because the settled state has **no visual diff** (transient + console-only fix, like PR #871), the reviewer is asked to rule N/A on the `user-attachments` requirement; the live-driven evidence above is the substantive verification.

- [x] Every new/renamed `className` in this PR has a matching CSS rule — `N/A — no className changes` (orphan-classname-check: PASS)
- [ ] Multi-viewport design QA: ≥10 `user-attachments` URLs — **N/A pending reviewer ruling** (no visual diff; transient/console-only fix; paste-upload blocked by sandbox isolation — see above)

## Test plan

- [x] `npm run -ws test` — green (frontend: 68 files / 1114 tests, incl. both #875 cases: `agg:` swallow + new `obs:` swallow; #872 placeholder + marker-clear; all other workspaces green)
- [x] New unit tests added — TDD red→green: the new per-observation `#875` test fails against the prior agg-only `aggregated && …` condition (got 1 warn, expected 0) and passes with the widened message-only swallow
- [ ] New Playwright e2e spec — not added; the bug is a sub-2 s transient timing window + console flood that the unit harness pins deterministically (fakeMap rerender + warn-spy); an e2e would be flaky against real fetch latency
- [x] `npm run build` — clean (`tsc -b` typecheck + vite build, exit 0)
- [x] `npx knip` — exit 0; `npm run lint` — exit 0 (incl. forbidden-token guard)
- [x] (UI) Playwright MCP live — CA→NV→TX transition (`Updating…` placeholder, no stale markers) + zoom to z≥6 per-obs mode (console clean of getClusterLeaves flood); see Screenshots

## Plan reference

Out of plan — bug-fix PR for bot-approved, agent-ready issues #872 and #875 (post-ship correctness/console-hygiene). Rebased onto #876 (PR-A) after it merged.

---

### DIFF
--- a/frontend/src/App.test.tsx
+++ b/frontend/src/App.test.tsx
@@ -2354,6 +2354,7 @@ describe('O8 (#784): React.memo render-count regression — FamilyLegend + Scope
 describe('#828: lede dedupe — count-only copy, no region, no time-window', () => {
   const STATES = [
     { stateCode: 'US-AZ', name: 'Arizona', bbox: [-114.82, 31.33, -109.05, 37.0] as [number, number, number, number] },
+    { stateCode: 'US-CA', name: 'California', bbox: [-124.41, 32.53, -114.13, 42.01] as [number, number, number, number] },
   ];
 
   function obs(overrides: Partial<{ speciesCode: string; comName: string; familyCode: string | null }> = {}) {
@@ -2539,5 +2540,54 @@ describe('#828: lede dedupe — count-only copy, no region, no time-window', () 
     const lede = await screen.findByTestId('map-lede');
     expect(lede).not.toHaveTextContent(/in the last/i);
     expect(lede).not.toHaveTextContent(/14 days/i);
+  });
+
+  // #872 — state→state stale-count guard. On a state→state scope change,
+  // use-bird-data keeps the PRIOR state's `observations` mounted while the new
+  // fetch is in flight (it only clears on resolve), so `observationCount` is
+  // NONZERO during the load. The cold-load guard (`:1003`) only fires at
+  // `observationCount === 0`, so it misses this case and the lede would show
+  // the stale prior count. The fix: a placeholder branch keyed on
+  // `observationsLoading` alone — the lede must NOT show the stale number while
+  // a refetch is in flight, even when the carried-over count is nonzero.
+  it('#872: shows a loading placeholder (not the stale count) during a state→state refetch', async () => {
+    mockUrlState.state = {
+      since: '14d', notable: false, speciesCode: null, familyCode: null,
+      view: 'map', scope: { kind: 'state', stateCode: 'US-AZ' },
+    };
+    // First state resolves a NONZERO species count.
+    mockGetObservations.mockResolvedValue({
+      data: [
+        obs({ speciesCode: 'vermfly' }),
+        obs({ speciesCode: 'gilwoo', comName: 'Gila Woodpecker' }),
+      ],
+      meta: { freshestObservationAt: new Date().toISOString() },
+    });
+    const { rerender } = render(<App />);
+    const lede = await screen.findByTestId('map-lede');
+    expect(lede).toHaveTextContent('2 species');
+
+    // Transition to a new state whose observations fetch never resolves —
+    // observationsLoading flips true while the prior (nonzero) observations are
+    // still mounted. This is the exact state→state window #872 describes.
+    mockGetObservations.mockReturnValue(new Promise(() => {}));
+    await act(async () => {
+      mockUrlState.state = {
+        since: '14d', notable: false, speciesCode: null, familyCode: null,
+        view: 'map', scope: { kind: 'state', stateCode: 'US-CA' },
+      };
+      rerender(<App />);
+    });
+
+    // The lede must NOT show the stale prior count while the new fetch loads.
+    await waitFor(() => {
+      const ledeNow = screen.queryByTestId('map-lede');
+      // Either the row is gone OR it shows a count-free placeholder — never the
+      // stale "2 species".
+      if (ledeNow) {
+        expect(ledeNow).not.toHaveTextContent(/\d+\s+species/i);
+        expect(ledeNow).not.toHaveTextContent(/\d+\s+sightings/i);
+      }
+    });
   });
 });
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -106,6 +106,16 @@ export const INITIAL_ZOOM_SEED = 3;
 const AGGREGATED_SEED_ZOOM = 3;
 
 /**
+ * #872: count-free lede placeholder shown while a refetch is in flight. On a
+ * state→state transition use-bird-data keeps the prior scope's nonzero
+ * `observations` mounted until the new fetch resolves, so the lede must show
+ * this neutral "updating" copy rather than the stale count. Count-free by
+ * design — no number can be trusted mid-refetch. The cold-load case (count 0)
+ * still returns null upstream so the row stays hidden until the first paint.
+ */
+const LEDE_LOADING_PLACEHOLDER = 'Updating…';
+
+/**
  * O2 (#770): Default-expanded breakpoint for FamilyLegend, moved here from
  * MapSurface so the legend can render as a persistent App-root sibling.
  *
@@ -1018,6 +1028,15 @@ export function App() {
     // Cold-load guard (#716/#720): suppress Template 1 while the first fetch is
     // in flight. Same discipline as MapLede's `loading` guard.
     if (observationsLoading && observationCount === 0 && speciesCount === 0) return null;
+    // #872: state→state stale-count guard. use-bird-data keeps the PRIOR scope's
+    // `observations` mounted while the new fetch is in flight (it clears only on
+    // resolve), so during a state→state transition `observationCount` is NONZERO
+    // and the cold-load guard above (which only fires at count === 0) misses it —
+    // leaking the previous state's number until the new data lands. Branch on
+    // `observationsLoading` alone so any in-flight refetch shows a count-free
+    // placeholder rather than the stale number. (The cold-load case still
+    // returns null above so the row stays hidden until the first data arrives.)
+    if (observationsLoading) return LEDE_LOADING_PLACEHOLDER;
     // #828: the lede is count-only. The region moved into the wordmark headline
     // (no longer repeated in the sentence) and the time-window dropped entirely
     // (it's discoverable via Filters). No period clause, no `${region}` — see the
--- a/frontend/src/components/map/MapCanvas.test.tsx
+++ b/frontend/src/components/map/MapCanvas.test.tsx
@@ -885,10 +885,12 @@ describe('MapCanvas', () => {
 
   // #875 — after `setData` re-indexes the supercluster, the idle-tick reconciler
   // can pass cluster_ids from the PRIOR generation into getClusterLeaves, which
-  // rejects with "No cluster with the specified id: NNNN". In aggregated mode
-  // that expected post-reindex rejection must be swallowed (zero console.warn);
-  // any OTHER rejection must still warn once via the existing warnedRejections
-  // path. Rendered in `mode='aggregated'` because the swallow is agg-only.
+  // rejects with "No cluster with the specified id: NNNN". That expected
+  // post-reindex rejection must be swallowed (zero console.warn) in BOTH render
+  // modes — QA confirmed the flood fires at z<6 aggregated (`agg:…`) AND z≥6
+  // per-observation (`obs:…`). Any OTHER rejection must still warn once via the
+  // existing warnedRejections path. This first case pins the AGGREGATED path
+  // (cache key prefix `agg:`); the per-observation case follows below.
   it('#875: swallows "No cluster with the specified id" rejection in aggregated mode, still warns on others', async () => {
     const DICT = new Map<string, { comName: string; familyCode: string }>([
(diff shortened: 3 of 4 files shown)


julianken:
@Mergifyio queue

Issue #873, opened 2026-06-05
fix(read-api,frontend): canonicalize state-scope bbox cache key (100% CF MISS + 12–14s near-timeout CONUS scans)


julianken:
> Revision 2 — corrected against `julianken-bot`'s review (verified vs `origin/main@f23a265`): fixed the root-cause mechanism (center-varying canonical bbox, not raw-float), scoped the fix to the aggregated path, and recalibrated severity.

## Problem
Every **state-scoped** `/api/observations` request is a Cloudflare **MISS** while national `?scope=us` is a HIT (in a 14-min QA window: 17 state-scope origin MISSes vs 0 national origin hits). The state-scope origin queries took **12–14s** five separate times (CA z5 12.6s; TX z5 13.6/12.9/12.4/13.6s) vs **0.14–0.23s** for state-tight (zoom≥6) queries — sitting near the Cloud Run / `statement_timeout` (~15s) cliff, masked by the green 200s.

## Root cause (corrected)
`canonicalFetchBboxParam` (`frontend/src/api/client.ts:48-51`) is applied to **every aggregated request (`zoom<6`) with no `stateCode` branch** — so state bboxes **are** canonicalized (#869 did not skip them). They MISS because the canonical box is **center-varying**: it's reconstructed from the viewport center (which differs per state and per pan) clamped to `CONUS_BOUNDS`, so each state/pan mints a **new** cache key (e.g. CA z5 transmits `-130.00,25.25,-91.25,49.75`; NV z5 `-130.00,26.25,-94.75,50.75`). Since CF keys the full query string (`cache-rule.tf`, `ignore_query_strings_order=true`), each distinct bbox is a distinct key → ~0% reuse. **Proven:** TX z5 `-122.25,20,-77.75,43.50` MISSed cold, then the **byte-identical** query HIT (304) later, while one-digit-different `...,43.75` MISSed as a separate key.

Crucially, for a state scope the `ST_Intersects` polygon clip (`packages/db-client/...observations.ts:374-383`) already fully bounds the result — so the wide bbox is **redundant for scoping** and serves only to fragment the cache key.

## Fix sketch
[...]

PR #903, opened 2026-06-06
perf(read-api,ingestor): precompute per-scope aggregation grid to kill 12–15s cold state fetch (#878)

### LINKED ISSUE #878: perf(read-api): high-volume state-scope aggregation is 12–14s at origin regardless of bbox (cold-fill near-timeout)
## Problem
State-scope `/api/observations` aggregation for **high-volume states is 12–14s at origin, independent of bbox tightness**. Prod measurement (read-api `bird-read-api-00094-q4s`, 2026-06-05, after #876/#873 shipped):

- `state=US-CA zoom=5 bbox=-124.50,32.50,-114.00,42.25` (**CA-tight** envelope) → **12.52s / 12.70s**
- `state=US-TX zoom=5 bbox=-106.75,25.75,-93.50,36.50` (**TX-tight** envelope) → **13.62s** (live re-measured 2026-06-05: **15.1s** cold)
- `state=US-AZ` tight bboxes → **0.13–0.22s** (moderate volume, fine)

So the 12–14s is **not** the wide-CONUS-bbox scan that #873 assumed — it's the cost of aggregating *all* of a high-volume state's observations once the `ST_Intersects` polygon clip pulls them in. #873's bbox-collapse correctly fixed the **cache** (state scopes are now cacheable; CA/TX cold-fill pays once per ~40-min `s-maxage` cycle, then HITs) and the loading/dedup issues — but the **cold/first request on CA/TX still pays 12–14s**, one bad scan from the ~15s `statement_timeout`. The near-timeout alert added in #873 (`monitoring.tf`) covers exactly this residual.

## Root cause — CONFIRMED (source + comparative, 2026-06-05 investigation)
`getObservationsAggregated` (`packages/db-client/src/observations.ts:335`) is a 5-CTE pipeline. The `base` CTE (`:430-439`) materializes **every** observation in scope:
```
base: SELECT round(ST_X/ST_Y * mult)/mult AS lng/lat_bucket, species_code, family_code
      FROM observations o LEFT JOIN species_meta sm …  WHERE <state ST_Intersects clip> AND obs_dt >= since
```
then `per_species` (GROUP BY bucket,family,species) → `ranked` (ROW_NUMBER window) → `per_family` (`jsonb_agg` top-N species) → `families` (per-bucket rollup) + `bucket_totals` aggregate that full row set at request time.

**The cost is the aggregation over the in-scope row set, not row-finding.** Row-finding is already indexed: `obs_geom_idx` (GIST on `geom`, migration `1700000006000`) backs the `ST_Intersects` spatial clip and `obs_dt_idx` backs the `since` filter. Proof it is aggregation-bound, not index-bound: **AZ and CA/TX run the identical query with the identical indexes** — AZ 0.13–0.22s vs CA/TX 12–15s. The only variable is the number of rows inside the state polygon, so the cost scales with row count = the per-request aggregation. (Corroborated by the `:416-428` #862 history: national aggregation once hit 147s on a jsonb-in-sort-key blowup, tuned to ~2.3s by splitting `bucket_totals`, but the fundamental "aggregate every in-scope row per request" remains.)

*Empirical confirmation (implementer's RED step):* `EXPLAIN (ANALYZE, BUFFERS)` on a cold CA/TX state scope — expect a cheap GIST Bitmap Index Scan feeding the `base` materialization, with the runtime concentrated in the HashAggregate / WindowAgg / Sort nodes over `base`. (The design below is already settled by the comparative evidence; the EXPLAIN is confirmation + the perf-guard baseline.)

## Approach — DECIDED: precompute (index rejected)
**An index cannot meet the AC.** It can only speed *finding* the rows (already fast — AZ proves the scan path is fine); it cannot reduce the cost of *aggregating* N rows. A BRIN/covering index or grid-precomputed column does not change the GROUP BY / window / `jsonb_agg` cost. Rejected.

**Precompute the aggregated grid per scope at ingest time** so the request is a cheap lookup. **Correction to the original issue text:** there is NO existing precompute to "mirror #860" — verified zero materialized views/tables in `migrations/`; #859/#860's "compute-on-write" is itself the request-time CTE pipeline above. This issue introduces the *first* materialized aggregation.

Design sketch (implementer to refine + justify with tests):
- **New table** e.g. `observation_grid_agg(scope_key TEXT, grid_multiplier INT, lng_bucket DOUBLE PRECISION, lat_bucket DOUBLE PRECISION, observation_count INT, species_count INT, families JSONB, PRIMARY KEY(scope_key, grid_multiplier, lng_bucket, lat_bucket))`. `scope_key` = state code (`US-CA`, …) plus a national key (`US`). Plain-SQL migration under `migrations/` (`-- Up/-- Down`), per repo convention.
- **Covers the slow default view:** unfiltered (`since`=14d, no `notable`/`speciesCode`/`familyCode`), at the 3 validated grid tiers (multiplier 2/4/8 — the read-api's closed zoom→multiplier switch lives in `services/read-api/src/app.ts:242`; the `observations.ts:392-395` comment only describes it). ~50 states + national × 3 tiers ≈ 150 grids — bounded and tiny.
- **Populate at ingest time:** after each `/recent` ingest + reconcile **and after the 14-day prune**, recompute the grids (run the existing aggregation per scope/tier, or a single grouped pass; consider denormalizing a `state_code` onto `observations` at ingest so the per-state grouping is cheap — implementer's call). One ingest cycle of staleness (~30 min) is acceptable — the edge cache already runs ~40 min.
- **Read path** (`getObservationsAggregated` / read-api): define the lookup predicate **positively** — use the precompute when `scope_key` resolves (a state code, or national) **AND** zoom<6 **AND** `since`=default(14d) **AND** no `notable`/`speciesCode`/`familyCode` **AND** a standard grid multiplier. **IGNORE the `bbox` in this decision.** A scoped state view ALWAYS sends the deterministic snapped state-envelope bbox (`frontend/src/api/client.ts:85-92`), so "`f.bbox` is present" must NOT route to the fallback — doing so would send the exact default view this fix targets back to the 12–15s live path and defeat the whole change. Fall back to the live CTE query only when a filter is set, `since` is non-default, or there is no resolvable `scope_key` (a free-pan bbox with no state/national scope) — those are less common and the filters reduce the row count anyway.

Must NOT:
- Serve stale cells across ingest/prune — the refresh must run after BOTH the ingest+reconcile and the prune.
- Change results for filtered/non-default requests — those keep the exact live path.
- Block the request path on the precompute — populate is an ingestor-side job, never request-time.

Non-goals (v1): precomputing filtered variants (`notable`-only, per-family) — live fallback is fine; extend later only if a filtered case proves slow.

## Why the perf guard missed it
`packages/db-client/src/observations-aggregated-perf.test.ts`'s state-scope guard passed at ~5855ms because its **550k-row testcontainers seed under-represents prod CA/TX observation volume**. The guard's seed/threshold need to reflect the heaviest real state.

## Acceptance criteria
- `EXPLAIN (ANALYZE, BUFFERS)` on a cold high-volume state scope captured (prod or a realistically-seeded testcontainers) confirming the runtime is in the aggregate nodes over `base` (the RED baseline).
- New precompute table + ingest-time populate + read-path lookup, with the live CTE retained as the fallback for filtered/non-default requests.
- Cold **default** state-scope for the highest-volume state (CA/TX) lands well under `statement_timeout` — target sub-second (comparable to AZ), proven by a perf-guard test **re-seeded to realistic high-volume**; the guard fails on `main` at the current cost and passes after the fix (RED→GREEN discriminator).
- Filtered/non-default requests verified byte-identical to the live path (no regression).
- Correctness across an ingest cycle AND the 14-day prune (a test that the grid refreshes after both).
- `test` / `lint` / `build` / `e2e` / `knip` green; plain-SQL `-- Up/-- Down` migration.

## Severity
**P2** — latent reliability (cold CA/TX near-timeout) now *mitigated* by #873's edge cache (HIT after first MISS) but not eliminated; the cold-fill is still 12–14s and one bad scan from the timeout. App not yet publicized.

## References
Surfaced by the #876 caching-track prod validation (2026-06-05); root cause + approach settled by the 2026-06-05 source investigation (this issue). Follow-up to #873 (fixed the cache/loading, not the origin aggregation cost). The #873 `monitoring.tf` near-timeout alert covers this residual. Query: `packages/db-client/src/observations.ts:335-500`; indexes: migration `1700000006000_observations.sql`.

julianken:
## Diagrams

```mermaid
sequenceDiagram
    participant Browser
    participant ReadAPI
    participant Postgres
    participant Ingestor

    Note over Ingestor,Postgres: off the request path (every cycle)
    Ingestor->>Postgres: recent ingest + reconcile (upsert)
    Ingestor->>Postgres: refreshGridAgg() — rebuild observation_grid_agg
    Ingestor->>Postgres: prune (14d delete)
    Ingestor->>Postgres: refreshGridAgg() — rebuild (drop aged-out cells)

    Note over Browser,Postgres: default state view (eligible)
    Browser->>ReadAPI: GET /api/observations?state=US-TX&bbox=<envelope>&zoom=5&since=14d
    ReadAPI->>ReadAPI: isPrecomputeEligible? (scope + default since + no filters + std mult)
    ReadAPI->>Postgres: PK lookup observation_grid_agg (US-TX, mult 8)  ~19ms
    Postgres-->>ReadAPI: precomputed buckets (byte-identical to live CTE)
    ReadAPI-->>Browser: mode=aggregated

    Note over Browser,Postgres: filtered / non-default (fallback)
    Browser->>ReadAPI: ...&notable=true  (or since=7d, or non-std mult)
    ReadAPI->>Postgres: live 5-CTE getObservationsAggregated (unchanged)
```

```mermaid
flowchart LR
    obs[observations<br/>14d window] -->|ST_Intersects join<br/>state_boundaries| scoped[scoped rows<br/>US + US-XX]
    scoped -->|CROSS JOIN mult 2/4/8| base[base buckets]
    base --> pipeline[per_species → ranked →<br/>per_family → families<br/>+ bucket_totals]
    pipeline --> grid[(observation_grid_agg<br/>scope_key, grid_multiplier,<br/>lng/lat_bucket, counts, families)]
```

## Summary

- **Why:** the state-scope low-zoom aggregation aggregates *every* in-state observation at request time — `EXPLAIN (ANALYZE, BUFFERS)` on a realistically-seeded high-volume Texas scope shows **14,119ms over 240,980 in-TX rows**, with the runtime in the `Sort`/`GroupAggregate`/`WindowAgg` nodes over `base`, NOT the GIST row-find (which returns 240k rows in ~948ms). This is the prod 12–15s cold CA/TX fill, one bad scan from the 15s `statement_timeout`. An index can't help — it speeds *finding* rows, not *aggregating* them (confirmed by AZ vs CA/TX running the identical query+indexes).
- **What:** precompute the aggregated grid per scope (national + every CONUS state) × the 3 standard grid multipliers (2/4/8) into a new `observation_grid_agg` table at ingest time, and serve the default low-zoom view as a cheap PK lookup. The populate shares the **exact** live CTE pipeline (extended with a `scope_key` key), so a precomputed cell is byte-identical to the live query for the matching `{ since:'14d', stateCode? }` request. The read-path predicate is **positive** (scope resolves + default `since` + no `notable`/`species`/`family` + standard multiplier); the snapped state-envelope `bbox` is **ignored** in the decision (it's a function of the scope, co-extensive with the polygon clip — routing on "bbox present" would defeat the fix). Everything else falls through to the live CTE, unchanged.

## Screenshots

N/A — not UI. Backend-only change (migration + db-client + read-api routing + ingestor hook); the frontend is untouched and the response shape/contents are byte-identical to the live path.

## Test plan

- [x] `npm run test` (affected workspaces) — green: db-client 145 pass, read-api 174 pass, ingestor 231 pass
- [x] New unit / integration tests added (real Postgres+PostGIS testcontainers, no mocks):
  - `observation-grid-agg.test.ts` — **equivalence** (precompute byte-identical to live CTE for national + per-state at every multiplier), **positive predicate** (`isPrecomputeEligible`), **refresh correctness across an ingest delta AND a prune** (no stale cells)
  - `app.test.ts` (#878 routing) — default state scope routes to the **lookup** (sub-second) proven via a stale-cache discriminator; a `notable` filter and a non-default `since` route to the live **fallback**
  - `cli.test.ts` (#878) — refresh fires after a successful `recent` AND `prune`, fires on `partial`, **skipped** on a failed run and on other kinds, refresh failure is non-fatal
  - `observations-aggregated-perf.test.ts` — re-seeded to single-high-volume-state (TX) scale; the **RED→GREEN discriminator**: live TX aggregate **9244ms** vs precompute lookup **19ms** (state-scope guard lookup ~55ms), byte-identical (258 buckets)
- [x] `npm run build` — clean production build (all workspaces)
- [x] `npx knip` — clean (only the pre-existing unrelated config hint on the old prototype dir)
- [x] EXPLAIN (ANALYZE, BUFFERS) RED baseline captured (cold high-volume TX, live path):
  ```
  Execution Time: 14119.128 ms   (240,980 in-TX rows, 14d window)
  CTE base → Index Scan obs_geom_idx: actual 1.760..455.895 rows=240980   (row-find is cheap)
  GroupAggregate ← Sort (external merge, Disk: 8040kB): 1313..1392ms       (cost is the aggregate)
  WindowAgg / GroupAggregate / Incremental Sort: loops=258 over CTE base
  ```
  After the fix the same default TX view is a PK lookup: **~19ms** (live aggregate **9244ms** at the in-process testcontainer measurement → lookup **19ms**).

## Plan reference

Out of plan — fixes #878 (P2 latent reliability: cold high-volume state aggregation 12–15s, near `statement_timeout`). Follow-up to #873 (fixed the cache/loading, not the origin aggregation cost).

---

### DIFF
--- a/migrations/1700000051000_observation_grid_agg.sql
+++ b/migrations/1700000051000_observation_grid_agg.sql
@@ -1,0 +1,43 @@
+-- Up Migration
+--
+-- observation_grid_agg — precomputed per-scope aggregation grid (#878).
+--
+-- The state-scope low-zoom `/api/observations` aggregation (getObservationsAggregated,
+-- packages/db-client/src/observations.ts) aggregates EVERY in-state observation at
+-- request time. For high-volume states (CA/TX, ~12-15s cold) the cost is the
+-- HashAggregate/WindowAgg/Sort over the in-scope row set, NOT row-finding (AZ runs
+-- the identical query + indexes in 0.13-0.22s; only row count differs). An index
+-- cannot help — it speeds finding rows, not aggregating them. So we precompute the
+-- aggregated grid per scope at ingest time and serve the default low-zoom view as a
+-- cheap PK lookup.
+--
+-- One row per (scope_key, grid_multiplier, lng_bucket, lat_bucket):
+--   scope_key       — a state code ('US-CA', …) OR the national key 'US'.
+--   grid_multiplier — the read-api zoom→grid switch (2/4/8; app.ts:242).
+--   lng_bucket /    — round(coord * grid_multiplier) / grid_multiplier, IDENTICAL
+--   lat_bucket        to the live `base` CTE bucket key (observations.ts:432-433).
+--   observation_count / species_count — the live `bucket_totals` (over ALL rows in
+--                     the cell, incl. NULL-family rows).
+--   families        — the live `families` per-bucket jsonb rollup, byte-identical to
+--                     the live CTE: families ordered (count desc, code asc); species
+--                     nested top-8 (count desc, code asc); NULL-family rows excluded
+--                     from families[] but counted in the totals.
+--
+-- This is the FIRST materialized aggregation in the schema (verified: zero
+-- materialized views / agg tables pre-#878). The populate (refreshGridAgg in
+-- db-client/observations.ts) runs ingest-side after each /recent ingest+reconcile
+-- AND after the 14-day prune — never on the request path. One ingest cycle of
+-- staleness (~30 min) is acceptable (the edge cache already runs ~40 min).
+CREATE TABLE observation_grid_agg (
+  scope_key         TEXT             NOT NULL,
+  grid_multiplier   INTEGER          NOT NULL,
+  lng_bucket        DOUBLE PRECISION NOT NULL,
+  lat_bucket        DOUBLE PRECISION NOT NULL,
+  observation_count INTEGER          NOT NULL,
+  species_count     INTEGER          NOT NULL,
+  families          JSONB            NOT NULL,
+  PRIMARY KEY (scope_key, grid_multiplier, lng_bucket, lat_bucket)
+);
+
+-- Down Migration
+DROP TABLE IF EXISTS observation_grid_agg;
--- a/packages/db-client/src/index.ts
+++ b/packages/db-client/src/index.ts
@@ -4,6 +4,9 @@ export { getHotspots, upsertHotspots, type HotspotInput } from './hotspots.js';
 export {
   getObservations, getObservationsAggregated, upsertObservations,
   runReconcileStamping, getFreshestObservationAt,
+  // #878 — precomputed per-scope aggregation grid.
+  refreshGridAgg, getAggregatedGridFromCache, isPrecomputeEligible,
+  resolveScopeKey, NATIONAL_SCOPE_KEY, STANDARD_GRID_MULTIPLIERS,
   type ObservationInput,
 } from './observations.js';
 export {
--- a/packages/db-client/src/observation-grid-agg.test.ts
+++ b/packages/db-client/src/observation-grid-agg.test.ts
@@ -1,0 +1,291 @@
+/**
+ * #878 — precomputed per-scope aggregation grid (observation_grid_agg).
+ *
+ * Real Postgres+PostGIS testcontainer (no DB mocks, per repo convention). The
+ * seed is small but geometrically representative: rows inside two state
+ * polygons (US-AZ, US-CA) plus a few outside both, so the per-state clip and the
+ * national rollup are both exercised. The serial-plan GUC (max_parallel_workers
+ * _per_gather = 0) makes the float-keyed bucket aggregation run-to-run
+ * deterministic so the byte-identity assertion against the live CTE is exact.
+ *
+ * Guards:
+ *   1. EQUIVALENCE — refreshGridAgg's grid is byte-identical to the live
+ *      getObservationsAggregated CTE for the default unfiltered scope (national
+ *      AND a state), at every standard multiplier (2/4/8): same buckets, counts,
+ *      species counts, and families jsonb.
+ *   2. PREDICATE — isPrecomputeEligible is positive: default state/national view
+ *      (with the always-present state-envelope bbox) is eligible; any filter,
+ *      non-default since, or non-standard multiplier is not.
+ *   3. REFRESH CORRECTNESS — after an ingest delta AND after a prune the grid
+ *      reflects the new state with no stale cells.
+ */
+import { describe, it, expect, beforeAll, afterAll } from 'vitest';
+import { PostgreSqlContainer, type StartedPostgreSqlContainer } from '@testcontainers/postgresql';
+import { readFileSync, readdirSync } from 'node:fs';
+import { join, resolve } from 'node:path';
+import pg from 'pg';
+// Side-effect import: registers pool-wide type parsers (NUMERIC → number).
+import './pool.js';
+import {
+  getObservationsAggregated,
+  refreshGridAgg,
+  getAggregatedGridFromCache,
+  isPrecomputeEligible,
+  resolveScopeKey,
+  NATIONAL_SCOPE_KEY,
+  STANDARD_GRID_MULTIPLIERS,
+} from './observations.js';
+import type { AggregatedBucket } from '@bird-watch/shared-types';
+
+let container: StartedPostgreSqlContainer;
+let pool: pg.Pool;
+
+// A handful of (lng, lat) points inside US-AZ and US-CA, plus a couple outside
+// both (a Gulf-of-Mexico point and a mid-Atlantic point) so the national grid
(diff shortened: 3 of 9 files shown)


julianken:
@Mergifyio queue

PR #934, opened 2026-06-09
refactor(frontend): move hashSubId into deconflict.ts (U4, #888)

julianken:
## Diagrams

```mermaid
graph LR
    subgraph before["Before"]
        MC1["MapCanvas.tsx<br/>function hashSubId (local)"]
        DT1["deconflict.test.ts<br/>function hashForTest (mirror dup)"]
        MC1 -. "-hashSubId(subId)" .-> CID1["DeconflictInput.cluster_id<br/>(negative pseudo-id)"]
    end
    subgraph after["After"]
        DC2["deconflict.ts<br/>export function hashSubId (appended)"]
        MC2["MapCanvas.tsx<br/>import { hashSubId }"]
        DT2["deconflict.test.ts<br/>import { hashSubId } + direct tests"]
        DC2 --> MC2
        DC2 --> DT2
        MC2 -. "-hashSubId(subId)<br/>(unchanged call site)" .-> CID2["DeconflictInput.cluster_id<br/>(negative pseudo-id)"]
    end
    before ==> after
```

## Summary

- **Why:** `hashSubId`'s only purpose is the deconflict layer's negative-pseudo-id contract — callers negate its always-positive djb2 result to mint a NEGATIVE `DeconflictInput.cluster_id` so silhouette inputs ride through `buildGroups` alongside real (positive) supercluster ids without collision. That sign convention is enforced in `deconflict.ts`, so the helper belongs next to it, not in the `MapCanvas.tsx` god-component (U4 of the consolidation epic #884, Wave 0).
- **What:** Append-only extraction — `hashSubId` moved to `deconflict.ts` (no edits to existing exports, so U10/#895 inherits a clean append), imported back into `MapCanvas.tsx` (the `cluster_id: -hashSubId(subId)` call site is unchanged → rendered output byte-identical), and the mirrored test-only `hashForTest` duplicate deleted from `deconflict.test.ts`.
- **Net test delta:** +4 — direct `hashSubId` coverage (determinism, positivity, distinctness, djb2-reference match) replaces the deleted duplicate.

## Screenshots

N/A — behavior-preserving extraction, no visible UI change

- [x] Every new/renamed `className` in this PR has a matching CSS rule in
      `frontend/src/styles.css` or `frontend/src/components/ds/ds-primitives.css`
      (or marked `N/A — class is consumed only by a primitive that ships its own CSS`) — N/A, no className changes
- [x] Multi-viewport design QA: PR body has ≥10 `user-attachments` URLs
[...]

PR #933, opened 2026-06-09
refactor(frontend): extract MapCanvas type declarations → MapCanvas.types.ts (#885)

julianken:
## Diagrams

```mermaid
graph LR
    subgraph before["MapCanvas.tsx (before)"]
        A1["type ResolvedAdaptiveData"]
        A2["interface MapCanvasProps (export)"]
        A3["interface SelectedObsState"]
        A4["runtime: hooks, JSX,<br/>reconciler, callbacks"]
    end

    subgraph after["after — same render, types relocated"]
        T["MapCanvas.types.ts<br/>ResolvedAdaptiveData · MapCanvasProps · SelectedObsState"]
        M["MapCanvas.tsx<br/>runtime only + import type back<br/>+ re-export MapCanvasProps"]
        M -- "import type" --> T
        M -. "export type { MapCanvasProps }" .-> T
    end

    before -->|"U1 extraction (#885)"| after
```

## Summary

- **Why:** `MapCanvasProps`, `SelectedObsState`, and `ResolvedAdaptiveData` are type-only (erased at build, zero runtime), so co-locating them with the ~2,800-line imperative core only adds noise to the very file the rest of epic #884 exists to shrink. This is the lowest-risk Wave-0 cut and gives later units a stable home for shared prop/state types.
- **What:** Move the three top-level type declarations (with their load-bearing doc-comments — the App.tsx scope/camera/artboard wiring contract #736/#740/#760/#761/#782 and the #718 popover-projection note) into `frontend/src/components/map/MapCanvas.types.ts`. `MapCanvas.tsx` back-imports all three and re-exports `MapCanvasProps`, so rendered output and the module's public surface are byte-identical. Orphaned type imports the move left behind (`FamilySilhouette`, `AdaptiveTile`, `ResolvedGrid`) are pruned from `MapCanvas.tsx` and now resolve in the new file.
- **knip stays clean:** the new file's exports are all consumed by the back-import, and the re-exported `MapCanvasProps` is covered by `ignoreExportsUsedInFile` (its in-file destructuring consumer at `}: MapCanvasProps`).

## Screenshots

N/A — behavior-preserving extraction, no visible UI change. (Type-only `.ts` move; Wave 0 is no-Playwright per the epic plan.)

- [x] Every new/renamed `className` in this PR has a matching CSS rule — `N/A — type-only .ts file, no className changes`
- [x] Multi-viewport design QA — `N/A — not UI`

## Test plan

[...]

PR #927, opened 2026-06-07
feat(db-client): server drift-proof family names via COALESCE in both aggregated CTEs

julianken:
## Diagrams

```mermaid
flowchart LR
    subgraph read["getObservationsAggregated (live CTE)"]
        O1[observations] --> B1[base<br/>+ family_name<br/>+ fs.common_name]
        SM1[species_meta] -.LEFT JOIN.-> B1
        FS1[family_silhouettes<br/>UNIQUE family_code] -.LEFT JOIN.-> B1
        B1 --> PF1[per_family<br/>min&#40;common_name&#41;<br/>min&#40;family_name&#41;]
        PF1 --> J1["families jsonb_build_object<br/>'name', COALESCE&#40;common_name, family_name&#41;"]
    end
    subgraph precompute["refreshGridAgg (precompute CTE)"]
        O2[observations] --> R2[recent<br/>+ family_name<br/>+ fs.common_name]
        SM2[species_meta] -.LEFT JOIN.-> R2
        FS2[family_silhouettes<br/>UNIQUE family_code] -.LEFT JOIN.-> R2
        R2 --> PF2[per_family<br/>min&#40;common_name&#41;<br/>min&#40;family_name&#41;]
        PF2 --> J2["families jsonb_build_object<br/>'name', COALESCE&#40;common_name, family_name&#41;"]
    end
    J1 -. "byte-identical literal<br/>(#878 guard)" .- J2
    J2 --> AGG[(observation_grid_agg.families<br/>JSONB — no DDL)]
```

## Summary

- **Why:** PR1/PR2 deliver the entire *visible* win from `/api/silhouettes`. This PR upgrades the *invisible drift case* — a brand-new family observed before its `family_silhouettes` migration row exists has nothing in `/api/silhouettes`, but eBird `species_meta.family_name` does. Projecting `name` server-side names it anyway. The stale-cache window is user-invisible (the frontend already resolved the name from the silhouette layer), so **no Cloudflare `/api/observations` purge tooling is needed** — that is the de-risking the caching audit bought.
[...]

PR #862, opened 2026-06-03
fix(map): anchor portaled CellPopover to its cell via getBoundingClientRect (#863)

julianken:
## Diagrams

```mermaid
flowchart TD
    click["User clicks a family silhouette cell<br/>(aggregated / low zoom)"] --> mount["CellPopover mounts<br/>anchorEl = clicked cell button"]
    mount --> portal["createPortal to document.body<br/>(#859 — escapes the MapLibre marker's<br/>transform stacking context)"]

    subgraph bug["BEFORE (#863 bug)"]
        portal --> noPos["No computed position;<br/>relies on CSS position: absolute"]
        noPos --> origin["Portaled to body → no positioned ancestor,<br/>no left/top → collapses to body origin"]
        origin --> corner["Card renders in the<br/>BOTTOM-LEFT corner ❌"]
    end

    subgraph fix["AFTER (#863 fix)"]
        portal --> measure["useLayoutEffect:<br/>anchorEl.getBoundingClientRect()"]
        measure --> compute["computePopoverPosition()<br/>below + left-aligned to cell"]
        compute --> edge{"Off-screen?"}
        edge -->|overflows bottom| flip["flip above the cell"]
        edge -->|overflows right/left| clamp["clamp into viewport"]
        edge -->|fits| place["place below"]
        flip --> inline["inline position: fixed; left; top<br/>(wins over CSS position: absolute)"]
        clamp --> inline
        place --> inline
        inline --> anchored["Card anchors next to<br/>the clicked cell ✅"]
    end
```

## Summary

- **Why:** The per-family `CellPopover` was rendering in the **bottom-left corner** instead of next to the clicked cell. Root cause: #859 portaled the card to `document.body` (the z-index/stacking-context fix so cluster pills can't paint over it), but the card never computed a position — it relied on `.cell-popover`'s CSS `position: absolute`. Portaled to `document.body` there is no positioned ancestor and no `left`/`top`, so `absolute` collapses to the body origin.
- **Fix:** Mirror the existing `CellHoverPreview` pattern — compute placement from `anchorEl.getBoundingClientRect()` in a `useLayoutEffect` (pre-paint, no flash) and apply an inline `position: fixed; left; top` that wins over the CSS `absolute`. Anchors below + left-aligned to the cell, with edge handling (flip above on bottom overflow, clamp into the viewport on any edge). The `createPortal(..., document.body)` stays (removing it would reintroduce pills-over-popover).
[...]

PR #936, opened 2026-06-09
test(db-client): recalibrate national perf guard 8000→11000ms (CI-runner-noise, query unchanged)

julianken:
## Diagrams

N/A — test-only threshold recalibration (one constant + its rationale comment). No control flow, data flow, or structural change to diagram. For orientation, the guard's place in the system:

```mermaid
flowchart LR
    req[National low-zoom request<br/>gridMult=2, since=14d] -->|eligible| grid[(precompute grid<br/>#903/#878 ~19ms PK lookup)]
    req -->|ineligible fallback| cte[getObservationsAggregated CTE<br/>guarded by this test]
    cte -->|must finish under| to[pool statement_timeout 15000ms<br/>guard trips first at 11000ms]
```

## Summary

- **Why:** the national `getObservationsAggregated(gridMultiplier=2, since=14d)` perf guard in `packages/db-client/src/observations-aggregated-perf.test.ts` has flaked red on CI with **no query change** (#933 = 8144ms on main `f3b8e8e9`; #934 = 8093ms). A 4-agent investigation concluded this is **CI-runner noise, not a query regression** — so the threshold, not the query, was wrong.
- **The miscalibration:** the "~2.3s baseline" was only a design-target *comment* in the #862 commit, never a logged CI number. The real CI-measured floor (best-of-3 min on the testcontainer) has been **~5.3–7.1s** for the whole recorded window, with slow-runner mins reaching **~7.6–8.0s**. `8000ms` sat only **~1.1–1.3× over the true floor** — far too tight for one-sided runner contention.
- **The change:** raise `PERF_THRESHOLD_MS` `8000 → 11000` and rewrite the rationale comment to record measured reality. This is a **recalibration, not a weakening**: 11000ms keeps a hard guard **4s under** the 15000ms pool `statement_timeout` (>1 full best-of-3 sample of slack), clears the slow-runner mins with margin, and still catches the pre-#862 ~147s plan by **~13×**. Every regression class that matters crosses 15s; 11s trips first.

### Evidence the query is unchanged (so this is not masking a regression)

- `git log df65336..HEAD -- packages/db-client/src/observations.ts` is **empty** — the query body is **byte-identical since #927**.
- #927's `family_silhouettes` LEFT JOIN is **inert in this test**: the seed uses synthetic `fam-000..fam-039` codes; `family_silhouettes` holds real eBird codes, so the JOIN matches **0 rows**. The CTE was already **~6.5s before #927** landed.
- **Zero prod impact:** prod serves the national view from the precompute grid (#903/#878, ~19ms PK lookup). This CTE is only the ineligible-request fallback. The guard's real job is protecting the 15000ms pool `statement_timeout` (catching a return of the pre-#862 ~147s plan).
- `PERF_RUNS` / `fastestOf` best-of-3 de-noising is **unchanged** — only the threshold constant and its comment move.

## Screenshots

N/A — not UI.

## Test plan

- [x] `git diff` confirms the only change is `PERF_THRESHOLD_MS` `8_000 → 11_000` plus its rationale comment; `observations.ts` and every query are untouched (`git log df65336..HEAD -- packages/db-client/src/observations.ts` empty).
- [x] `npm test --workspace @bird-watch/db-client` — run locally against real Postgres+PostGIS via testcontainers; the national perf guard passes at the 11000ms threshold. CI re-validates the real timing on its runners.
- [ ] New unit / integration tests added (if behavior changed) — N/A, this *is* the test; behavior of the system under test is unchanged.
- [ ] New Playwright e2e spec added (if user-visible behavior changed) — N/A, no user-visible change.
- [ ] `npm run build` — N/A for a comment+constant change in a test file; no build-surface impact.
- [ ] (UI only) Playwright MCP smoke — N/A — not UI.

## Plan reference

Out of plan — flaky-perf-guard recalibration following the 4-agent investigation into #933/#934 (CI-runner noise, query byte-identical since #927).

---

### DIFF
--- a/packages/db-client/src/observations-aggregated-perf.test.ts
+++ b/packages/db-client/src/observations-aggregated-perf.test.ts
@@ -55,25 +55,47 @@ import {
 const TARGET_ROWS = Number(process.env.PERF_ROWS ?? 550_000);
 
 // Hard ceiling: the read-api pool's statement_timeout is 15_000ms (pool.ts).
-// We assert well under it — the #862 fix runs ~2.3s at prod scale on CI-class
-// hardware, but containers + shared CI runners are noisy, so 8_000ms is the
-// guard threshold (comfortably below 15s, comfortably above the ~2.3s fix, and
-// FAR below the pre-#862 ~147s). A run over this means the timeout regression
-// is back.
+// We assert well under it. The "~2.3s" figure in this file's history was a
+// design-target COMMENT from the #862 commit, NOT a logged CI number — the
+// actual CI-measured floor of the best-of-3 min on this testcontainer has been
+// ~5.3–7.1s for the entire recorded window, with slow-runner mins reaching
+// ~7.6–8.0s. The old 8_000ms threshold sat only ~1.1–1.3× over that true floor:
+// far too tight for one-sided runner noise, so it flaked red on CI without any
+// query change (#933 = 8144ms on main f3b8e8e9; #934 = 8093ms). A 4-agent
+// investigation confirmed this is CI-runner noise, NOT a query regression: the
+// query body in observations.ts is BYTE-IDENTICAL since #927
+// (`git log df65336..HEAD -- packages/db-client/src/observations.ts` is empty),
+// and the CTE was already ~6.5s BEFORE #927 landed (#927's family_silhouettes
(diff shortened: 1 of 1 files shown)


julianken:
@Mergifyio queue

Issue #884, opened 2026-06-05
Epic: Consolidate MapCanvas.tsx — extract the ~900 lines of accretion around the irreducible imperative core


julianken:
## Why this epic exists

An external skill-assessment review flagged `frontend/src/components/map/MapCanvas.tsx` as a "2,446-line god-component" with "risk-inverted test allocation." A 4-phase adversarial investigation (2026-06-04; full report in-repo at `docs/analyses/2026-06-04-mapcanvas-decomposition-investigation/report.md`) reached a **split verdict**:

- **God-component → PARTIAL.** "Deferral, not inability" is **supported** (~140 unit tests on already-extracted sibling modules in the same folder; the #760/#762/#763 blank-map logic was extracted 2026-05-29 in #765/#766). But "2,446-line god-component" is **overstated**: ~44% is comments/blank, and ~60% of the code is genuinely irreducible MapLibre orchestration that must own the live `mapRef`. **The real work is consolidation of ~900 lines of extractable accretion around a thick-but-load-bearing imperative core — not a break-up or rewrite.**
- **Risk-inverted tests → REFUTED (both lenses).** Reading test *bodies* (not line counts), the deepest unit coverage already sits on the highest-blast surfaces; the only sibling-test-less surfaces (`use-states`, `use-theme`, `use-breakpoint`) are low-blast/cosmetic. The one residual the investigation surfaced — the Central-US ingest national-row-cap gap (#841/#844/#846) — was in fact **already fixed** by the per-state fan-out (PR #841, closing issue #840, merged 2026-06-02; `services/ingestor/src/handler.ts` no longer pins `regionCode:'US'`). Only a coverage-invariant **guard test** remains, tracked separately and **out of scope here**. (The source report's "open gap" framing was already stale when written.)

The sharpest true finding: this deferral was **silent** — no issue tracked it. **Filing this epic is the fix for that** — it converts silent deferral into tracked, scheduled consolidation.

## Goal / end-state

Shrink `MapCanvas.tsx` to its **true irreducible core** — `handleLoad`, the reconciler shell, the `<MapView>/<Source>/<Layer>` JSX, the `useStateArtboard`/`useScopeCamera` hooks, and the click-routing callbacks (`handleGroupClick`, `openClusterListFromGroup`) — all anchored on the un-movable `mapRef`. Then re-characterize the file's header JSDoc as that core and **stop**. Do not chase further extraction that would invert control and re-open the state-switch bug class.

[...]

Issue #459, opened 2026-05-11
fix(a11y): aria-pressed on ThemeToggle + cluster-pill target size + nested-interactive markers


julianken:
## Context

Three independent WCAG accessibility fixes surfaced by the 2026-05-11 audit. Each is self-contained (no shared code paths); all can ship in a single PR of ~20 LoC.

### Fix 1 — `aria-pressed` on ThemeToggle (WCAG 4.1.2)

`frontend/src/components/ThemeToggle.tsx` renders a `<button>` that toggles between light and dark themes but does not set `aria-pressed`. Screen readers therefore cannot announce the button's current state to the user — a WCAG 4.1.2 Name, Role, Value violation.

**Fix:** add `aria-pressed={theme === 'dark'}` to the ThemeToggle button AND change the `aria-label` to a static value (`"Switch theme"` or equivalent). Keeping an action-describing label (`"Switch to dark theme"`) alongside `aria-pressed` is a known antipattern — AT may announce "Switch to dark theme, toggle button, pressed" — incoherent. The correct pattern: static label + `aria-pressed={theme === 'dark'}` (preferred) or keep action-describing label swap + `role="switch" aria-checked` (alternative). Do NOT keep the action-describing label swap and add `aria-pressed` simultaneously.

### Fix 2 — cluster-pill touch target size (WCAG 2.5.8)

`.cluster-pill` measures approximately 35–37 × 21px on desktop viewports. WCAG 2.5.8 (Level AA, WCAG 2.2) requires a minimum target size of 24 × 24px. The pill falls below the height threshold.

**Fix:** add `min-height: 24px` to the `.cluster-pill` CSS rule in `ds-primitives.css` (or `styles.css`).

### Fix 3 — nested interactive elements in MapLibre markers (WCAG 4.1.2)

MapLibre marker DOM wraps a child `<button>` (or interactive element) inside a parent `<div role="button">` — a nested-interactive pattern that violates WCAG 4.1.2 and causes double-announcement in screen readers. `MapMarkerHitLayer.tsx:135` already renders a separate keyboard-accessible `<button>` per marker; the MapLibre wrapper's `role="button"` is redundant.

**Fix:** remove `role="button"` from the marker wrapper element (or replace with `role="presentation"`). Do NOT use `aria-hidden="true"` on the wrapper — this propagates to children, hiding the inner `<button>` from AT and silently regressing keyboard semantics.

## Evidence

[...]

PR #939, opened 2026-06-09
refactor(frontend): extract PresentationMarker → PresentationMarker.tsx (U5)

### LINKED ISSUE #890: U5 — Extract PresentationMarker → PresentationMarker.tsx
Part of #884 (MapCanvas consolidation epic). **Wave 0b · risk: low · 1 PR.**

**Goal:** Extract the WCAG role-strip Marker wrapper.

**Pulls (by symbol):** `PresentationMarker` + `PresentationMarkerProps` — wraps react-map-gl `<Marker>` and strips maplibre's auto `role=button` after mount (nested-interactive WCAG 4.1.2 fix, #459). Zero coupling to MapCanvas state/refs.

**Target module:** `frontend/src/components/map/PresentationMarker.tsx`

**Idiom:** presentational component taking already-derived props (exemplar `AdaptiveGridMarker.tsx`).

**Depends on:** none.

**Test (same PR — NEW `PresentationMarker.test.tsx`, RTL):**

The post-mount effect is `useEffect(..., [])` that calls `markerRef.current.getElement().setAttribute('role', 'presentation')`, guarded by `if (mk && typeof mk.getElement === 'function')` (`MapCanvas.tsx:269-274`). The existing `react-map-gl/maplibre` Marker mock in `MapCanvas.test.tsx:283-291` is a plain non-ref `<div>` with **no ref and no `getElement()`** — against it, `markerRef.current` stays `null`, the guard short-circuits, and `role="presentation"` is **never set**. Do **not** copy that mock: it can only assert the no-op path, which under-delivers the #459 guard this issue exists to lock in.

Instead, the new spec authors its **own** `forwardRef` Marker mock that exposes a `getElement()` returning a real DOM node via `useImperativeHandle` — for example:

```tsx
import { forwardRef, useImperativeHandle } from 'react';

const markerEl = document.createElement('div'); // per-render real DOM node
vi.mock('react-map-gl/maplibre', () => ({
  Marker: forwardRef(function Marker({ children }: any, ref: any) {
    useImperativeHandle(ref, () => ({ getElement: () => markerEl }), []);
    return <div data-testid="mock-marker">{children}</div>;
  }),
}));
```

Then assert the effect actually ran:

1. **Sets presentation role:** render one `<PresentationMarker>`; after mount, assert `markerEl.getAttribute('role') === 'presentation'` (#459). (The guard requires a real `getElement()` — that is exactly what this mock supplies.)
2. **Re-fires per remount:** render the marker inside a parent `map()` keyed by `g.key`; change the keyed item so React unmounts the old marker and mounts a fresh one (the effect has `[]` deps, so a new mount — not a re-render — is what re-runs it); assert the freshly-mounted marker's element carries `role="presentation"`. Reset the element's role attribute between mounts so the assertion proves the *new* mount set it, not a residue from the prior one.

**Note:** `PresentationMarker` has **no `className`** (verified) — orphan-classname-check does not apply here. Render output is byte-identical (location-only move), so `N/A — not UI` is defensible **provided** the implementer drives one Playwright sanity pass (markers still render, clean console) — no 10-capture set needed.

**Acceptance:** single green PR + 1 Playwright sanity drive.

julianken:
## Diagrams

```mermaid
graph LR
    subgraph Before
        MC1[MapCanvas.tsx]
        MC1 -->|defines + renders| PM1[PresentationMarker<br/>WCAG role-strip wrapper]
    end
    subgraph After
        PM2[PresentationMarker.tsx<br/>presentational module]
        MC2[MapCanvas.tsx] -->|imports| PM2
        MC2 -->|renders ×3 call sites| PM2
        U11[U11 future consumer] -.->|will import| PM2
    end
    Before ==>|location-only move<br/>byte-identical render| After
```

## Summary

- **Why:** `PresentationMarker` is presentational with zero coupling to MapCanvas state/refs — it takes already-derived `longitude`/`latitude`/`anchor` props and only strips maplibre's auto `role="button"` after mount (#459 W4-C nested-interactive WCAG 4.1.2 fix). Extracting it shrinks MapCanvas and lets U11 consume the wrapper directly (epic #884, Wave 0b).
- **What moved:** the 24-line JSDoc + `PresentationMarkerProps` interface + `PresentationMarker` function → new `frontend/src/components/map/PresentationMarker.tsx`. MapCanvas now imports it; the three existing call sites are unchanged. `Marker` dropped from MapCanvas's `react-map-gl/maplibre` import (now used only inside the extracted module).
- **Behavior-preserving:** same Marker import source, same props, same `[]`-dep post-mount effect, same rendered JSX → byte-identical output.

## Screenshots

N/A — not UI. Behavior-preserving extraction (location-only move); no visible UI change. Per the testing protocol, a Playwright sanity drive was still run: the app boots clean, chrome + Map region + scope-picker render, `browser_console_messages` returns **0 warnings**. The only 3 console errors are upstream `502 Bad Gateway` from the dev `/api/*` proxy (no backend reachable in the isolated worktree) — environmental, not from this change; the app degrades gracefully ("Couldn't load the state list" path). No React render error, no `.error-screen`, no uncaught exception.

- [x] Every new/renamed `className` in this PR has a matching CSS rule in
      (or marked `N/A — class is consumed only by a primitive that ships its own CSS`) — N/A, `PresentationMarker` has no `className` (verified); orphan-classname-check does not apply.
- [x] Multi-viewport design QA: PR body has ≥10 `user-attachments` URLs
      (5 viewports × 2 themes per #445). — `N/A — not UI` (behavior-preserving extraction, no visible UI change).

## Test plan

- [x] `npm run typecheck && npm run test` — green (frontend build = `tsc -b && vite build` clean; **1204 tests passed across 74 files**, including the new colocated spec).
- [x] New unit / integration tests added — `PresentationMarker.test.tsx` (RTL). TDD: wrote the spec first, confirmed RED (module absent → import fails), then GREEN. Authors a purpose-built `forwardRef` Marker mock exposing `getElement()` via `useImperativeHandle` (the repo's only existing Marker mock is a plain non-ref `<div>` that can assert only the effect's no-op short-circuit). Asserts (1) `role="presentation"` is set after mount (#459 set-path), and (2) the effect re-fires on each fresh remount (keyed by `g.key`, `[]`-dep effect).
- [x] New Playwright e2e spec added (if user-visible behavior changed) — N/A, no user-visible behavior change.
- [x] `npm run build` — clean production build.
- [x] (UI only) Playwright MCP smoke — ran `npm run dev --workspace @bird-watch/frontend`, drove the app; `browser_console_messages` returns 0 warnings (3 errors are environmental `/api/*` 502s, not from this change). Markers require live observation data unavailable in the sandbox; the move is verified byte-identical via the unit suite + diff review.

## Plan reference

Out of plan — MapCanvas consolidation epic #884, unit U5 (#890), Wave 0b. Tracked in GitHub issues (no `docs/plans/` file; plans live in issues for this epic).

Closes #890
Part of #884

---

### DIFF
--- a/frontend/src/components/map/MapCanvas.tsx
+++ b/frontend/src/components/map/MapCanvas.tsx
@@ -13,7 +13,6 @@ import {
   Map as MapView,
   Source,
   Layer,
-  Marker,
 } from 'react-map-gl/maplibre';
 import type { MapLayerMouseEvent, MapRef } from 'react-map-gl/maplibre';
 import 'maplibre-gl/dist/maplibre-gl.css';
@@ -58,6 +57,7 @@ import { mergeLeafBuckets } from '../../data/bucket-aggregates.js';
 import type { SpeciesDictionary } from '../../data/use-species-dictionary.js';
 import { ObservationPopover } from './ObservationPopover.js';
 import { AdaptiveGridMarker } from './AdaptiveGridMarker.js';
+import { PresentationMarker } from './PresentationMarker.js';
 import { ClusterListPopover } from './ClusterListPopover.js';
 import { ClusterPill } from '../ds/ClusterPill.js';
 import { registerSilhouetteSprite } from './silhouette-sprite.js';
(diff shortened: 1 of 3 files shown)


julianken:
@Mergifyio queue

Issue #953, opened 2026-06-09
epic: transitions.dev catalog motion for the transient layer (2026-06-09 animation audit)


julianken:
## Summary

Apply the transitions.dev catalog to the un-animated **transient layer** of bird-maps.com, per the 2026-06-09 animation audit (5-lens static analysis + live Chrome DevTools driving of prod + 20-agent adversarial verification; report local at `docs/analyses/2026-06-09-transitions-dev-animation-audit/report.md`, untracked — each child issue below is self-contained).

The audit's headline: the app is already highly motion-disciplined — one global `motion.css` reduced-motion policy, no `transition: all`, no FLIP, no motion libs, and the shipped field-guide motion (#08 sheet morph, #07 detail rail, #18 identity stagger, #01 card-resize, #14/#15 shimmers) is catalog-faithful and **untouched by this epic**. Live perf baseline: LCP 153ms / CLS 0.00 entrance, INP 51ms chooser→map, CLS 0.00 + no jank at 4× CPU throttle during re-cluster. This epic adds the missing catalog entrances without disturbing any of that.

## Children (implementation order)

- [ ] #948 — chore: motion hygiene (lede tabular-nums, stale 300ms comments, shared ds-shimmer token) — *zero-risk, lands first*
- [ ] #949 — feat: Photo skeleton→image reveal (#14, image fade-in only) — *standalone, trivial*
- [ ] #950 — feat: transient-popover enter family (#05/#06/#07/#12/#17, 7 surfaces, one PR) — *the core of the epic*
- [ ] #951 — feat: scope disclosure → #07 panel-reveal — *deepest precondition (double display:none lock)*
- [ ] #952 — feat: controls micro-motion (theme #09, badge #03, chevron rotate)

Ship serially per the multi-PR convention (Mergify `batch_size: 1`; rebase dismisses approval).

## Hard constraints (every child)

- transitions.dev 18-recipe catalog only — **no FLIP** (rejected twice), no motion libraries.
- Reduced-motion is owned globally by `frontend/src/styles/motion.css` — no per-element `prefers-reduced-motion` queries; every conditionally-mounted surface **rests at the visible end-state** (the global guard zeroes the interpolation, not the start value — a hidden rest-state means stuck-hidden for reduced-motion users).
- maplibre-gl 5.x owns all camera/cluster/marker motion — no CSS enter/exit on map markers.
- Four-corner floating-card contract intact — these are all transient-layer or in-corner-card motions; no new bands.

[...]

PR #955, opened 2026-06-09
chore: standardized local dev-data seed (npm run db:seed)

julianken:
## Summary

Local dev has **no observation data** out of the box: migrations seed only static reference data (`family_silhouettes`, `state_boundaries`, `species_meta`), while every sighting otherwise comes from a live eBird ingest. A fresh clone therefore renders "No recent sightings" with an empty map/lede/legend unless you have an eBird API key. This adds a deterministic **`npm run db:seed`** that fills that gap.

- `packages/db-client/src/dev-seed.ts` — generates ~400 representative observations across ~24 real families (FK-correct against `family_silhouettes`) and ~18 CONUS states, within the 14-day recency window, ~8% notable; then calls `refreshGridAgg()` so the precompute-backed low-zoom map, lede count, and family legend all populate.
- Built on the real, tested db-client functions (`upsertSpeciesMeta` / `upsertObservations` / `refreshGridAgg`) so it stays correct as schema evolves.
- `package.json`: `db:seed` script. `knip.ts`: registers the seed entry so it isn't flagged as unused.
- `README.md`: local-dev quickstart (`db:up && db:migrate && db:seed`).

## Screenshots

`N/A — not UI` (backend/tooling: a seed script + test + docs).

## Reviewer notes

- [x] Multi-viewport design QA: `N/A — not UI`
- Deterministic: `seedDevData(pool, nowMs)` takes an injected clock; `obs_dt` is computed relative to `now()` so seeded rows stay inside the 14-day window on any run.
- Idempotent: `ON CONFLICT DO NOTHING`, re-runnable.

## Test plan

- `packages/db-client/src/dev-seed.test.ts` (testcontainers, no mocks): after migrate + seed, asserts `observations > 0`, `observation_grid_agg > 0`, and that every observation's `species_code` resolves to a `species_meta` row whose `family_code` exists in `family_silhouettes` (no orphan joins).
- Validated live against a local DB: observations=400, observation_grid_agg=645, 24 families / ~18 states; read-api serves correct aggregated buckets with real family names.

## Linked

Closes the "no local dev seed" gap surfaced during the transitions.dev epic (#953) verification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,10 +38,18 @@ when you're zoomed out, individual family-coloured markers when you're zoomed in
 npm install
 npm run db:up        # local Postgres + PostGIS via Docker
 npm run db:migrate
+npm run db:seed      # ~400 sample sightings so the map isn't empty
 npm run dev --workspace @bird-watch/frontend
 ```
 
 To also run the API the map reads from, see [`services/read-api/README.md`](services/read-api/README.md).
+
+`db:seed` writes a deterministic, idempotent spread of ~400 observations across
+~24 bird families and ~20 states — enough to render a realistic map without an
+eBird API key. (Migrations only seed reference data: family silhouettes, state
+boundaries, and a small species set; real sightings otherwise come from a live
+eBird ingest.) Re-run it any time; it reads `DATABASE_URL` and defaults to the
+local Docker database.
 
 ## Architecture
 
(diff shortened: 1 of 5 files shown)


julianken:
@Mergifyio queue

PR #958, opened 2026-06-09
feat(frontend): scope disclosure open → #07 panel-reveal (remove the double display:none lock)

julianken:
## Diagrams

```mermaid
stateDiagram-v2
    [*] --> Closed: scope active, rows MOUNTED
    Closed --> Open: click 🔍 (data-open=true)
    Open --> Closed: ✕ / Esc (data-open=false)

    state Closed {
        [*] --> offset
        offset: translateY(14px) · opacity 0 · blur 2px<br/>visibility hidden (out of a11y tree + tab order)
    }
    state Open {
        [*] --> rest
        rest: translateY(0) · opacity 1 · blur 0<br/>visibility visible — recipe #07 enter tween
    }

    note right of Open
        REST state = the VISIBLE end-state.
        Only CLOSED is offset, so the GLOBAL
        reduced-motion guard (motion.css zeroes
        the interpolation, not the start value)
        lands the form fully visible on open.
    end note
```

## Summary

- **Why #07, not #01:** the scope disclosure (🔍 → state/ZIP form in the top-left identity card) now opens with transitions.dev recipe **#07 panel-reveal** (transform/opacity/blur, composites cleanly) instead of the rejected **#01 card-resize** (per-frame height reflow, can't tween to `height:auto` without JS measurement — the FLIP-adjacent trap). The identity card is `position:absolute`, so its growth never reflows the map canvas.
- **The double `display:none` lock is gone.** The collapse was locked twice — the React `hidden` prop on `.app-header-scope-rows` AND a CSS `[data-open='false'] { display:none }` rule. Both are removed; either left in force clips the start frame and the transition never fires.
- **Rest-at-visible + reduced-motion + a11y.** The rows REST at the visible open end-state and are only offset while closed, so the global reduced-motion guard (`motion.css`, which zeroes the interpolation not the start value) lands the form fully visible on open. Closed-state uses `visibility: hidden` — not `inert` alone — which keeps the form out of the a11y tree + tab order AND satisfies the e2e `toBeHidden()` asserts (an `inert` element still occupies layout and would fail them). The element stays painted (`visibility`, not `display:none`), so the synchronous `data-open` flip tweens without a post-paint rAF. The open-edge focus move (first `<select>`) is unchanged.

## Screenshots

[...]

PR #842, opened 2026-06-02
fix(frontend): polish scope-card disclosure — soft focus, even rhythm, legend clearance (#828/#837)

julianken:
## Diagrams

```mermaid
flowchart TB
    subgraph TL["Top-left identity card (expanded scope disclosure)"]
        direction TB
        WM["Wordmark / Region"]
        LD["Lede (count only)"]
        DV["hr divider"]
        SEL["state select (first field)"]
        ZIP["ZIP + Go"]
        EXIT["Whole US / Change scope"]
        WM --> LD
        LD --> DV
        DV --> SEL
        SEL --> ZIP
        ZIP --> EXIT
    end
    subgraph BOT["Bottom corners (phone, 480px and below)"]
        direction LR
        LEG["family legend (expanded, 280px)"]
        ATT["map-attribution OpenFreeMap / eBird"]
    end
    F1["Fix 1 soft accent focus ring, not a black box"] -.->|on open| SEL
    F2["Fix 2 even space-sm rhythm, divider margin 0"] -.->|harmonises| DV
    F3["Fix 3 legend lifts while entries show via has-selector"] -.->|clears| LEG
    F4["Fix 4 firstFieldRef forwarded, no querySelector"] -.->|focus target| SEL
    F5["Fix 5 ZIP row gets its own line and can shrink"] -.->|keeps Go in| ZIP
    LEG -.->|"overlap was 43x25, now 0"| ATT
    classDef fix fill:#f5853b,stroke:#9a4d18,color:#1a1a1a
    class F1,F2,F3,F4,F5 fix
```

## Summary

Five review-driven follow-ups to the just-shipped collapsible scope card (#828 / #837). Julian reviewed the live **expanded** card and flagged five things; this fixes all of them. Behaviour + layout changes are unit/e2e-tested; the focus-style + spacing are CSS, verified live across the canonical viewport set in both themes.

[...]

PR #979, opened 2026-06-10
fix(frontend): collapse identity card when scope disclosure is closed (#975)

### LINKED ISSUE #975: fix(frontend): identity card doesn't collapse when scope disclosure is closed (#958 regression)
## Bug

The top-left **identity card** does not collapse to fit its content when the scope disclosure is **closed** — it leaves a large empty white region below the lede ("No recent sightings"). The card stays at the height it has when the disclosure is *open*. Visible in production (Colorado scope, disclosure closed).

## Root cause (regression from #958)

`visibility: hidden` keeps an element **in layout flow** — it reserves its box, just renders it invisible. PR #958 (`feat(frontend): scope disclosure open → #07 panel-reveal`, commit `e362a4c`) replaced the closed-state `display:none` lock on `.app-header-scope-rows` with `visibility:hidden` (so the #07 reveal stays animatable and `toBeHidden()` passes). But `.app-header-scope-rows` is an in-flow flex child of the vertical-column `.app-header-identity-card` (`frontend/src/styles.css:351-353`), so when closed it still reserves the full height of the embedded `ScopeControl` form (state `<select>` + ZIP row + "Whole US / Change scope" links) + the parent `gap` — the empty region under the lede.

- `frontend/src/styles.css:529-534` — `.app-header-scope-rows.t-panel-slide[data-open='false'] { … visibility: hidden }` (in-flow; the bug).
- `frontend/src/components/AppHeader.tsx:316-321` — the always-mounted scope-rows (the `hidden` prop was removed in #958; the element stays mounted for the `aria-controls` target contract).

(The sibling #07 SpeciesDetailRail doesn't hit this because it's `position:absolute` AND unmounts when closed. The scope rows stay mounted + in-flow.)

## Fix — grid `0fr→1fr` clip wrapper (collapse height, keep #07 inner reveal)

Add a clip wrapper around the existing `.t-panel-slide` element that animates `grid-template-rows: 0fr ↔ 1fr` (the modern "tween to auto-height without JS measurement" technique — this is the concrete answer to the #01 "can't tween to `auto`" limitation that pushed this surface to #07). The inner element keeps the **unchanged** #07 transform/opacity/blur reveal.

**Markup** (`AppHeader.tsx`, wrap the scope-rows block; keep `id`/`ref`/`aria-controls` on the INNER element):
```jsx
{scopeActive && (
  <div className="app-header-scope-clip" data-open={scopeOpen}>
    <div id={scopeRegionId} ref={scopeRowsRef} className="app-header-scope-rows t-panel-slide" data-open={scopeOpen}>
      {scopeOpen && <hr className="app-header-divider" aria-hidden="true" />}
      <ScopeControl ref={firstScopeFieldRef} /* …unchanged… */ embedded />
    </div>
  </div>
)}
```

**CSS** (`styles.css`):
```css
.app-header-scope-clip {
  display: grid;
  grid-template-rows: 1fr;
  transition: grid-template-rows var(--panel-open-dur) var(--panel-ease);
}
.app-header-scope-clip[data-open='false'] { grid-template-rows: 0fr; }
.app-header-scope-clip > .app-header-scope-rows { min-height: 0; overflow: hidden; }
```
Keep the existing closed inner rule (`styles.css:529-534`) including `visibility:hidden` — it now sits in a 0-height grid track so it contributes **zero** layout height while remaining out of the a11y tree + tab order.

## Constraints (must all hold)
- transitions.dev catalog only — **no FLIP**, no motion libs, no JS height measurement. The #07 inner reveal is unchanged; the grid-rows tween is the height-collapse mechanism on the same clock (`--panel-open-dur`/`--panel-ease`).
- Reduced-motion: global `motion.css` zeroes durations → grid jumps to `1fr` and the inner jumps to its open rest-state instantly; the form lands fully visible. **No** new per-element `prefers-reduced-motion` query.
- a11y: closed form stays out of the a11y tree + tab order (`visibility:hidden` retained); `frontend/e2e/scope-disclosure.spec.ts` `toBeHidden()` asserts (≈ :96/:97/:164/:178) stay green. Don't change the `id`/`aria-controls`/Page-Object selectors.

## Acceptance criteria
- [ ] Disclosure **closed** → identity card height = wordmark + lede + padding, **no empty region** below the lede, at all 5 viewports × both themes.
- [ ] Disclosure **open** → #07 panel-reveal still plays (translateY 14px→0, opacity 0→1, blur 2px→0) and the card grows smoothly; closing reverses + collapses.
- [ ] Closed form not focusable, absent from the a11y tree.
- [ ] Reduced-motion: open lands fully visible, no off-screen/blurred residue; closed fully collapses.
- [ ] **New regression test:** extend `scope-disclosure.spec.ts` to assert the card collapses when closed — e.g. `.app-header-scope-rows` `boundingBox().height ≤ 1px` (or the closed `.app-header-identity-card` height ≪ its open height) after rest AND after Esc-close. (The existing `toBeHidden()` passes on `visibility:hidden` alone, which is exactly why this regression escaped.)
- [ ] test / lint / build / e2e / knip green; live-verify both the OPEN and CLOSED states; screenshots per protocol — **capture the CLOSED/resting card too**, not just the open form.

Discovered in production after #953 (transitions epic) shipped. Sibling regression to the hover-preview-while-detail-open bug.

---

## Review amendment (julianken-bot, must fold into the fix)

- **IMPORTANT — the parent `gap` leaves an ~8px residual band.** `.app-header-scope-clip` becomes a direct flex child of `.app-header-identity-card` (`display:flex; flex-direction:column; gap: var(--space-sm)`), and a flex `gap` applies **above a zero-height item** — so an ~8px empty band survives below the lede even when the clip collapses to 0fr, contradicting AC #1. Eliminate that gap contribution when closed: move the inter-element spacing **inside** the clip (so it collapses with the grid row), or zero the card's gap for this child when closed. Do NOT leave the card relying on `gap` for the scope-rows spacing.
- **AC correction:** the closed-state regression assertion must check the **card's total height** (no slack) — not just `.app-header-scope-rows` `boundingBox().height ≤ 1px` (that per-row guard passes with the 8px gap still present). Assert the closed `.app-header-identity-card` height ≈ wordmark + lede + padding, materially less than its open height.
- **SUGGESTION — `overflow:hidden` + focus rings.** The clip's `overflow:hidden` can clip the open form's focus rings / the `<select>` native dropdown overflow. When OPEN, ensure overflow doesn't clip interactive affordances (e.g. only clip while collapsing, or use `overflow:clip` with padding, or drop overflow once `data-open='true'`). Verify a keyboard-focused state visually.

julianken:
## Diagrams

State machine — the identity card's height now tracks the disclosure state through a grid-rows clip, instead of freezing at the open height.

```mermaid
stateDiagram-v2
    [*] --> Closed: scope active, disclosure closed (rest)
    Closed --> Open: 🔍 click / openScopeDisclosure()
    Open --> Closed: ✕ click / Esc

    state Closed {
        clip_closed: .app-header-scope-clip[data-open=false]
        note right of clip_closed
            grid-template-rows: 0fr  →  track height 0
            margin-block-start: -8px →  cancels card flex gap
            inner: visibility:hidden + overflow:hidden (a11y out, clipped)
            CARD HEIGHT = wordmark + lede + padding (no empty band)
        end note
    }

    state Open {
        clip_open: .app-header-scope-clip[data-open=true]
        note right of clip_open
            grid-template-rows: 1fr  →  track = content height
            no negative margin       →  card gap = lede→divider 8px
            inner #07 reveal: translateY 14→0, opacity 0→1, blur 2→0
            CARD GROWS to fit the full scope form
        end note
    }
```

## Summary

- **Why:** #958 replaced the closed scope-rows `display:none` lock with `visibility:hidden` (so the #07 panel-reveal stays animatable and `toBeHidden()` passes). But `visibility:hidden` keeps the element **in layout flow** — the closed scope form, an in-flow flex child of the vertical-column identity card, still reserved the full ScopeControl box, so the card froze at its **open** height with a large empty white band below the lede (visible in production, e.g. Colorado scope, disclosure closed).
- **Fix:** wrap the existing `.t-panel-slide` in a `.app-header-scope-clip` grid wrapper that tweens `grid-template-rows: 0fr ↔ 1fr` (the catalog-only, JS-free "tween to content height" technique) on the same `--panel-open-dur`/`--panel-ease` clock. Closed → a 0-height track → the form contributes zero layout height. The inner element keeps its **unchanged** #07 transform/opacity/blur reveal, its `visibility:hidden` rest state, and its `id`/`ref`/`aria-controls` target.
- **Two amendment details folded in:** (1) the card's flex `gap` would still paint an ~8px band above the 0fr-track item, so a negative top-margin cancels it **when closed** (dropped when open, where the gap provides the lede→divider rhythm); the inner element carries no block padding so the `0fr` track floors at a true 0. (2) `overflow:hidden` (the clip) is scoped to the **closed** state only, so an open keyboard focus ring is never clipped. No FLIP, no motion lib, no JS height measure, no new per-element `prefers-reduced-motion` query.
- **CI-caught follow-up (`min-width:0`):** the `.app-header-scope-clip` grid wrapper makes the inner scope-rows a grid item, whose default `min-width:auto` refused to shrink below the embedded ZIP/[Go] min-content width — spilling 32px past the 360px card cap at mobile 390px (`scope-zip-overflow.spec.ts` #842 went red on first push). Added `min-width:0` to the unconditional inner rule (alongside the existing `min-height:0`) so the item is constrained on both axes in both disclosure states. Full e2e re-run: **272 passed / 0 failed**.

## Screenshots

Live-verified on the PR head (`fix/identity-card-collapse`, post-merge of `main`) against a seeded local stack (Colorado scope). **Zero console errors/warnings** at all 5 viewports × both themes. Measured: closed card **88px** (collapsed two-line — scope-rows contributes **0px** layout height) ↔ open card **213px** (#07 reveal landed at rest: opacity 1, transform identity, blur 0). After the `min-width:0` follow-up commit, the mobile-open ZIP/[Go] row sits **~139px inside** the 360px card edge (it was spilling **32px** past — the `scope-zip-overflow.spec.ts` #842 guard, now green).

### Closed — the fix: identity card collapses to its two-line height, no empty band below the lede

| Viewport | Light | Dark |
|---|---|---|
| 390×844 mobile |  |  |
| 768×1024 tablet |  |  |
| 1024×768 landscape |  |  |
| 1440×900 desktop |  |  |
| 1920×1080 wide |  |  |

### Open — disclosure reveal (#07) still plays + mobile [Go] stays inside the card (#842)

| State | Light | Dark |
|---|---|---|
| Desktop 1440 — scope form revealed, card grown |  |  |
| Mobile 390 — ZIP/[Go] row inside the card edge |  |  |

## Test plan

- [x] `tsc --noEmit` — green; `npm test --workspace @bird-watch/frontend` — 1291 passed
- [x] New Playwright e2e spec added: `scope-disclosure.spec.ts` asserts the identity card's **total** height collapses materially below its open height after rest AND after Esc-close (≤2px slack). Verified to **fail** on the pre-fix baseline (open−resting was only ~9px), pass with the fix, and stable across `--repeat-each=3`.
- [x] `npm run lint` (incl. forbidden-raw-token guard) — green; `npx knip` — exit 0; `scripts/check-orphan-classnames.sh` — PASS (`.app-header-scope-clip` has a matching CSS rule)
- [x] (UI) Live-verified via chrome-devtools MCP at all 5 viewports × 2 themes (closed collapse + open reveal + mobile [Go]-inside-card); 14 stills published above; zero console errors/warnings

## Plan reference

Out of plan — bug fix for the #958 regression (identity card doesn't collapse when scope disclosure is closed). Closes #975.

---

### DIFF
--- a/frontend/e2e/scope-disclosure.spec.ts
+++ b/frontend/e2e/scope-disclosure.spec.ts
@@ -166,6 +166,56 @@ test.describe('Scope disclosure (#828)', () => {
     await expect(app.scopeDisclosureTrigger).toBeFocused();
   });
 
+  test('the identity card COLLAPSES to its two-line height when the disclosure is closed (#975)', async ({ page, apiStub }) => {
+    // #975 regression guard: #958 swapped the closed scope-rows lock to
+    // `visibility:hidden`, which keeps the form IN FLOW — the card froze at its
+    // OPEN height with an empty band below the lede. The existing `toBeHidden()`
+    // asserts pass on `visibility:hidden` alone (they never measure layout), so
+    // they could NOT catch this. This test measures the CARD'S TOTAL HEIGHT
+    // (per the review amendment — NOT just `.app-header-scope-rows ≤ 1px`, which
+    // would still pass with an 8px residual flex `gap` band present): the closed
+    // card must be MATERIALLY SHORTER than the open card.
+    const app = await setup(page, apiStub);
+
+    // Read the card's rendered border-box height directly in the page so each
+    // poll re-measures a fresh frame (a Playwright Locator boundingBox snapshots
+    // once). Returns 0 if absent so `expect.poll` can retry rather than throw.
+    const cardHeight = () =>
+      page.evaluate(() => {
+        const el = document.querySelector('.app-header-identity-card');
+        return el ? el.getBoundingClientRect().height : 0;
+      });
+
+    // --- At REST (disclosure closed) ---
+    await expect(app.scopeDisclosureTrigger).toHaveAttribute('aria-expanded', 'false');
+    const restingHeight = await cardHeight();
+
+    // --- OPEN the disclosure and let the grid-rows tween SETTLE ---
+    // The grid-rows + #07 tweens run for --panel-open-dur; poll the SETTLED open
+    // height rather than reading a mid-animation frame. `expect.poll` retries the
+    // measurement until the card has grown materially (the full scope form: state
+    // <select> + ZIP row + Whole US / Change scope links + divider), which only
+    // holds once the tween lands — so `openHeight` below is the settled value.
+    await app.openScopeDisclosure();
+    await expect(app.scopeControlStateSelect).toBeVisible();
+    await expect.poll(cardHeight).toBeGreaterThan(restingHeight + 40);
+    const openHeight = await cardHeight();
+
+    // --- Esc-CLOSE and assert the card COLLAPSES back to its resting height ---
+    await page.keyboard.press('Escape');
+    await expect(app.scopeDisclosureTrigger).toHaveAttribute('aria-expanded', 'false');
+    await expect(app.scopeControlStateSelect).toBeHidden();
+    // Poll until the collapse tween has fully settled back to ≈the resting
+    // (two-line) height — NO empty region below the lede, no residual flex `gap`
+    // band. The closed card must be ≈ wordmark + lede + padding (≤2px sub-pixel
+    // rounding slack), MATERIALLY less than its open height — which the per-row
+    // `scope-rows ≤ 1px` guard would NOT catch (it passes with the gap band
+    // present). This is the missing #975 regression guard.
+    await expect.poll(cardHeight).toBeLessThanOrEqual(restingHeight + 2);
+    const collapsedHeight = await cardHeight();
+    expect(collapsedHeight).toBeLessThan(openHeight - 40);
+  });
+
   test('clicking ✕ collapses the form', async ({ page, apiStub }) => {
     const app = await setup(page, apiStub);
 
--- a/frontend/src/components/AppHeader.tsx
+++ b/frontend/src/components/AppHeader.tsx
@@ -312,26 +312,38 @@ export function AppHeader({
             not lose a typed ZIP). The divider only renders when open — the
             resting card is two lines. */}
         {scopeActive && (
-          <div
-            id={scopeRegionId}
-            ref={scopeRowsRef}
-            className="app-header-scope-rows t-panel-slide"
-            data-open={scopeOpen}
-          >
-            {/* Divider only renders when open — the resting card is two lines.
-                (The ScopeControl below stays mounted regardless so aria-controls
-                always resolves to a present target.) */}
-            {scopeOpen && <hr className="app-header-divider" aria-hidden="true" />}
-            <ScopeControl
-              ref={firstScopeFieldRef}
-              scope={scope as ScopedView}
-              states={states}
-              onPickState={onPickState}
-              onPickWholeUs={onPickWholeUs}
-              onExit={onExitScope}
-              onResolve={onResolveZip}
-              embedded
-            />
+          /* #975: grid 0fr↔1fr CLIP wrapper. The inner .t-panel-slide keeps the
+             UNCHANGED #07 reveal (transform/opacity/blur + visibility); this
+             wrapper is the height-collapse mechanism. Closed → 0fr track →
+             the in-flow form contributes ZERO layout height, so the identity
+             card shrinks to wordmark + lede + padding (no empty band). The
+             id/ref/aria-controls target stays on the INNER element so the a11y
+             contract + Page-Object selectors are untouched. The residual parent
+             flex `gap` band below the lede (which would survive a 0fr collapse)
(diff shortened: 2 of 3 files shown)


julianken:
@Mergifyio queue

PR #1075, opened 2026-06-12
fix(db-client): attach pool-level 'error' handler so an idle-client error never crashes the process

### LINKED ISSUE #1069: test(ci): intermittent unhandled pg-protocol exception fails the test workflow — testcontainers pool not closed before container stop
> **Revised 2026-06-12 after @julianken-bot review** ([comment](https://github.com/julianken/bird-sight-system/issues/1069#issuecomment-4693071501)). The original diagnosis (a teardown-ordering asymmetry between suites) was **factually wrong** — `TestDb.stop` (`packages/db-client/src/test-helpers.ts:41-44`) already does `await pool.end()` *then* `await container.stop()`, and every integration suite ends its pools before stopping its container. The real gap is below.

## Symptom

The `test` CI workflow (the Mergify-gating `npm test --workspaces`) intermittently fails with an **unhandled `pg-protocol` exception**, not an assertion failure:

```
⎯⎯ Unhandled Errors ⎯⎯ / ⎯⎯ Uncaught Exception ⎯⎯
 ❯ parseErrorMessage ../../node_modules/pg-protocol/dist/parser.js:305:11
 ❯ Parser.parse      ../../node_modules/pg-protocol/dist/parser.js:37:38
 ❯ Socket.<anonymous> ../../node_modules/pg-protocol/dist/index.js:11:42
 ❯ TCP.onStreamRead   node:internal/stream_base_commons:191:23
```

Vitest treats an unhandled error as a run failure, so a suite that passes every assertion still reports the `test` job red. Observed on PR #1068 (`0ac523a`); a plain CI re-run of the same SHA went green (confirming transience). `main` (`fe8b7c2`) is green on the same suite, and #1068's diff (`tools/photo-curation/**` + `docs/**`) touches no `pg` code — so the flake is pre-existing and PR-independent.

## Root cause

Production `createPool` (`packages/db-client/src/pool.ts:31`) constructs `new pg.Pool(...)` and **never attaches a `pool.on('error', …)` handler**. Per node-postgres docs: *"the pool will emit an `error` event if a backend or network error occurs on any **idle** client. It is important you add an event listener to the pool to catch errors. Otherwise the node process will exit."* When a pooled client errors while idle — testcontainers tearing the container down between suites in CI, or a network blip against a real backend — the pool emits `'error'` with no listener → uncaught exception → the `test` run fails. The `pg-protocol` socket-read stack (an error parsed off an idle socket, inside no test body) matches this exactly.

This is **not test-only**: the same missing handler means a long-running service (`read-api`, `ingestor`) would crash on any idle-connection error in production. CI's container teardown just surfaces a latent prod gap reliably enough to flake.

## Fix (durable, pg-docs-mandated — touches production code)

Attach a pool-level `'error'` handler in `createPool` that **structured-logs and swallows** (does not rethrow), matching the repo's logging convention (the `{severity, message, …}` shape used elsewhere in db-client):

```ts
const pool = new pg.Pool({ /* …existing… */ });
pool.on('error', (err) => {
  // An idle client erroring must NOT become an uncaught exception (node-postgres
  // docs: an unhandled pool 'error' exits the process). Log and let the pool
  // evict/replace the client.
  console.error(JSON.stringify({ severity: 'ERROR', message: 'db_pool_idle_client_error', error: err.message }));
});
```

Match the exact logger/shape the package already uses. The handler is attached once per constructed pool (the `POOLS` memo means keyed pools get it once). This single production change covers every consumer of `createPool` — CI and prod alike.

## Tests (no real container needed)

- db-client unit test: build a pool via `createPool`, emit `'error'` on it (`pool.emit('error', new Error('idle client boom'), fakeClient)`), assert the process does **not** treat it as uncaught and the handler logged. (A real testcontainers reproduction of the race is timing-dependent and not required once the handler is proven attached.)
- Optional belt-and-suspenders: the integration `test-helpers` pool inherits the same handler since it goes through `createPool` — verify it does (or add it there if test pools bypass `createPool`).

## Acceptance criteria

- [ ] `createPool` attaches a no-rethrow `pool.on('error')` handler that structured-logs; unit test proves an emitted idle-client error does not surface as an uncaught exception.
- [ ] If the flake can be reproduced on the unfixed tree (loop the db-client integration suites), capture one red before/green after; if not reliably reproducible, the test above stands in as the regression guard (note which in the PR).
- [ ] No vitest `retry` / `dangerouslyIgnoreUnhandledErrors` introduced (root-cause fix, per CLAUDE.md).
- [ ] Full CI gate (`test`, `lint`, `build`, `e2e`) green.

## Out of scope
- The promote-scores-to-prod epic (#1074) and the photo-judge eval work.
- Retry/suppression band-aids.

## Evidence
- Failing run: PR #1068 CI `27427422360`, `test` job — uncaught `pg-protocol` parser exception, zero failed assertions; clean re-run green.
- `packages/db-client/src/pool.ts:27-40` — `createPool` constructs the pool with no `'error'` listener.
- `packages/db-client/src/test-helpers.ts:41-44` — `stop()` ends the pool before stopping the container (teardown order already correct; original diagnosis retracted).

julianken:
## Diagrams

```mermaid
sequenceDiagram
    participant Idle as Idle pooled client
    participant Pool as pg.Pool via createPool
    participant Proc as node process

    Note over Idle: backend or network error<br/>CI testcontainers teardown<br/>prod network blip
    Idle->>Pool: emit error event
    alt before this PR — no listener
        Pool-->>Proc: EventEmitter rethrows, uncaught exception
        Proc-->>Proc: process exits, test CI job reds
    else after this PR — pool.on error handler
        Pool->>Pool: console.error structured log, severity ERROR
        Pool->>Pool: pg evicts and replaces the dead client
        Note over Proc: process survives
    end
```

## Summary

- **Why:** node-postgres emits `'error'` on the pool when an *idle* pooled client hits a backend/network error. Per its docs, with no pool-level listener that event becomes an **uncaught exception and the node process exits**. `createPool` (`packages/db-client/src/pool.ts`) built `new pg.Pool(...)` with no such listener, so the Mergify-gating `test` CI job intermittently went red with an unhandled `pg-protocol` parser exception (an error parsed off an idle socket during testcontainers teardown — zero failed assertions, clean re-run green). The same gap would crash `read-api`/`ingestor` on any prod network blip. Fixes #1069.
- **What:** Attach a `pool.on('error', …)` handler in `createPool` that structured-logs in the repo's `{severity, message, …}` shape and **swallows** (does not rethrow — pg evicts/replaces the dead client itself, so rethrowing only re-creates the crash). Attached once per constructed pool (the `POOLS` memo builds keyed pools once), covering every `createPool` consumer in CI and prod. The integration `test-helpers` pool is built directly (not via `createPool`) to keep its harness config (`max: 4`, no `statement_timeout`), but it is the pool most exposed to the teardown-timing flake, so it gets the same guard (belt-and-suspenders, per the issue).
- **Not done:** no vitest `retry` / `dangerouslyIgnoreUnhandledErrors` (root-cause fix per CLAUDE.md); teardown ordering left unchanged — it was already correct (`TestDb.stop` ends the pool before stopping the container), the original teardown-ordering diagnosis was retracted in the issue.

## Screenshots

N/A — not UI

## Test plan

- [x] `npm run build` — clean production build (all workspaces)
- [x] `npm run test -w @bird-watch/db-client` — green (19 files, 163 passed | 1 todo) against real testcontainers Postgres
- [x] New regression test `packages/db-client/src/pool-error-handler.test.ts` (container-free): builds a pool via `createPool`, emits `'error'`, asserts the emit does **not** throw (the uncaught-exception mechanism) and the handler logs in the expected `{severity:'ERROR', message:'db_pool_idle_client_error', error}` shape. Confirmed red on the unfixed tree (EventEmitter rethrew `idle client boom`), green after the fix.
- [x] `npx knip` — exit 0
- [ ] (UI only) Playwright MCP smoke — N/A, no `frontend/**` change

Reproduction note (AC #2): the CI race is timing-dependent and not reliably reproducible on demand; per the issue, the container-free unit test stands in as the regression guard, proving the handler is attached and absorbs an emitted idle-client error.

## Plan reference

Out of plan — bug fix for #1069 (pre-existing, PR-independent CI flake: unhandled `pg-protocol` exception in the `test` workflow).

---

Closes #1069

### DIFF
--- a/packages/db-client/src/pool-error-handler.test.ts
+++ b/packages/db-client/src/pool-error-handler.test.ts
@@ -1,0 +1,65 @@
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import pg from 'pg';
+import { createPool, closePool } from './pool.js';
+
+// These tests need NO real database. They construct a pool (which does not open
+// a connection until a query runs) and assert that an idle-client `'error'`
+// event emitted on the pool is caught by a handler `createPool` attaches —
+// rather than becoming an uncaught exception that exits the process.
+//
+// Regression guard for #1069: node-postgres docs mandate a pool-level `'error'`
+// listener; without one, an idle client erroring (a CI testcontainers teardown
+// blip, or a prod network blip) crashes the process. The `test` CI workflow
+// flaked on exactly this — an uncaught `pg-protocol` parser exception off an
+// idle socket, with zero failed assertions.
+
+afterEach(() => {
+  vi.restoreAllMocks();
+});
+
+describe('createPool idle-client error handling (#1069)', () => {
+  it('attaches a pool-level error listener so an idle-client error is not uncaught', () => {
+    const pool = createPool({ databaseUrl: 'postgresql://unused:unused@127.0.0.1:1/none' });
+    try {
+      // Without a listener, EventEmitter rethrows an emitted 'error' (which is
+      // what becomes the process-killing uncaught exception in CI). The handler
+      // createPool attaches must absorb it.
+      expect(pool.listenerCount('error')).toBeGreaterThan(0);
+    } finally {
+      closePool(pool);
+    }
+  });
+
+  it('swallows (does not rethrow) an emitted idle-client error', () => {
+    const pool = createPool({ databaseUrl: 'postgresql://unused:unused@127.0.0.1:1/none' });
+    const fakeClient = {} as pg.PoolClient;
+    try {
+      // If no handler is attached, EventEmitter throws here. The handler must
+      // not rethrow, so this emit returns cleanly.
+      expect(() =>
+        pool.emit('error', new Error('idle client boom'), fakeClient),
+      ).not.toThrow();
+    } finally {
+      closePool(pool);
+    }
+  });
+
+  it('structured-logs the idle-client error in the repo {severity, message} shape', () => {
+    const errorSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
+    const pool = createPool({ databaseUrl: 'postgresql://unused:unused@127.0.0.1:1/none' });
+    const fakeClient = {} as pg.PoolClient;
+    try {
+      pool.emit('error', new Error('idle client boom'), fakeClient);
+
+      expect(errorSpy).toHaveBeenCalledTimes(1);
+      const logged = errorSpy.mock.calls[0][0];
+      expect(typeof logged).toBe('string');
+      const parsed = JSON.parse(logged as string);
+      expect(parsed.severity).toBe('ERROR');
+      expect(parsed.message).toBe('db_pool_idle_client_error');
+      expect(parsed.error).toContain('idle client boom');
+    } finally {
+      closePool(pool);
+    }
+  });
+});
--- a/packages/db-client/src/pool.ts
+++ b/packages/db-client/src/pool.ts
@@ -35,6 +35,23 @@ export function createPool(opts: PoolOptions): pg.Pool {
     statement_timeout: opts.statement_timeout ?? 15_000,
     connectionTimeoutMillis: opts.connectionTimeoutMillis ?? 10_000,
   });
+  // An idle pooled client that hits a backend/network error emits 'error' on
+  // the pool. node-postgres docs are explicit: without a pool-level listener,
+  // that event becomes an uncaught exception and the node process exits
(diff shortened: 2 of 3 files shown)


julianken:
@Mergifyio queue

PR #1104, opened 2026-06-13
test(frontend): de-flake SpeciesDetailSheet analytics by syncing on the panel_opened effect (#1081)

### LINKED ISSUE #1081: test(frontend): intermittent SpeciesDetailSheet analytics-instrumentation flake (panel_opened timing/isolation race)
## Symptom

The `test` CI workflow intermittently fails in **`frontend/src/components/SpeciesDetailSheet.test.tsx`** → `analytics instrumentation`, with two assertions that are order/timing-sensitive:

```
FAIL  SpeciesDetailSheet > analytics instrumentation > fires panel_opened with has_description=true when descriptionBody is present
  AssertionError: expected "capture" to be called with arguments: [ 'panel_opened', …(1) ]   (line 650)
FAIL  SpeciesDetailSheet > analytics instrumentation > does NOT fire panel_opened before species data resolves
  AssertionError: expected [ [ 'panel_opened', {…} ] ] to have a length of +0 but got 1       (line 675)
```

Observed on PR #1080 (`b7d0062`), whose diff is `tools/photo-curation/**` + `docs/**` only — it touches **zero** frontend files, so the failure is not caused by that PR. A re-run of the same SHA was triggered to confirm transience. `main` is green on this suite, and recent PR/merge test runs are overwhelmingly green — this is a low-frequency flake.

## Why it's not the obvious cause

The component fires `panel_opened` in a `useEffect` keyed on `data?.speciesCode` (`SpeciesDetailSheet.tsx:220`, guarded by `if (!data?.speciesCode) return`). `useSpeciesDetail` keeps a **module-level cache** (`_moduleCache`, `frontend/src/data/use-species-detail.ts:31`) that resolves cache hits **synchronously**. The natural hypothesis — cross-test pollution because the "never resolves" test reuses `speciesCode="vermfly"` from the test directly above it — is **already defended**: the test file calls the exported `__resetSpeciesDetailCache()` in `beforeEach` (`SpeciesDetailSheet.test.tsx:65`, again at `:541`).

So the residual race is subtler. Leading candidates for the repro-loop to discriminate:
- A **microtask/effect from a prior test's async teardown** (a pending `getSpecies`/`getSilhouettes` promise, or the `panel_dwell_ms` cleanup) resolving after the next test has begun, firing `capture` against the new test's spy.
- A **cross-`describe` `beforeEach` gap** — there are multiple `beforeEach` blocks (`:53`, `:972`); confirm the analytics `describe` actually inherits the cache-reset one and isn't shadowed.
- The synchronous cache-hit path interacting with `findByTestId`/`waitFor` ordering so the effect runs a tick earlier/later than the assertion expects.

1. Reproduce by looping just this file (`vitest run SpeciesDetailSheet.test.tsx` in a loop, and/or with `--sequence.shuffle` / varied worker counts) until the two assertions fail, to confirm whether it's intra-file ordering or cross-file module-state.
2. Pin the leak: instrument `analytics.capture` to log which test's spy receives the stray `panel_opened` and from which pending promise/effect.
3. Fix the isolation gap at the source — ensure every pending `getSpecies`/`getSilhouettes` promise and the dwell-cleanup are flushed/awaited before a test ends (e.g. resolve the never-resolving promise in an `afterEach`, or assert via `findBy*`/`waitFor` rather than a bare post-render read), and confirm the cache reset covers the analytics `describe`.
4. Do NOT add vitest `retry` or `dangerouslyIgnoreUnhandledErrors` (CLAUDE.md: fix the root cause).

## Acceptance criteria
- [ ] The race identified and named; the two analytics assertions made deterministic (looped ≥30× locally with `--sequence.shuffle`, zero failures).
- [ ] No retry/suppression; no production component behavior changed (test-isolation fix, unless a real effect-timing bug in `SpeciesDetailSheet.tsx` is found).
- [ ] Full CI gate green.

## Out of scope
- The promote-scores epic (#1074) and the photo-judge eval work.

## Evidence
- Failing run: PR #1080 CI `27431998489`, `test` job — two `SpeciesDetailSheet` analytics assertions (`:650`, `:675`), zero relation to the PR diff.
- Component effect: `frontend/src/components/SpeciesDetailSheet.tsx:220-235` (`panel_opened` keyed on `data?.speciesCode`).
- Cache + reset: `frontend/src/data/use-species-detail.ts:31` (`_moduleCache`), `:34` (`__resetSpeciesDetailCache`), reset at `SpeciesDetailSheet.test.tsx:65`.

julianken:
## Diagrams

```mermaid
sequenceDiagram
    participant Test as analytics test
    participant React
    participant Effect as panel_opened useEffect
    participant Spy as analytics.capture spy

    Note over Test,Spy: OLD path gates on an unrelated render output
    Test->>React: render SpeciesDetailSheet
    React-->>Test: heading committed
    Note right of Test: waitFor heading settles
    Test->>Spy: read mock.calls synchronously
    Note right of Spy: panel_opened not yet flushed, length 0, flake at line 650
    React->>Effect: passive effect flushes a tick late
    Effect->>Spy: capture panel_opened bleeds into next test, flake at line 675

    Note over Test,Spy: NEW path synchronizes on the effect itself
    Test->>React: render SpeciesDetailSheet
    React->>Effect: passive effect flushes
    Effect->>Spy: capture panel_opened
    Note right of Test: waitFor polls the spy, not the heading
    Test->>Spy: assertion passes once capture observed
    Note over Test,Spy: afterEach drains queue, no cross-test bleed
```

## Summary

- **Root cause:** the `analytics instrumentation` tests assert on a side effect of a passive `useEffect` (`analytics.capture('panel_opened', …)`, keyed on `data?.speciesCode`) but synchronize only on an *unrelated render output* — the heading appearing (positive tests) or the sheet shell mounting (the never-resolving negative test). React 18 flushes passive effects on a tick separate from the commit that renders the heading, and `test-setup.ts`'s `afterEach` `cleanup()` unmounts between tests without awaiting a still-scheduled effect. Under scheduler jitter the `panel_opened` effect lands a tick late, producing two leak directions both seen in #1081 CI: positive tests read the spy before the effect flushed (`length 0`, issue `:650`); the never-resolving "does NOT fire" test sees a *prior* test's late-flushing effect bleed into its fresh spy (`got 1`, issue `:675`).
- **Fix (test-only — `SpeciesDetailSheet.tsx` is unchanged):** positive tests now poll the assertion on the `panel_opened` capture itself via `waitFor` (wait for the effect instead of racing it); the negative test drains the passive-effect queue + macrotask boundary, clears the spy, drains again, then asserts zero (isolated to a settled queue); a describe-scoped `afterEach` drains the queue after every analytics test so nothing leaks across the test boundary.
- **No retry / no suppression / no `.skip`** — root cause only, per CLAUDE.md.

## Screenshots

N/A — test-only (frontend/** test-isolation fix, no UI change)

## Test plan

- [x] `npm run test --workspace @bird-watch/frontend` — green (1403/1403, full suite run 3x)
- [x] **Causation proof:** with OLD test code + a 5ms macrotask lag injected into the effect (worst-case jitter emulation), all three #1081 failure modes reproduce deterministically (9/15 runs: `:650`-class `length 0`, `has_description=true` not-called, `:675` `got 1`). Fixed tests pass **0/30** under the same lag.
- [x] **Determinism:** real (pristine) production code — `SpeciesDetailSheet.test.tsx --sequence.shuffle` looped **40×** (0 failures), and **30×** under 6-core CPU load (0 failures).
- [x] `npm run lint` — exit 0
- [x] `npm run build` — exit 0 (clean production build)
- [x] `npx knip` — exit 0
- [ ] New unit / integration tests added (if behavior changed) — N/A, isolation fix to existing tests
- [ ] New Playwright e2e spec added — N/A, no user-visible behavior change
- [x] (UI only) Playwright MCP smoke — N/A, test-only change, no UI surface touched

## Plan reference

Out of plan — flaky-test fix for #1081 (intermittent `test` CI failure unrelated to any feature PR's diff).

Closes #1081

---

### DIFF
--- a/frontend/src/components/SpeciesDetailSheet.test.tsx
+++ b/frontend/src/components/SpeciesDetailSheet.test.tsx
@@ -1,6 +1,6 @@
 import { readFileSync } from 'node:fs';
 import { join } from 'node:path';
-import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { render, screen, waitFor, act } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { SpeciesDetailSheet, resolveContentTier } from './SpeciesDetailSheet.js';
@@ -609,6 +609,23 @@ describe('<SpeciesDetailSheet>', () => {
       return client;
     }
 
+    // Flush React's passive-effect queue and the microtask/macrotask boundary.
+    //
+    // `panel_opened` fires from a passive useEffect keyed on data?.speciesCode
+    // (SpeciesDetailSheet.tsx). React 18 schedules passive effects on a tick
+    // SEPARATE from the commit that renders the heading, and the cleanup() in
+    // test-setup.ts unmounts between tests WITHOUT awaiting any
+    // still-scheduled effect. Without an explicit barrier, a prior analytics
+    // test's `panel_opened` effect could flush a tick LATE — into the next
+    // test, whose freshly-installed spy then records a stray call (the
+    // "does NOT fire" test's `length 0 → got 1` flake). Draining the queue
+    // here, after every analytics test, makes the cross-test boundary clean.
+    afterEach(async () => {
+      await act(async () => {
+        await new Promise<void>(resolve => setTimeout(resolve, 0));
+      });
+    });
+
     it('fires panel_opened once on data resolve with has_description=false when no description', async () => {
       const captureSpy = vi.spyOn(analytics, 'capture');
       render(
@@ -619,11 +636,16 @@ describe('<SpeciesDetailSheet>', () => {
           mainRef={{ current: mainEl }}
         />,
       );
-      await waitFor(() =>
-        expect(screen.getByRole('heading', { name: 'Vermilion Flycatcher' })).toBeInTheDocument(),
-      );
-      const openCalls = captureSpy.mock.calls.filter(([name]) => name === 'panel_opened');
-      expect(openCalls).toHaveLength(1);
+      // Synchronize on the EFFECT under test (the panel_opened capture), not on
+      // an unrelated render output. The heading appears on the data-arrival
+      // commit, but the panel_opened passive effect can flush a tick later under
+      // scheduler jitter; gating on the heading then reading the spy
+      // synchronously raced that effect (the `length 0` flake, #1081). waitFor
+      // polls until the effect itself has fired.
+      await waitFor(() => {
+        const openCalls = captureSpy.mock.calls.filter(([name]) => name === 'panel_opened');
+        expect(openCalls).toHaveLength(1);
+      });
       expect(captureSpy).toHaveBeenCalledWith('panel_opened', {
         species_code: 'vermfly',
         has_description: false,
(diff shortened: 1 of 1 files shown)


julianken:
@Mergifyio queue

PR #1111, opened 2026-06-13
test(frontend): de-flake clarity.test.ts via mock re-resolution after resetModules (#1105)

### LINKED ISSUE #1105: test(frontend): intermittent clarity.test.ts flake — vi.stubEnv/resetModules env-bleed
## Symptom

Under parallel full-suite load, `frontend/src/clarity.test.ts` (the `calls Clarity.init …` case) intermittently fails. Surfaced while stress-testing the suite during the #1081 de-flake (looping with `--sequence.shuffle` + high worker counts); **not** reliably reproducible on an idle machine — a low-frequency, load-dependent flake, distinct in mechanism from #1081.

## Root-cause hypothesis (to confirm via repro-loop, not assume)

Cross-file environment bleed via `vi.stubEnv('PROD', …)` + `vi.resetModules()`. The Clarity wrapper (`frontend/src/clarity.ts`, the `safeClarity` guard — see [the SDK-no-preinit memory rule]) reads env at module init; a stubbed `PROD` from one test file, combined with `resetModules`, can leak into another file's fresh import depending on worker reuse and teardown ordering, so `Clarity.init` is/ isn't called against the wrong env.

1. Reproduce by looping the file (and the suite) with `--sequence.shuffle` / varied `maxThreads` until it fails; capture the order.
2. Pin whether the leak is intra-file (stub not restored) or cross-file (worker module-state reuse). Confirm every `vi.stubEnv` is paired with `vi.unstubAllEnvs()` in `afterEach`, and that `resetModules` doesn't strand a half-initialised Clarity module.
3. Fix the isolation at the source. **No** `retry` / `.skip` / `dangerouslyIgnoreUnhandledErrors` (repo CLAUDE.md forbids it).

## Acceptance criteria
- [ ] Race named; the test made deterministic (looped ≥30× with `--sequence.shuffle`, zero failures).
- [ ] No retry/suppression; no production change unless a real `clarity.ts` bug is found.
- [ ] Full CI gate green.

## Provenance
Diagnosed during the #1081 fix (PR #1104) while stress-loading the frontend suite. Sibling flake also filed for `MapSurface.test.tsx`. Out of scope: #1081 (fixed), the photo-judge eval work.

julianken:
## Diagrams

```mermaid
sequenceDiagram
    participant Test as clarity.test.ts (prodinit case)
    participant Reg as Vitest module registry
    participant Clar as ./clarity.js
    participant Mock as @microsoft/clarity (mock)

    Note over Test: beforeEach → vi.resetModules() (invalidates registry incl. mock)
    Test->>Test: vi.stubEnv(PROD=true, ID=abc123xyz)
    Test->>Reg: await import('./clarity.js')
    Reg->>Clar: fresh eval — top-level reads stubbed env
    Clar->>Mock: Clarity.init('abc123xyz')  ✅ live mock instance
    Note over Test,Mock: BEFORE — assert on top-level const initMock = vi.fn()<br/>captured at collection time → DIFFERENT instance → 0 calls → FLAKE
    Note over Test,Mock: AFTER — currentInitSpy() re-imports '@microsoft/clarity'<br/>→ SAME instance clarity.ts called → deterministic
```

## Summary

- **Why it flaked:** the per-test `vi.resetModules()` invalidates the mocked `@microsoft/clarity`; under `--sequence.shuffle` the freshly-imported `clarity.ts` bound a *different* mock `init` instance than the top-level `const initMock = vi.fn()` the test asserted on — so `Clarity.init` fired on the live mock while `expect(initMock)` saw zero calls. Reproduced deterministically (seeds 3 and 9; ~55% of random-shuffle file-only runs). In-module instrumentation confirmed the identity mismatch (guard=true, isMock=true, distinct instance); `vi.stubEnv` was correct at eval time, so the original cross-file-env-bleed hypothesis is refuted.
- **The fix (test-only):** resolve the `init` spy through the same mocked `@microsoft/clarity` specifier *after* the dynamic import, so the assertion always targets the exact instance `clarity.ts` just called, regardless of how `resetModules` re-wired the registry. No retry/skip/suppression.
- **`clarity.ts` is unchanged** — the production env-gating guard was proven correct by the instrumentation.

## Screenshots

N/A — test-only (frontend/** test-isolation fix)

## Test plan

- [x] `npm run test --workspace @bird-watch/frontend` — green (87 files, 1413 tests, 0 failures; after `npm run build -w @bird-watch/shared-types` to provide the workspace artifact the suite resolves)
- [x] New unit / integration tests added (if behavior changed) — N/A; reworked the existing assertions, no behavior change
- [x] New Playwright e2e spec added (if user-visible behavior changed) — N/A; no user-visible change
- [x] `npm run build` — clean production build (frontend `tsc -b && vite build`)
- [x] Determinism proof — previously-failing fixed seeds 3 and 9 pass 5/5 each; **40× random-seed `--sequence.shuffle` loop on `clarity.test.ts` with zero failures** (was ~55% fail before); `npm run lint` and `npx knip` both clean (exit 0)
- [ ] (UI only) Playwright MCP smoke — N/A — not UI

## Plan reference

Out of plan — flake fix for issue #1105 (diagnosed during the #1081 de-flake, PR #1104).

Closes #1105

---


julianken:
@Mergifyio queue

Issue #1193, opened 2026-06-14
Epic: eleatic trace tree + span inspector


julianken:
## Canonical shared contracts (authoritative — every child cites verbatim)

Every issue MUST target these interfaces VERBATIM. They are the anti-drift anchor. File paths are repo-root-relative. All references re-anchored against origin/main (the local main checkout is 35 commits stale and does NOT contain packages/eleatic — verify with `git show origin/main:<path>`, never the working tree).

PRECEDENT FACTS (verified on origin/main, load-bearing):
- Existing renderTrace (ui/trace.js) renders a flat `{spans:[{name,input,output,usage}]}`; its usage line is "12 prompt · 34 completion · 250 ms · $cost" (NO combined token count, NO "tok" suffix). It is total/never-throws (trace.test.ts). It STAYS, unchanged — the drawer's inline preview + the non-conforming fallback.
- EvalRowRecord.trace is `trace?: unknown` (src/types.ts) — opaque at the store boundary; getRow → /api/row is the only read that layers it back. NO schema/serde/store/queries change anywhere in this epic.
- Today's producer buildTraceSpan (tools/photo-curation/src/eval/eleatic-adapter.ts:127-145) emits ONE span: `{spans:[{name:'judge', input:{prompt,species:{comName,sciName,family},rubricVersion,model,imageUrl?}, output:{parsed,raw?}, usage:{latencyMs,promptTokens?,completionTokens?,costUsd?}}]}`.
- The runner (scripts/run-eval-local.ts) imports keepAgreement/scoreMAE/keepConfusion (line 46) — NOT criteriaAxisMAE. It already computes keep=keepAgreement().score, mae=scoreMAE().score, confusion=keepConfusion().metadata for accumulators, then builds traceInput and calls buildTraceSpan(traceInput) (~:218), then eleatic.recordRow(toEleaticRow(result, trace)).
- Scorer names/shapes (src/eval/scorers.ts): keepAgreement→{name:'keep_agreement',score:0|1}; scoreMAE→{name:'score_mae',score:[0,1]}; keepConfusion→{name:'keep_confusion',score:0|1, metadata:{falseKeep,falseReplace}}; criteriaAxisMAE→AxisScoreResult[] one per CRITERIA_KEYS axis, {name:'criteria_mae_<axis>', score:number|null} where null = deliberate axis-skip.
[...]

PR #1195, opened 2026-06-14
feat(eleatic): EvalSpan/EvalTrace doc types + pure backward-compat trace-tree builder

julianken:
## Diagrams

`buildTraceTree(trace)` — the pure, never-throws render-time tree builder. One conformance gate, then three load-bearing shape branches, then a single normalize + depth pass:

```mermaid
flowchart TD
    A["buildTraceTree(trace)"] --> B{"object && Array.isArray(trace.spans)?<br/>(same gate as trace.js)"}
    B -- no --> X["{ ok:false, roots:[] }<br/>caller falls back to renderTrace(trace)"]
    B -- yes --> C{"spans.length === 0?"}
    C -- yes --> E0["{ ok:true, roots:[] }"]
    C -- no --> D{"any span has a string id?"}
    D -- "no (legacy id-less)" --> L["wrap ALL spans under one synthesized<br/>{ id:'legacy:root', name:'trace', kind:'eval' }<br/>children get 'legacy:&lt;i&gt;' ids, depth 1"]
    D -- "yes (keyed)" --> K["index by id (first wins);<br/>link by parentId;<br/>orphan → root (never dropped);<br/>dup id → demote later to root;<br/>self-cycle / back-edge → re-root<br/>(node-identity seen-set ⇒ terminates)"]
    L --> N["normalize metrics per node:<br/>span.metrics ?? map(span.usage) ?? undefined<br/>(latencyMs→durationMs the only rename)"]
    K --> N
    N --> O["assign depth (root 0, child +1) +<br/>stable sibling order<br/>(array order; startMs sort only when BOTH finite)"]
    O --> R["{ ok:true, roots: TraceNode[] }"]
```

## Summary

- **Why a render-time builder, not a stored tree:** `trace_json` stays one OPAQUE blob eleatic never destructures (`EvalRowRecord.trace` remains `trace?: unknown`); the tree is reconstructed only when a renderer needs it. This is the single anti-drift point Wave-2 children (T2 producer, T3/T4 panes) all import.
- **Why it must tolerate everything:** `buildTraceTree` is PURE and never throws across today's legacy id-less flat trace (wrapped under one synthesized `legacy:root` so even legacy data shows a tree — Decision 3), a keyed `parentId` tree (orphan-tolerant, duplicate-id-safe, cycle-safe via a node-identity seen-set), and any non-conforming blob (`ok:false` → caller uses the lossless `renderTrace` fallback).
- **Why the normalization lives here:** this is the SINGLE place legacy `usage` is mapped to a node's canonical `metrics` (`latencyMs`→`durationMs` the only rename); T3/T4 renderers read `node.metrics` and never touch `span.usage`. `EvalSpan`/`EvalTrace` are documentation types only and do NOT retype `EvalRowRecord.trace`, so legacy + non-conforming blobs still round-trip.

## Screenshots

N/A — packages/ tool (eleatic ui/ is not frontend/**).

## Test plan

- [x] `npm run test -w @bird-watch/eleatic` — green (20 files, 221 tests, incl. the existing `no-coupling.test.ts`; new `trace-tree.test.ts` adds 28 cases)
- [x] New unit tests added — `ui/trace-tree.test.ts` covers the conformance gate (non-object / missing-spans / scalar / array → `ok:false`), legacy id-less → single synthesized root + `legacy:` ids + `usage`→`metrics` mapping, keyed `eval→task→judge+scorers` depths & child order, orphan promotion, duplicate-id demotion, self-cycle / back-edge re-rooting + termination, stable sibling order, `startMs` sort only when both finite, and a never-throws totality sweep
- [x] `npm run build -w @bird-watch/eleatic` — clean (tsc)
- [x] `npx knip --no-progress` — green (only the pre-existing unrelated `map-v1-prototype` hint; `trace-tree.js` is auto-traced by its `.test.ts` sibling, so NO knip ignore entry)
- [x] `npm run lint` — green
- [ ] (UI only) Playwright MCP smoke — N/A, no `frontend/**` change

## Plan reference

Part of epic #1193 (T1).

---

### DIFF
--- a/packages/eleatic/src/index.ts
+++ b/packages/eleatic/src/index.ts
@@ -11,6 +11,9 @@
 // --- Write store (E1) ---
 export { openStore, EleaticStore } from './store.js';
 export type { EvalRunRecord, EvalRowRecord, EvalAdjudicationRecord } from './types.js';
+// Doc types for the conventional tree layered over the opaque trace blob (the
+// trace-tree epic). Additive — EvalRowRecord.trace stays `trace?: unknown`.
+export type { EvalSpan, EvalTrace } from './types.js';
 
 // --- Read query API (E2) ---
 export { makeReader } from './queries.js';
--- a/packages/eleatic/src/types.ts
+++ b/packages/eleatic/src/types.ts
@@ -53,6 +53,57 @@ export interface EvalRowRecord {
   trace?: unknown;
 }
 
+/**
+ * The conventional TREE layout layered over the opaque `trace_json` blob.
+ *
+ * DOCUMENTATION types only: they describe the shape `buildTraceTree`
+ * (ui/trace-tree.js) reconstructs a tree from, and the shape the photo-curation
+ * producer emits. They DO NOT retype `EvalRowRecord.trace`, which STAYS
+ * `trace?: unknown` so legacy + non-conforming blobs still round-trip unchanged.
+ * A producer MAY locally type its trace value as `EvalTrace`; the store and the
+ * read path remain blob-opaque.
+ *
+ * A span is keyed by `id` and points at its parent by `parentId` (null = a
+ * root). `kind` is free-form and producer-owned — eleatic STYLES by it but never
+ * branches on its semantics. `metrics` is the canonical camelCase metrics object
+ * new producers emit; a legacy span instead carries `usage` (latencyMs, …),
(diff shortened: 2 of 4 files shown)


julianken:
@Mergifyio queue

PR #1207, opened 2026-06-14
fix(ingestor): decouple grid-agg precompute from live reads to stop periodic 503s

julianken:
## Diagrams

```mermaid
sequenceDiagram
    participant Cron as Cloud Scheduler
    participant Job as bird-ingestor (recent)
    participant PG as Cloud SQL (db-g1-small)
    participant User as Live /api/observations

    Note over Cron,PG: BEFORE - contention burst at :00 / :30
    Cron->>Job: :00 recent tick
    Job->>PG: refreshGridAgg (DELETE+INSERT, ~430MB pgsql_tmp spill, pins 1 vCPU)
    Cron->>Job: :02 cache-warm fires 77 reads into the window
    User->>PG: aggregated CTE (state/low-zoom fallback)
    PG--xUser: 15s statement_timeout -> 503

    Note over Cron,PG: AFTER - decoupled
    Job->>PG: refreshGridAgg + SET LOCAL work_mem=96MB (smaller spill, shorter window)
    Cron->>Job: :08 cache-warm after refresh settles
    User->>PG: aggregated CTE
    PG-->>User: 200 OK (~200ms)
```

## Summary

- **Why:** bird-maps.com serves intermittent 503s on `/api/observations`, clustered at the `:00`/`:30` ingest ticks and chronic since at least 2026-05-22. The 503s are origin-emitted (`canceling statement due to statement timeout`), not a Cloudflare problem — Cloudflare's `stale-while-revalidate` actually masks most of them. Latency is bimodal (~200 ms normal vs a hard 15 s wall), the signature of resource starvation, not heavy compute.
- **Root cause:** the `recent` ingest's `refreshGridAgg` precompute rebuilds the whole `observation_grid_agg` table every 30 min and, at the Cloud SQL default `work_mem`, spills ~430 MB to `pgsql_tmp` — pinning the `db-g1-small` instance's single shared vCPU + disk. Concurrent live read-path aggregations (the state/low-zoom CTE that `getAggregatedGridFromCache` falls through to) starve and blow the 15 s `statement_timeout`. `bird-cache-warm` at `2,32` made it worse: its `+2 min` offset assumed `recent` finishes in 2 min, but at national scale it runs ~4.5 min, so the warmer's ~77 reads fired straight into the precompute window.
- **Fix (decouple, chosen over a DB-tier bump):** (1) `db-client` — a transaction-scoped `SET LOCAL work_mem = '96MB'` in `refreshGridAgg` shrinks the spill + contention window (conservative for a ~1.7 GB shared-core box; fully eliminating the spill needs ~192 MB+/node which would risk an instance OOM — that's the argument for a larger tier, tracked separately). (2) `infra` — move `bird-cache-warm` `2,32` → `8,38` so it runs after `recent`+refresh settle.

## Screenshots

N/A — not UI (db-client query + Terraform scheduler cron).

## Test plan

- [x] `npm run build --workspace @bird-watch/db-client` (tsc typecheck) — clean
- [x] Full `db-client` suite (`vitest run`) — **187 passed + 1 todo across 21 files**, including a new test asserting `SET LOCAL work_mem` is issued after `BEGIN` and before the populate (mirrors the #878 `statement_timeout` test)
- [x] New unit/integration test added (work_mem ordering)
- [x] `terraform fmt -check` on `infra/terraform/ingestor.tf` — clean
- [ ] N/A — not UI (no Playwright e2e / MCP smoke)

**Deploy note:** this does not take effect on merge alone — the `work_mem` change ships in the next `deploy-ingestor` image rollout, and the `cache-warm` cron change requires a `terraform apply` (the repo's `terraform-plan-drift-check` is IAM-blocked, so applies are manual).

## Plan reference

Out of plan — production 503 incident remediation (2026-06-14). Investigated via GCP logs + Chrome DevTools live repro + Cloud SQL log correlation. Follow-up (separate issue): eligible state low-zoom views fall through to the live CTE because `getAggregatedGridFromCache` returns empty for the scope — serving them from the grid would make those reads PK lookups immune to contention.

---


julianken:
@Mergifyio queue

PR #1208, opened 2026-06-14
fix(frontend): anchor keyboard-focus hover preview + hide it immediately on leave

julianken:
## Diagrams

Two follow-ups to #1179's `CellHoverPreview` teardown fix. `CellHoverPreview` renders bimodally on `cursorPos`: a viewport point ⇒ `position:fixed` + portaled to `<body>` (properly positioned); `null` ⇒ CSS-anchored `position:absolute`, inline in the marker, **no offsets** (unpositioned, clipped). Both fixes live in `AdaptiveGridMarker`'s per-cell handlers.

```mermaid
flowchart TD
    subgraph Issue1["Issue 1 — keyboard focus (a11y)"]
        F[Tab to cell ⇒ onCellFocus] -->|"BEFORE: cursorPos = null"| B1["CSS fallback: position:absolute, inline,\nno offsets ⇒ tooltip stuck at marker corner, clipped"]
        F -->|"AFTER: cursorPos = cell.getBoundingClientRect()"| A1["fixed + portaled, anchored just below the cell\n(same mode as the cursor path)"]
    end
    subgraph Issue2["Issue 2 — hide latency"]
        L[mouse-leave cell ⇒ onCellMouseLeave] -->|"BEFORE: setTimeout 250ms (spec §4.5 dwell)"| B2["tooltip lingers ~250ms ⇒ visible lag"]
        L -->|"AFTER: setTimeout 0ms (next tick)"| A2["disappears immediately; an adjacent-cell\nre-enter still cancels it ⇒ no cross-cell flicker"]
    end
```

## Summary

- **Issue 1 (a11y / keyboard):** Tabbing to a cell rendered the hover preview in its CSS-anchored fallback (`position:absolute`, inline, no offsets, not portaled) — an unpositioned tooltip glued to the marker corner and clipped, the same "messed up format" the old mouse-leave glitch showed. `onCellFocus` now anchors `cursorPos` to the focused cell's `getBoundingClientRect()`, so the preview renders in the same `fixed`+portaled mode as the cursor path, positioned just below the cell.
- **Issue 2 (hide latency):** The preview lagged ~250ms after the pointer left (the spec §4.5 dwell). Dropped to a next-tick (`0ms`) deferral so it hides the moment the mouse leaves. The deferral is still load-bearing — entering an adjacent cell sets `activeCell` to the new cell before the `0ms` timer runs, so its index guard no-ops and the preview swaps cells instead of blinking through an empty frame.
- Both build on #1179 and keep its no-mode-flip guarantee (cursorPos is never cleared on leave).

## Screenshots

N/A (live capture) — unit-test evidence. Issue 1 has a visual aspect (keyboard-tooltip placement), but it is pinned deterministically by unit assertions (`position:fixed`, anchored to the cell rect at `left/top` from `getBoundingClientRect`) rather than a static screenshot, and the local map stack is currently held by a concurrent dev session. Live keyboard-placement capture can follow once the stack frees up if desired.

- [x] N/A — no new/renamed `className` in this PR (handler-logic change only)
- [x] N/A — multi-viewport stills: keyboard-tooltip placement is pinned by unit assertions; live capture deferred (contended stack)

## Test plan

- [x] Typecheck (`tsc -b`) exit 0 + full frontend `vitest run` green (1552 tests, incl. 4 new/updated for these fixes)
- [x] New unit tests (RED→GREEN confirmed):
  - `a11y: keyboard focus anchors the preview (position:fixed near the cell), not the broken CSS fallback` — was `''`, now `'fixed'` at the cell rect
  - `tooltip disappears within a frame of mouse-leave (no 250ms lag)` — was present through 250ms, now gone within one frame
  - `moving between adjacent cells swaps the preview without unmounting (no flicker)` — guards the next-tick deferral
  - `mouse-leave keeps the cursor-anchored render until unmount — no mode-flip` — retains the #1179 guarantee
- [x] N/A — UI-only Playwright MCP smoke: agreed unit-test-evidence bar (contended stack); visual aspect pinned by the assertions above

## Plan reference

Out of plan — direct follow-up to #1179 (`CellHoverPreview` teardown). Two issues reported live: keyboard-focus tooltip format + hide latency. No tracking issue.

---


julianken:
@Mergifyio queue

Issue #1221, opened 2026-06-14
epic(frontend): basemap-style decoupling + 5-theme selector (BasemapDescriptor)


julianken:
## Goal

Decouple every map-rendering subsystem from the concrete OpenFreeMap basemap style, then expose a user-facing selector for **5 themes**. Today the code hardcodes facts about two specific styles (positron land cream `#f4f1ea`, dark land `#0E1116`, a label-id regex, a luminance gate, marker halo `#ffffff`, float outline/halo hexes). This epic replaces that with a single **central table** mapping a theme name to a `BasemapDescriptor`; consumers read facts off an injected descriptor and **never branch on the concrete theme**.

> Core principle (Julian): the concrete theme/style is NEVER referenced directly in the code that uses it. Consumers receive an injected `BasemapDescriptor` and read facts off it. Only **one** central table maps a theme name to a descriptor.

## Architecture (approved)

**`BasemapDescriptor`** — one plain object per style:

| field | type | meaning |
|---|---|---|
| `id` | `string` | label/test identifier — **never branched on** |
| `url` | `string` | OpenFreeMap style URL (`https://tiles.openfreemap.org/styles/<id>`) |
| `kind` | `'light' \| 'dark'` | canvas polarity that DRIVES the app chrome `[data-theme]` so the whole UI syncs |
| `landColor` | hex | dominant land hex — family-palette + float-contrast reference |
| `markerHaloColor` | hex | silhouette `icon-halo-color` |
| `floatColors` | `{ outline, halo }` | state-artboard float layer colors |
| `darkLabelTextColors` | `{ road, place, water }` | only required for `kind: 'dark'` styles |

**Injection = PARAMETER injection** into the existing **pure** helper functions. No React context, no module global. The descriptor is resolved **once** at the swap site and threaded down as an explicit argument.

### The reactive source of truth is the THEME ID, not `[data-theme]` (load-bearing — fixes the lossy-trigger trap)

[...]

Issue #1213, opened 2026-06-14
feat(frontend): active-theme-id state + id-driven basemap-swap effect (replaces the [data-theme] trigger)


julianken:
## Problem (the lossy-trigger trap — HIGH)

The entire basemap swap is keyed on the `[data-theme]` attribute, which only ever holds `'light'`|`'dark'`. `use-state-artboard.ts` (lines 346-400 on `origin/main`) runs a `MutationObserver` on `[data-theme]` that fires `map.setStyle(next === 'dark' ? BASEMAP_DARK : BASEMAP_LIGHT)` (the `setStyle` call is at line 384-385); `MapCanvas.tsx` resolves the initial URL from `getAttribute('data-theme') === 'dark'` (lines 1058-1060, 1942-1944). With the 5-theme model there are TWO dark-kind themes (`dark`, `fiord`) and THREE light-kind (`positron`/`bright`/`liberty`). Switching WITHIN a kind (dark→fiord, positron→bright) does NOT change `descriptor.kind`, so `[data-theme]` does not change, the observer never fires, and the basemap URL never swaps. **`[data-theme]` is structurally incapable of triggering same-kind swaps (5→2 is lossy); 3 of 5 themes would be unreachable.**

This child introduces the **active `ThemeId` React state** as the reactive source of truth and rewrites the swap effect to depend on the ID (not the attribute), so C2/C3/C4 can consume an already-threaded descriptor and C8 can reach all 5 themes. **Behavior-preserving:** only `positron`/`dark` are reachable here (C1's `ThemeId` union is still `'positron' | 'dark'`), so the observable behavior (toggle flips light↔dark and swaps the basemap) is unchanged — this PR is the plumbing, not the user-facing change. **No visible change.**

## The injection seam (READ FIRST — this is what makes the regression guard reachable)

The whole point of this child is a regression guard proving **the id — not the kind — drives the swap**. But C1 (#1212) defines `ThemeId = 'positron' | 'dark'` (two DIFFERENT kinds), `resolveDescriptor` as a **closed `THEME_REGISTRY[id]` lookup**, and the swap effect resolves the descriptor internally from the id. With only that API there is **no seam to inject two same-kind descriptors**, so the guard would silently collapse into a cross-kind `positron`↔`dark` assertion — which does NOT exercise the trap (any kind-keyed mechanism passes a cross-kind test). This child MUST create the seam. Do BOTH:

[...]

Issue #1212, opened 2026-06-14
feat(frontend): BasemapDescriptor type + registry + resolver + isLabelLayer + LAND_COLORS


julianken:
## Problem

The map subsystems hardcode facts about two specific OpenFreeMap styles. There is no single object that carries a style's facts, and no central table mapping a theme name to those facts. Today `frontend/src/components/map/geometry/basemap-style.ts` only exports two raw URL strings (`BASEMAP_LIGHT = 'https://tiles.openfreemap.org/styles/positron'`, `BASEMAP_DARK = 'https://tiles.openfreemap.org/styles/dark'`) plus deprecated `basemapStyle*` aliases (knip-ignored at `knip.ts:202-222`).

This child lays the **pure, fully unit-tested foundation**: the `BasemapDescriptor` type, the `THEME_REGISTRY` table, `resolveDescriptor`, the runtime `isLabelLayer` detector, AND a frozen `LAND_COLORS` table that declares all 5 lands + kinds up front (the single source of truth that C5 and C6 both consume — this is what breaks the C5↔C6 mutual-dependency cycle). It registers **only** `positron` and `dark` DESCRIPTORS so behavior is reproduced exactly; the other 3 descriptors are added in C6 (but their LAND colors live here from the start). **No visible change.**

## Files

- EDIT `frontend/src/components/map/geometry/basemap-style.ts` — becomes the descriptor registry + resolver + `LAND_COLORS`. KEEP `BASEMAP_LIGHT`/`BASEMAP_DARK` as thin aliases (`BASEMAP_LIGHT = THEME_REGISTRY.positron.url`, `BASEMAP_DARK = THEME_REGISTRY.dark.url`) so all current callers (`MapCanvas.tsx:14,189,1058-1060,1942-1944`, `use-state-artboard.ts:4,384`) keep type-checking and rendering identically. KEEP the `basemapStyle*` deprecated aliases too (do NOT touch knip.ts in this PR — alias removal + the matching knip rule deletion happen together in C6).
- EDIT `frontend/src/components/map/geometry/basemap-style.test.ts` — extend the existing test file.

## Implementation outline

1. Types:
```ts
export type ThemeId = 'positron' | 'dark'; // C6 widens this union to all 5
export type BasemapKind = 'light' | 'dark';
export interface BasemapDescriptor {
  /** Stable id for labels/tests. NEVER branched on in rendering code. */
  id: ThemeId;
  url: string;                              // OpenFreeMap style URL
[...]

PR #1225, opened 2026-06-14
refactor(frontend): marker icon-halo-color onto descriptor.markerHaloColor (reactive memo deps)

### LINKED ISSUE #1216: refactor(frontend): marker icon-halo-color onto descriptor.markerHaloColor (reactive memo deps)
## Problem

`frontend/src/components/map/geometry/observation-layers.ts` (`buildUnclusteredPointLayerSpec`, ~line 369) hardcodes the silhouette marker halo: `'icon-halo-color': '#ffffff'`. The comment ties it to 'the light OpenFreeMap positron basemap.' The halo color is a per-style design fact and belongs on the descriptor.

Refactor `buildUnclusteredPointLayerSpec` to take the active `BasemapDescriptor` and read `descriptor.markerHaloColor`. **Behavior-preserving** for `positron`/`dark` (both declare `markerHaloColor: '#ffffff'`, so the paint is byte-identical).

## Correction from review (the 'rebuilt on style.load' claim is FALSE — load-bearing)

The original draft claimed the marker layer is 'rebuilt on style.load, the same re-apply path the theme swap uses.' **That is wrong.** `MapCanvas.tsx:697-698` build the specs via `useMemo(() => build…Spec(), [])` with EMPTY dependency arrays — they are react-map-gl declarative `<Layer>` specs built ONCE at mount, never recomputed on style.load or swap. The icon-halo paint only updates when react-map-gl diffs a NEW spec object. Therefore the descriptor MUST be reactive React state present in the memo deps:
```ts
const unclusteredLayer = useMemo(() => buildUnclusteredPointLayerSpec(descriptor), [descriptor]);
```
where `descriptor` is the reactive active-theme descriptor from C1.5's `useActiveThemeId()`. Without the memo-dep change, passing `resolveDescriptor(id)` with `[]` deps freezes the halo at the mount-time descriptor and it NEVER changes on swap. The update mechanism is react-map-gl's paint diff on a NEW spec object — NOT a style.load fixup.

## Files

- EDIT `frontend/src/components/map/geometry/observation-layers.ts`
- EDIT `frontend/src/components/map/geometry/observation-layers.test.ts`
- EDIT `frontend/src/components/map/MapCanvas.tsx` — change the `unclusteredLayer` memo (line 698) to `useMemo(() => buildUnclusteredPointLayerSpec(descriptor), [descriptor])` using the C1.5 reactive descriptor.

## Implementation outline

1. `buildUnclusteredPointLayerSpec()` → `buildUnclusteredPointLayerSpec(descriptor: BasemapDescriptor): LayerProps`.
2. Replace `'icon-halo-color': '#ffffff'` with `'icon-halo-color': descriptor.markerHaloColor`. Leave `icon-halo-width: 1.5` and the rest unchanged.
3. Update the inline comment: the halo is now style-aware; for positron/dark it remains `#ffffff`.
4. In `MapCanvas.tsx`, add `descriptor` to the memo deps (line 698).

## The notable-ring amber is INTENTIONALLY kind-coupled — carved OUT of this PR (documented)

`buildNotableRingLayerSpec` (lines ~404-428) uses `circle-stroke-color: notableColor()` which reads `--color-accent-notable-fg`; `AdaptiveGridMarker.tsx` picks `NOTABLE_AMBER_LIGHT #c43a1a` vs `NOTABLE_AMBER_DARK #f59e0b` by `kind` (per C50/#1032: `#f59e0b` on cream `#f4f1ea` is 1.90:1). This is a basemap-luminance-coupled fact, but the epic DECIDES to keep it **kind-coupled, NOT theme-coupled** — it is a CSS-token-driven design choice, not a per-style descriptor field. Rationale: the amber is a semantic accent shared with non-map chrome, and folding it into every descriptor would over-thread it. **C4 does NOT change the notable ring.** Instead, C5 adds an acceptance that BOTH amber values clear 3:1 against ALL 5 lands (so the kind-coupling stays valid on bright/liberty/fiord). The `notable-ring` paint, being kind-driven (CSS token), already re-reads on a kind flip; same-kind switches keep the same amber by design (acceptable: the amber is a kind-level accent, not a per-land legibility fix).

## Acceptance criteria

- `buildUnclusteredPointLayerSpec(descriptor)` reads `descriptor.markerHaloColor`; no literal `'#ffffff'` remains for the marker halo.
- `positron`/`dark` descriptors → `paint['icon-halo-color'] === '#ffffff'` (byte-identical).
- The `unclusteredLayer` memo at `MapCanvas.tsx:698` includes the reactive `descriptor` in its dependency array, so a live theme swap recomputes the spec (the AC that catches the latent C8 bug).
- The notable ring is unchanged (carve-out documented); C5 owns the both-ambers-vs-all-lands assertion.
- `npm run build`, `lint`, `test`, knip green. No visible change for positron/dark.

## Test expectations (unit)

In `observation-layers.test.ts`:
- `buildUnclusteredPointLayerSpec(positronDescriptor).paint['icon-halo-color'] === '#ffffff'`; same for `darkDescriptor`.
- A synthetic descriptor with `markerHaloColor: '#123456'` → `paint['icon-halo-color'] === '#123456'` (proves wiring, not a hardcode).
- Rest of spec (`icon-size 0.85`, `icon-color ['get','color']`, the `icon-opacity` case expression, `icon-halo-width 1.5`) unchanged.
- NOTE for C8: the live click-through e2e (in C8) must assert that switching to a theme with a non-`#ffffff` halo actually changes the rendered halo — C4's unit test proves only the pure function; the live-swap guarantee lives in C8's e2e because that is where a non-white-halo theme becomes reachable.

## Depends on

C1 (needs `BasemapDescriptor`) and C1.5 (reactive descriptor for the memo deps).

Part of #1221

julianken:
## Diagrams

```mermaid
flowchart LR
    Swap["theme swap → setThemeId(id)"] --> Hook["useActiveThemeId()"]
    Hook -->|"reactive descriptor"| Memo["useMemo(() => buildUnclusteredPointLayerSpec(descriptor), [descriptor])"]
    Memo -->|"NEW spec object"| RMG["react-map-gl &lt;Layer&gt; paint diff"]
    RMG --> Paint["icon-halo-color = descriptor.markerHaloColor"]
```

Before this change the `unclusteredLayer` memo had an EMPTY dep array and the halo was the literal `'#ffffff'`. react-map-gl `<Layer>` specs are built once at mount and only re-diff on a NEW spec object — they are NOT rebuilt on `style.load` or the theme swap. So the halo could only ever update if the memo produced a new spec. Threading the reactive `descriptor` into the deps is what makes a future non-white-halo theme actually re-render.

## Summary

- The silhouette marker halo (`icon-halo-color`) was hardcoded `'#ffffff'` in `buildUnclusteredPointLayerSpec` — a per-style design fact that belongs on the `BasemapDescriptor`. The spec now takes the active descriptor and reads `descriptor.markerHaloColor`.
- `MapCanvas.tsx` memoizes the spec on the reactive `[activeDescriptor]` from C1.5's `useActiveThemeId()` (was `[]`). Without this, the halo would freeze at the mount-time descriptor and never update on a live swap — the latent bug C8's e2e asserts against.
- **Behavior-preserving / no visual delta**: `positron` and `dark` descriptors both declare `markerHaloColor: '#ffffff'`, so the production paint is byte-identical. The notable-ring amber stays kind-coupled (CSS token), carved out per epic #1221; C5 owns the all-lands amber assertion.

## Screenshots

N/A — behavior-preserving refactor, no visual delta; halo value unchanged via descriptor (positron/dark both carry `markerHaloColor: '#ffffff'`, so the rendered halo is byte-identical to before).

- [x] Every new/renamed `className` in this PR has a matching CSS rule — `N/A — no className changes`
- [x] Multi-viewport design QA — `N/A — not UI` (behavior-preserving; no visual change)

## Test plan

- [x] `npm run test --workspace @bird-watch/frontend` — green (91 files / 1580 tests)
- [x] New unit tests added: positron→`#ffffff`, dark→`#ffffff` (byte-identical), and a synthetic descriptor with `markerHaloColor: '#123456'` → paint `'#123456'` (proves wiring, not a hardcode). TDD: synthetic test went red first, then green after the impl.
- [x] No new Playwright e2e — no user-visible behavior change (the non-white-halo live-swap guarantee lives in C8's e2e, where such a theme first becomes reachable)
- [x] `npx tsc -b frontend` — exit 0; `npx knip` — exit 0; `bash scripts/ci/check-orphan-classnames.sh` — PASS; `npm run lint` — exit 0
- [x] (UI only) Playwright MCP smoke — `N/A — behavior-preserving refactor, no visual delta`

## Plan reference

Closes #1216. Part of #1221 (child C4). Depends on C1 (#1212) and C1.5 (#1213), both merged to `main`.

---

### DIFF
--- a/frontend/src/components/map/MapCanvas.tsx
+++ b/frontend/src/components/map/MapCanvas.tsx
@@ -525,7 +525,8 @@ export function MapCanvas({
   // only reachable production ids are `positron`/`dark`, so the observable
   // behavior is unchanged; this is the plumbing that lets the swap depend on the
   // id (not the lossy `[data-theme]` attribute) so C2–C4/C8 can reach all 5 themes.
-  const { themeId: activeThemeId } = useActiveThemeId();
+  const { themeId: activeThemeId, descriptor: activeDescriptor } =
+    useActiveThemeId();
 
   // State-artboard machinery consolidated into ONE hook (`use-state-artboard.ts`,
   // epic #884 · U13 / #898; the #760/#762/#763/#765/#849/#850 blank-map class).
@@ -710,7 +711,15 @@ export function MapCanvas({
   const clusterCountLayer = useMemo(() => buildClusterCountLayerSpec(), []);
   const clustersHitLayer = useMemo(() => buildClustersHitLayerSpec(), []);
   const notableRingLayer = useMemo(() => buildNotableRingLayerSpec(), []);
-  const unclusteredLayer = useMemo(() => buildUnclusteredPointLayerSpec(), []);
+  // #1216: the marker halo color now comes from the active descriptor, so the
+  // spec must be recomputed when the descriptor changes (a live theme swap).
+  // react-map-gl `<Layer>` specs are built once at mount and only re-diff on a
+  // NEW spec object — without `activeDescriptor` in the deps the halo would
+  // freeze at the mount-time descriptor and never update on swap.
+  const unclusteredLayer = useMemo(
+    () => buildUnclusteredPointLayerSpec(activeDescriptor),
+    [activeDescriptor],
+  );
 
   // Observation lookup by subId for click handler. Pure derive extracted to
   // obs-derive.ts (#892); prototype-free record built there.
--- a/frontend/src/components/map/geometry/observation-layers.test.ts
+++ b/frontend/src/components/map/geometry/observation-layers.test.ts
@@ -12,6 +12,7 @@ import {
   CLUSTER_RADIUS,
   FALLBACK_SILHOUETTE_ID,
 } from './observation-layers.js';
+import { resolveDescriptor, type BasemapDescriptor } from './basemap-style.js';
 import { FAMILY_COLOR_FALLBACK } from '@/data/family-color.js';
 
 function makeObs(partial: Partial<Observation> = {}): Observation {
@@ -205,7 +206,12 @@ describe('layer specs', () => {
     // Each feature renders its family's silhouette via icon-image, tinted by
     // the per-feature color property (sourced from the silhouettes join in
     // observationsToGeoJson).
-    const spec = buildUnclusteredPointLayerSpec();
+    //
+    // #1216: the marker halo is now a per-style descriptor fact — the spec
+    // takes the active BasemapDescriptor. positron carries
+    // markerHaloColor '#ffffff', so this assertion is byte-identical to the
+    // pre-refactor hardcode.
+    const spec = buildUnclusteredPointLayerSpec(resolveDescriptor('positron'));
     expect(spec.id).toBe('unclustered-point');
     expect(spec.type).toBe('symbol');
     // Epic #539 cutover: inStack filter removed alongside auto-spider.
@@ -224,6 +230,8 @@ describe('layer specs', () => {
     // SDF tint sourced from the per-feature color property — this is what
     // makes the silhouettes share the family color from the legend.
     expect(paint['icon-color']).toEqual(['get', 'color']);
+    // #1216: halo width is unchanged; only the color is threaded.
+    expect(paint['icon-halo-width']).toBe(1.5);
     // _FALLBACK silhouette renders at 50% opacity so missing-Phylopic
     // families read as "we don't have a shape for this" instead of
     // visually equal to the rest. The first branch (issue #554 scope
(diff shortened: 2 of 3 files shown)


julianken:
@Mergifyio queue

Issue #1240, opened 2026-06-16
feat(frontend): "Copy link to this view" control + live-camera exposure


julianken:
## Problem

Part 1 of epic #1238 needs an in-app, console-free way to **capture** the current view as a link. This child adds a **"Copy link to this view"** control to the top-right controls cluster. On click it reads the live camera + viewport, builds the link via the C1 codec, and writes it to the clipboard with an accessible confirmation.

**Hidden prerequisite (plan-review finding):** `App.tsx` holds **no map ref**, and `MapSurface`'s `onViewportChange(bounds, zoom)` gives App only bounds + zoom — **no center / bearing / pitch / dpr**. The MapLibre `getMap()` ref is private to `MapCanvas`. So this child must first **expose the live camera** to the header before the button can read it.

## Files

- EDIT `frontend/src/components/map/MapCanvas.tsx` + `frontend/src/components/MapSurface.tsx` + `frontend/src/App.tsx` — **camera exposure wiring** (see Implementation). Pick ONE approach:
  - **(a) `useImperativeHandle`** on MapCanvas exposing `getCamera(): { center:[lng,lat]; zoom; bearing; pitch }`, forwarded through MapSurface to a ref held in App; OR
  - **(b) widen** the existing `onViewportChange` payload to also carry `center / bearing / pitch`, store the latest in a `cameraRef` in App.
  - Either way, read `window.innerWidth / innerHeight / devicePixelRatio` directly for the `&v=` tag (globals — fine); source zoom/center/bearing/pitch from the map, not from globals.
- ADD `frontend/src/components/CopyViewLinkButton.tsx` — the pill button + its confirmation live region.
- EDIT `frontend/src/components/AppHeader.tsx` — mount `<CopyViewLinkButton/>` as a **sibling pill inside `.app-header-controls-pill`**, adjacent to the ⓘ Credits and Filters buttons; match their icon-button contract (verify the current button order + insertion point in the file — do not trust a stale line number).
[...]

Issue #1238, opened 2026-06-16
epic(frontend): shareable viewbox links — capture & restore exact map view (camera in URL hash)


julianken:
## Goal

Make an exact map view a **first-class, copyable artifact**, so a prod bug can be reproduced — by Julian later, by a teammate, or by Claude — without pasting a console function. Today the URL serializes scope + filters + theme (`frontend/src/state/url-state.ts`), but the **camera (center / zoom / bearing / pitch) is never serialized** — it is derived from scope via `fitBounds` / `maxBounds` / `flyTo` in `frontend/src/App.tsx`. So there is no durable, console-free way to capture, share, or replay a precise pan / zoom.

This is especially load-bearing for **data-inconsistency bugs**, where the same place renders or fetches differently across viewports or zooms: the `/api/observations` bbox is **viewport-derived** (`map.getBounds()`), and there is an **aggregated ↔ per-observation mode boundary at zoom 6** (decided server-side in `services/read-api/src/app.ts:243`; the frontend reads `envelope.mode` in `frontend/src/data/use-bird-data.ts`). Reproducing those requires the exact camera **and** viewport, neither of which the URL carries.

This epic adds:

- **Part 1 — capture + replay:** a "Copy link to this view" control that emits a link encoding the exact camera + viewport, plus a committed **replay tool** that opens N such links at their captured viewport / camera and diffs both the render and the live `/api/observations` request+response — the mechanism for reproducing a multi-view inconsistency.
- **Part 2 — restore:** the app reads the camera back from the URL on load (cooperating with the scope `fitBounds` machinery), so a link becomes a plain "just open it" deep link.

Design validated by a 4-lens plan-review (URL-state/codec · camera/fitBounds restore · capture UX/contract/a11y · testing/decomposition) before filing.

## Link format (the "viewbox")

Existing query params (`?state=`, `?scope=us`, `?species=`, `?family=`, `?since=`, `?notable=`, `?view=`, `?detail=`) are **unchanged**. The **camera** is added to the URL **hash** (MapLibre-standard; the hash is verifiably free — zero `location.hash` usage anywhere in `frontend/src`):

```
#map=<zoom>/<lat>/<lng>[/<bearing>[/<pitch>]]
```

- `zoom` `toFixed(3)`; `lat`/`lng` `toFixed(5)` (~1.1 m); `bearing`/`pitch` `toFixed(1)`, **emitted only when non-zero**.
[...]

PR #1250, opened 2026-06-19
fix(frontend): reset success-check on idle so the Copied checkmark doesn't stick

julianken:
## Diagrams

State of the two nested transitions-dev recipes on the `copied → idle` reset. The OUTER `.t-icon-swap` was reset; the INNER `.t-success-check` was not — and its `forwards`-fill `opacity:1` out-ranks the outer's static `opacity:0`, so the drawn check stuck on top of the link glyph. The fix adds the inner reset.

```mermaid
stateDiagram-v2
    direction LR
    state "copied (dwell)" as Copied {
        outer_b: outer .t-icon-swap data-state b (show check)
        inner_in: inner .t-success-check data-state in (forwards opacity 1)
    }
    state "BEFORE fix (idle)" as Before {
        outer_a1: outer reset to a (static opacity 0 for check)
        inner_stuck: inner STILL in (forwards opacity 1 wins) - check stuck
    }
    state "AFTER fix (idle)" as After {
        outer_a2: outer reset to a (static opacity 0 for check)
        inner_out: inner reset to out - static opacity 0 wins, check fades
    }
    Copied --> Before: setCopyState('idle') only
    Copied --> After: setCopyState('idle') + checkRef -> data-state 'out'
```

## Summary

- **Why:** clicking "Copy link to this view" left a stuck checkmark overlapping the link glyph after the "Copied" confirmation finished. The animation in (link → drawn check) was fine; only the *return to the link* was broken.
- **Root cause:** recipe 09 (icon-swap, link⇄check) wraps recipe 10 (success-check). On copy, `replayCheck()` sets the inner `.t-success-check` wrapper's `data-state="in"`, which runs `t-check-fade … forwards` — pinning the check at `opacity:1` and KEEPING it via the `forwards` fill. The idle-reset callback (`setTimeout(…, COPIED_MS)`) only called `setCopyState('idle')`, flipping the OUTER `.t-icon-swap` back to `'a'` (static `opacity:0` for the now-hidden check icon). A `forwards` animation value overrides a static `opacity:0`, and nothing reset the inner wrapper's `data-state` from `"in"` back to `"out"` — so the drawn check stayed at `opacity:1` on top of the link.
- **Fix:** in both reset-timer callbacks (the `copied` branch and the `error` branch), reset the check wrapper with `checkRef.current?.setAttribute('data-state', 'out')` alongside `setCopyState('idle')`. Dropping the `data-state="in"` removes the `forwards`-opacity so the icon-swap's static `opacity:0` wins and the check fades cleanly back to the link. `replayCheck` (out→in on copy) is unchanged, and recipe 10's `forwards` stays so the check remains DRAWN during the dwell — the bug was only the missing reset on exit.

## Screenshots

Live verification (orchestrator). This is a **JS state-reset motion fix — no chrome / layout / theme change** — so the 5-viewport × 2-theme matrix is N/A; the meaningful evidence is the *temporal* behavior (the check must vanish after the 1.6s dwell) + the deterministic regression test.

**Before** (the reported bug): after clicking, the success-check stayed painted over the link glyph. **After** (this fix): click → confirmation plays → after the dwell the check returns cleanly to the link, `checkDataState` reset to `"out"` (was stuck at `"in"`). Verified live, both themes (`checkDataState==='out'` + `data-state` swap back to `a` confirmed):

| Light (rested after click) | Dark (rested after click) |
|---|---|
| rested light | rested dark |

The regression test in `CopyViewLinkButton.test.tsx` asserts the check returns to `"out"` after `COPIED_MS` — **fails on the un-fixed code, passes with the fix**. Console 0/0; the during-dwell animation is unchanged (recipe-10 `forwards`-fill still keeps the check drawn through the dwell — only the missing exit-reset was added).

      (or marked `N/A — class is consumed only by a primitive that ships its own CSS`) — **N/A — no className added/renamed; this is a JS state-reset fix, no CSS or markup change.**
- [ ] Multi-viewport design QA: PR body has ≥10 `user-attachments` URLs
      (5 viewports × 2 themes per #445) — **PENDING — orchestrator captures and attaches.**

## Test plan

- [x] `npm run -w @bird-watch/frontend test` — green (1731 passed / 94 files; CopyViewLinkButton suite 9 passed)
- [x] New unit test added — regression test for the stuck-check exit (asserts `.t-success-check` reaches `data-state="in"` during `copied`, then returns to `"out"` after `COPIED_MS`). Verified it FAILS with the fix reverted (check stays `"in"`) and PASSES with it, while the other 8 button tests are unaffected.
- [ ] New Playwright e2e spec added (if user-visible behavior changed) — N/A; the unit test deterministically pins the `data-state` lifecycle, which is the exact mechanism of the bug.
- [x] `npm run -w @bird-watch/frontend build` — clean production build (`tsc -b && vite build`, typechecks the test changes)
- [x] `npm run lint` — green (repo `lint` runs `npm run lint --workspaces --if-present`; no workspace defines a lint script, so it is a no-op, plus the forbidden-token grep — unaffected, no CSS token usage touched)
- [ ] (UI only) Playwright MCP smoke — **PENDING — orchestrator drives the motion fix live (subagents cannot drive a browser).**

## Plan reference

Out of plan — small UI motion bug fix on the C2 CopyViewLinkButton (#1240, epic #1238): the success-check is not reset on the idle return, leaving a stuck checkmark over the link glyph.

---

### DIFF
--- a/frontend/src/components/CopyViewLinkButton.test.tsx
+++ b/frontend/src/components/CopyViewLinkButton.test.tsx
@@ -1,5 +1,5 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
-import { render, screen, waitFor } from '@testing-library/react';
+import { render, screen, waitFor, fireEvent, act } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { CopyViewLinkButton } from './CopyViewLinkButton.js';
 import type { ViewboxCamera } from '@/state/viewbox-link.js';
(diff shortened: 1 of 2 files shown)


julianken:
@Mergifyio queue

PR #1249, opened 2026-06-18
feat(frontend): restore camera from #map= hash on load + write-back

julianken:
## Diagrams

Cold-load restore + write-back across the App → MapSurface → useScopeCamera seam:

```mermaid
sequenceDiagram
    participant URL as URL hash (#map=)
    participant App as App.tsx
    participant Hook as useScopeCamera
    participant Map as MapLibre

    Note over URL,Map: Cold load — ?state=US-AZ#map=z/lat/lng
    App->>App: parse hash ONCE (useRef init) → rawHashCamera
    App->>Hook: initialHashCamera (first paint) + hashCameraInScope (null→validated)
    Hook->>Map: initialViewState = hash camera (no scope/CONUS flash)
    Note over App: /api/states resolves real AZ envelope
    App->>Hook: hashCameraInScope = true (center inside AZ)
    Hook->>Map: jumpTo(hash) + SUPPRESS scope fitBounds (first run only)
    Hook-->>App: restoredHashCamera → data-hash-camera attribute

    Note over URL,Map: User pans (settle window closed, scope active)
    Map->>Hook: idle
    Hook->>URL: replaceState(encodeViewbox(live camera)) — debounced, idempotent

    Note over URL,Map: Out-of-scope hash (Texas under AZ) OR a later scope change
    App->>Hook: hashCameraInScope = false / boundsKey flips
    Hook->>Map: fitBounds(scope envelope) — hash ignored
```

## Summary

- **Why:** Part 2 of epic #1238 (#1242 / C4). A copied `#map=<z>/<lat>/<lng>` viewbox link should be self-restoring — a teammate opens it in a normal browser and sees exactly the captured view, no replay tool needed. The camera must cooperate with the scope `fitBounds`/`maxBounds` machinery and survive subsequent in-app navigation.
- **The camera seam is `use-scope-camera.ts`** (the `fitBounds`/`flyTo` effect was extracted there in #884). A third, highest-precedence branch restores the hash on the FIRST run only (`hashSettledRef` + a mount-value `boundsKey` guard so the holding→real `/api/states` transition keeps the same key and never clobbers it); a genuine later scope change flips `boundsKey` and reframes normally. The debounced idle write-back is `replaceState`-only + idempotent, gated on `scopeActive` + the existing `scopeMoveUntilRef` settle window. `maxBounds` stays scope-derived (untouched).
[...]

PR #1246, opened 2026-06-17
feat(frontend): "Copy link to this view" control + live-camera exposure

julianken:
## Diagrams

Live-camera bridge + the click→copy→confirm flow (clipboard only — the app URL bar is never mutated; that is Part 2):

```mermaid
sequenceDiagram
    participant User
    participant Pill as CopyViewLinkButton
    participant App
    participant Map as MapCanvas
    participant Codec as encodeViewbox (C1)
    participant Clip as Clipboard

    Note over Map,App: MapCanvas populates App's cameraRef with a LIVE getter on style.load
    User->>Pill: click
    Pill->>App: getCamera()
    App->>Map: cameraRef.current()
    Map-->>App: zoom / center / bearing / pitch (live)
    App-->>Pill: ViewboxCamera (or null before ready → no-op)
    Pill->>Codec: encodeViewbox(camera, viewport)
    Codec-->>Pill: map=z/lat/lng&v=WxH@dpr
    Pill->>Clip: writeText(origin+pathname+search+#+fragment)
    alt write ok
        Clip-->>Pill: resolved
        Pill->>Pill: icon link→check (09+10), label Copy link→Copied! (04), role=status announce
    else denied / unavailable
        Pill->>Clip: execCommand('copy') fallback
        Pill->>Pill: selectable prefilled field + "Press Cmd/Ctrl C" + role=status
    end
```

State machine: `idle → copying → copied (1600ms) → idle`; on clipboard failure `copying → error (2600ms) → idle`. One reset timer, cleared on re-click + unmount.

## Summary

- **Why a live-camera bridge:** `App` held no map ref and `onViewportChange` exposed only bounds+zoom — no center/bearing/pitch/dpr. Rather than convert the 2000-line `MapCanvas` to `forwardRef`, App passes a `cameraRef` down (MapSurface → MapCanvas) that MapCanvas fills with a **live getter** reading the camera on each call. So a click — even mid-pan, before an `idle` settles — copies exactly what's on screen. Returns `null` (clean no-op) before `mapReady` / while unscoped.
- **Why clipboard-only:** mutating the app's own URL bar now would hit the `writeUrl` hash-drop bug and `#511`-class races — that's Part 2 (C3/C4). The query (scope/filters/theme) rides on `location.search`; the camera + capture viewport ride in the hash via C1's `encodeViewbox`.
[...]

PR #1262, opened 2026-06-21
fix(frontend): de-flake copy-view-link e2e — isolate the click assertion from the #1249 camera write-back

julianken:
## Diagrams

The failing assertion races two independent URL actors. The Copy-link click is clipboard-only; the #1249 write-back independently writes `#map=`. The old test conflated them; the fix synchronizes on the write-back, then measures the click's delta alone.

```mermaid
sequenceDiagram
    participant Test as e2e spec
    participant Btn as CopyViewLinkButton click
    participant Clip as navigator clipboard
    participant WB as use-scope-camera write-back
    participant URL as window location and history

    Note over WB,URL: independent of any click, fires once the scope-settle window closes
    WB-->>URL: replaceState writes the map hash

    rect rgb(230,245,230)
        Note over Test: FIX poll until the map hash lands, THEN snapshot
        Test->>URL: capture before-url and before-history-length
    end

    Test->>Btn: click
    Btn->>Clip: writeText link, clipboard ONLY
    Note over Btn,URL: click NEVER touches location or history

    Test->>URL: assert url equals before-url
    Test->>URL: assert history length equals before-history-length
```

## Summary

- **Why it failed:** `copy-view-link.spec.ts:132` captured `before = "…/?scope=us"` and asserted the URL never contains `#map=`. But the #1249 camera→hash write-back (an independent `map.on('idle')` listener in `use-scope-camera.ts`) serializes the live camera into `#map=` via `replaceState` regardless of any Copy-link interaction — so the bar legitimately becomes `…/?scope=us#map=3.791/38.56065/-97.50000` between snapshot and assertion. Ordering bug: #1246 introduced the test when nothing wrote `#map=`, then #1249 added the write-back and invalidated its premise. The e2e check has failed on every main push since (689a289 #1246, ec56752 #1249, 85fc5c4 #1250).
- **Why it's test-only:** the `CopyViewLinkButton` click handler reads the live camera and writes the clipboard (+ execCommand fallback) — it never touches `location`/`history`. Verified by reading the handler. No app coupling bug; this is a test-isolation race.
- **The fix:** synchronize on the independent write-back first (`expect.poll(page.url).toContain('#map=')`), snapshot URL + `history.length`, click, wait past the 300ms idle debounce, then assert the click changed NEITHER. The test now asserts its real contract — the copy CLICK does not further mutate the URL — and still fails if Copy-link ever starts writing the bar (value would differ, or a `pushState` would bump `history.length`). Drops the now-incompatible `not.toContain('#map=')` line.

## Screenshots

N/A — not UI (test-only). The only change is `frontend/e2e/chrome/copy-view-link.spec.ts`; no `frontend/src/**` and no visual/CSS surface is touched, so the canonical-viewport Playwright-MCP capture protocol is exempt per CLAUDE.md ("Test-only … PRs under `frontend/**` are exempt").

## Test plan

- [x] Reproduced the CI failure from the e2e run log on main (85fc5c4): `Received "http://localhost:5173/?scope=us#map=3.791/38.56065/-97.50000"` at `copy-view-link.spec.ts:145`.
- [x] Ran the fixed spec against a real local stack (read-api on 8787 + Vite on 5173 + chromium): **all 4 copy-view-link tests pass**, including the previously-failing line 132 (`the Copy-link CLICK does not mutate the app URL bar`).
- [x] `npm run lint` — no-op for this change (no workspace `lint` script; the lint-guard greps only `frontend/src/`, untouched here).
- [ ] N/A — no new unit/integration test (the existing e2e is the behavior under repair).
- [ ] N/A — no new e2e spec (this repairs an existing one); no user-visible behavior change.
- [x] Authoritative gate is this PR's CI `e2e` check.

## Plan reference

Out of plan — bug fix. De-flakes the e2e regression introduced by the viewbox-links epic (#1238: #1246 / #1249 / #1250) that has been failing the required `e2e` check on main since 2026-06-18.

---

### DIFF
--- a/frontend/e2e/chrome/copy-view-link.spec.ts
+++ b/frontend/e2e/chrome/copy-view-link.spec.ts
@@ -129,7 +129,9 @@ test.describe('Copy link to this view (C2 · #1240)', () => {
     ).toHaveCount(1);
   });
 
-  test('does NOT mutate the app URL bar (clipboard only — Part 2 owns URL writes)', async ({ page }) => {
+  test('the Copy-link CLICK does not mutate the app URL bar (clipboard only — #1249 write-back owns URL writes)', async ({
+    page,
+  }) => {
     await installClipboardSpy(page);
     const app = new AppPage(page);
     await app.goto();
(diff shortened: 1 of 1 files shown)


julianken:
@Mergifyio queue

Issue #298, opened 2026-04-27
infra(terraform): IAM grant for drift-check + CF map-v1 import + read-api FRONTEND_ORIGINS [G2 / IMPORTANT]


julianken:
## Context

Three Type-8 (Infra↔Live) gaps identified in the 2026-04-27 codebase drift audit. None require frontend changes; all are pure infrastructure/configuration work. They are grouped because fixing them in the wrong order is wasteful: the IAM grant (sub-task 1) must land first so the drift-check workflow can actually validate the Cloudflare import (sub-task 2) and the env-var injection (sub-task 3).

### Sub-task 1 — B2: Grant `storage.objects.list` to drift-check SA (GCP IAM only)

The `terraform-plan-drift-check.yml` workflow has failed every scheduled run since it was created. The workflow YAML's `permissions:` block is correctly structured (`id-token: write`); the failure is purely at the GCP IAM layer.

**Error:** `Failed to get existing workspaces: querying Cloud Storage failed: googleapi: Error 403: *** does not have storage.objects.list access`

**Fix:** Grant the deploy service account (referenced via `GCP_DEPLOY_SA_EMAIL`) `roles/storage.objectViewer` (or a custom role with `storage.objects.list` + `storage.objects.get`) on the `gs://bird-maps-tfstate` bucket. No YAML change needed.

**Evidence:** Latest failing run — https://github.com/julianken/bird-sight-system/actions/runs/24992811057 (2026-04-27T11:38:47Z). Leaf: `docs/analyses/2026-04-27-codebase-drift-audit/leaves/L10-github-ci.json` finding L10-3.

---

### Sub-task 2 — #235 follow-through: Import 4 Cloudflare map-v1 resources into TF state

Four resources were deployed live via PR #164 but were never imported into Terraform state. They exist in `infra/terraform/map-v1.tf` but `terraform state list` does not include them. A `terraform apply` would attempt to recreate them and name-conflict. They are currently allowlisted in `.github/drift-allowlist.yml` with expiry **2026-05-08** — the import must complete before then or the drift-check workflow (once IAM-fixed by sub-task 1) will fire on all four as phantom creates.

Resources to import:
- `cloudflare_r2_bucket.pmtiles`
- `cloudflare_workers_script.map_server`
- `cloudflare_workers_route.map_tiles`
- `cloudflare_record.tiles`

[...]

PR #1280, opened 2026-06-26
fix(read-api): emit ACAO on no-Origin requests to stop edge cache poisoning (#1278)

julianken:
## Diagrams

```mermaid
flowchart TD
    subgraph warm[Cache-warm cron / uptime probe / curl]
      W["GET /api/observations?...zoom=3<br/>NO Origin header"]
    end
    subgraph browser[Real browser]
      B["GET same canonical URL<br/>Origin: https://bird-maps.com"]
    end

    W --> O[read-api Hono origin]
    B --> CF

    O -- before fix --> X["cors origin=array:<br/>findAllowOrigin('') = null<br/>NO ACAO, but Vary: Origin set"]
    O -- after fix --> Y["cors origin=fn:<br/>empty Origin returns canonical<br/>ACAO: https://bird-maps.com"]

    X --> CF[Cloudflare edge cache + tiered cache]
    Y --> CF

    CF -- before fix --> P["Vary:Origin '(absent)' slot<br/>caches header-less body;<br/>variant collapse serves it to B<br/>=> intermittent CORS error"]
    CF -- after fix --> G["warmed entry carries ACAO;<br/>even on variant collapse B gets<br/>a body with a valid ACAO => OK"]
```

## Summary

- **Why:** the national `zoom=3` CONUS `/api/observations` prefetch intermittently failed with a browser CORS error. Per the RCA on #1278, this is an **edge** problem: the cache-warm cron (`run-cache-warm.ts`), uptime probes, and `curl` fetch the warmed canonical URLs with **no `Origin` header**. Hono's `cors` resolved that empty Origin to `null`, emitting no `Access-Control-Allow-Origin` while still setting `Vary: Origin`. Cloudflare cached that header-less body under the `Vary: Origin` "(absent)" slot, and tiered-cache variant collapse served it to a real browser.
- **What:** resolve the CORS `origin` via a function so a request with no Origin header still yields ACAO for the canonical primary origin (`origins[0]`, i.e. `https://bird-maps.com`). The warm/probe-seeded cache entry is no longer header-less. Allow-listed browser Origins are still echoed exactly; a present-but-disallowed Origin still gets no ACAO. Bodies are Origin-agnostic, so the canonical ACAO on the no-Origin path is correct, not a leak.
- **Test gap closed:** adds the `app.test.ts` cases the RCA called for — ACAO on the no-Origin 200 path, on a no-Origin 503 (`app.onError`) and an allow-listed 503, and on a 429 rate-limit response — locking in that CORS runs before the rate-limit/error handlers so a future refactor can't silently drop the header.

**Follow-up (not in this PR):** the complementary Cloudflare-infra fix — set ACAO at the edge and/or drop `Vary: Origin` on `/api/*` (a `http_response_headers_transform` rule sibling to `cache-rule.tf`) — would structurally eliminate the per-Origin cache namespace and the no-Origin-variant poisoning. That is deferred because `terraform apply` is IAM-blocked (#298). This origin-side fix needs no infra change and lands now.

## Screenshots

N/A — not UI

## Test plan

- [x] `npm run test --workspace @bird-watch/read-api` — green (186 passed, incl. 5 new #1278 CORS cases)
- [x] New unit / integration tests added (5 cases: no-Origin 200, no-Origin 503, allow-listed 503, 429, disallowed-origin still blocked)
- [x] Verified load-bearing: reverting `app.ts` fails the two no-Origin cases (200 + 503); the fix is what makes them pass
- [x] `npm run build --workspace @bird-watch/read-api` — clean
- [x] `npm run lint` — clean across all workspaces
- [x] N/A — Playwright MCP smoke (not a UI change; no `frontend/**` files touched)

## Plan reference

Out of plan — bug fix for #1278 (national `zoom=3` prefetch CORS-blocked by no-Origin cron-warm edge cache poisoning).

Fixes #1278

---

### DIFF
--- a/services/read-api/src/app.test.ts
+++ b/services/read-api/src/app.test.ts
@@ -1557,4 +1557,101 @@ describe('CORS middleware', () => {
     expect(res.headers.get('cache-control'))
       .toBe('public, max-age=604800');
   });
+
+  // #1278 — no-Origin cache poisoning. The cache-warm cron, uptime probes, and
+  // `curl` fetch the canonical `/api/observations` URLs with NO Origin header.
+  // Before the fix, Hono's cors set no ACAO on that path, so Cloudflare cached
+  // a header-less body under the `Vary: Origin` "(absent)" slot and (with
+  // tiered-cache variant collapse) served it to real browsers → intermittent
+  // CORS error on the national prefetch. The fix emits ACAO for the canonical
+  // origin whenever the request has no Origin header, so the warm-seeded cache
+  // entry is never header-less.
+  it('sets Access-Control-Allow-Origin to the canonical origin on a no-Origin request (#1278)', async () => {
+    delete process.env.FRONTEND_ORIGINS; // canonical = https://bird-maps.com
+    const app = createApp({ pool: db.pool });
+    // No Origin header — exactly what run-cache-warm.ts and probes send.
+    const res = await app.request('/api/hotspots');
+    expect(res.status).toBe(200);
+    expect(res.headers.get('access-control-allow-origin'))
+      .toBe('https://bird-maps.com');
+  });
+
+  it('keeps Access-Control-Allow-Origin on a no-Origin 503 from app.onError (#1278)', async () => {
+    delete process.env.FRONTEND_ORIGINS;
+    // A pool whose query rejects with a connection error → app.onError → 503.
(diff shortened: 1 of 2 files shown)


julianken:
@Mergifyio queue

Issue #1242, opened 2026-06-16
feat(frontend): restore camera from #map= hash on load + write-back (Part 2)


julianken:
## Problem

Part 2 of epic #1238: make a viewbox link **self-restoring**. The live app should read the camera back from `#map=…` on cold load and frame to it — so a teammate (or Claude) opens a link in a normal browser and sees exactly the captured view, with no replay tool needed. It must cooperate with the existing scope `fitBounds` / `maxBounds` / `flyTo` camera machinery, and the camera must survive subsequent in-app navigation.

Plan-review verdict: **MEDIUM-risk, filable directly (no spike)** — the integration seam is one well-documented effect plus one shared writer, but the acceptance criteria below are load-bearing and must be implemented exactly.

## How the camera is set today (do not regress)

For `?state=US-XX` / `?scope=us`, the cold-load camera comes entirely from the scope envelope: an uncontrolled `initialViewState` (first paint) + a load-gated `map.fitBounds(activeBounds, { maxZoom: 12, duration: 600 })` in the **camera-intent effect**, which lives in `frontend/src/components/map/hooks/use-scope-camera.ts` (`useScopeCamera` / `computeScopeBounds`, extracted out of `MapCanvas.tsx` in epic #884 · U12 / #897 — `MapCanvas.tsx` has **ZERO** `.fitBounds`/`.flyTo` calls now and only *consumes* the hook ~line 561). The effect is keyed `[mapReady, boundsKey, flyTo?.key, …]` and already branches `flyTo` over `fitBounds`. `maxBounds` is a separate reactive prop (`clampBounds`). The URL contributes only the data scope, never the camera.

## Files

- EDIT `frontend/src/App.tsx` — parse `window.location.hash` **once** (a `useState`/`useRef` initializer, NON-reactive — must never re-fire on later renders) into `initialHashCamera` via C1's `decodeViewbox`; thread it through to MapCanvas alongside `boundsKey`/`flyTo`. Gate the idle write-back on the existing `scopeMoveUntilRef` settle window + `scopeActive`.
- EDIT `frontend/src/components/MapSurface.tsx` — verbatim pass-through of the new `initialHashCamera` prop.
[...]

Issue #847, opened 2026-06-02
Birds don't populate after an in-app state switch (stale debouncedBbox de-pairs the first fetch)


julianken:
### Problem

Switching scope in-app from one state to another (via the in-card scope-control `<select>`, or the chooser) sometimes leaves the map empty: zero markers, the lede reads **"No recent sightings"**, and the family legend is blank — even though the new state has recent observations. The data only returns after the user **moves the map** (a window resize reliably rescues it) or **reloads the page** ("refresh fixes it"). The failure is **intermittent** — some switches self-recover, some don't. There is no thrown error and no console noise: the failing fetch is a clean HTTP 200 with an empty payload.

### Root cause

The App-level state vars that supply the bbox/zoom for `GET /api/observations` are **never reset when the scope changes**:

- `debouncedBbox` (`frontend/src/App.tsx:274`, seeded `DEFAULT_BBOX_CONUS = [-125,24,-66,50]`) and `debouncedZoom` (`App.tsx:279`, seeded `3`) are passed into `useBirdData` at `App.tsx:328-329`.
- Their **only** writers are inside the `onViewportChange` 250 ms debounce timeout (`App.tsx:596-597`). Nothing resets them on a scope change.

So an in-app state switch flips `scopeStateCode` (`App.tsx:319`) — a dependency of the observations effect (`use-bird-data.ts:199`) — and fires `GET /api/observations` with `{ state: NEW, bbox: PREVIOUS state's settled viewport }`. The server ANDs `stateCode` (PostGIS `ST_Intersects` against the state polygon) with `bbox` (`packages/db-client/src/observations.ts:159-165,179-184`, joined with `AND` at `:184`). Two disjoint states don't intersect → empty payload → 0 markers.

[...]

PR #1290, opened 2026-06-26
fix(frontend): seed the observations fetch from the restored hash camera so deep-linked high-zoom views load markers (#1289)

julianken:
## Diagrams

```mermaid
sequenceDiagram
    participant URL as "#map= deep-link"
    participant SC as useScopeCamera
    participant MC as MapCanvas
    participant App
    participant Data as use-bird-data

    URL->>SC: parse #map=z/lat/lng
    SC->>MC: map.jumpTo(restored camera) + set restoredHashCamera
    Note over MC,App: BEFORE (#1289 bug): the restore's settle idle lands inside<br/>App's scopeMoveUntilRef window -> onViewportChange swallows it.<br/>debouncedBbox/zoom stay at the CONUS z3 SEED -> 0 markers.
    MC->>App: onHashCameraRestored(liveBounds, floorZoom)  [NEW seam]
    App->>App: setDebouncedBbox / setDebouncedZoom (restored viewport)
    App->>Data: filters = {bbox: restored, zoom: >=6}
    Data->>Data: GET /api/observations at the restored view -> markers render
```

## Summary

- **Why:** deep-linking / sharing a `#map=<z>/<lat>/<lng>` link to a high-zoom view (zoom >= ~6, every scope) rendered **zero markers** — the camera restored but the `/api/observations` fetch stayed pinned to the CONUS z3 seed. Regression in the just-shipped viewbox/share-link epic (#1238 / #1242).
- **Root cause:** the camera-restore and the data-fetch never connected on a deep-link. The restore's `jumpTo` settle `idle` reaches `onViewportChange` *inside* App's `scopeMoveUntilRef` window (the #847 scope-`fitBounds` over-fetch guard), where the bbox/zoom commit is deliberately swallowed. On a clean deep-link no later user gesture fires, so the seed is never superseded.
- **Fix (lowest-risk, additive-only):** `MapCanvas` fires a new `onHashCameraRestored(liveBounds, liveZoom)` callback **once** when `restoredHashCamera` goes null→non-null; App seeds `debouncedBbox`/`debouncedZoom` directly from it — exactly the `{bbox, zoom}` pair `onViewportChange` would have committed. Does **not** touch the settle window, the 250ms debounce, the ~1e-4° epsilon guard, the canonical-key bbox machinery, or the #847 reseed. One-shot per restore (`firedRestoreSeedRef`); a genuine later pan still flows through the normal path. A redundant CONUS z3 seed fetch firing first is acceptable.

## Screenshots

Pending — orchestrator will capture before/after at the repro deep-links via MCP (subagents cannot drive MCP).

      (or marked `N/A — class is consumed only by a primitive that ships its own CSS`) — N/A — no className / CSS changes (behavior-only; the change makes markers appear that were missing).
- [ ] Multi-viewport design QA: PR body has ≥10 `user-attachments` URLs
      (5 viewports × 2 themes per #445). Verify:
      `gh pr view <N> --repo julianken/bird-sight-system --json body --jq '.body | [scan("user-attachments/assets/[a-f0-9-]++")] | length'`
      returns ≥ 10 (or marked `N/A — not UI`)

## Test plan

- [x] `npm run test -w @bird-watch/frontend` — green (1791 passed). Includes two new App-level units (`#1289`): the restored-viewport fetch fires at zoom>=6 with the restored bbox (not `{CONUS, 3}`), and the restore seed is one-shot (no double-fetch loop). Both confirmed RED on pre-fix code, GREEN after.
- [x] New unit / integration tests added (behavior changed) — `frontend/src/App.test.tsx`.
- [x] New Playwright e2e spec added (user-visible behavior changed) — `frontend/e2e/scope/viewbox-restore.spec.ts`: asserts the `/api/observations` **fetch** fired at zoom>=6 with a small bbox enclosing the restored center (the gap the existing `data-hash-camera`-only specs missed), plus a WebGL-guarded marker-presence check. Ran green locally against an isolated dev stack.
- [x] `npm run build -w @bird-watch/frontend` — clean production build.
- [ ] (UI only) Playwright MCP smoke — orchestrator to run the before/after live drive at the repro deep-links (subagents cannot drive MCP).

## Plan reference

Out of plan — HIGH-severity regression fix for #1289 (deep-linked high-zoom views render zero markers), regression in the shipped viewbox/share-link epic #1238 / #1242.

---

### DIFF
--- a/frontend/e2e/scope/viewbox-restore.spec.ts
+++ b/frontend/e2e/scope/viewbox-restore.spec.ts
@@ -164,4 +164,118 @@ test.describe('viewbox-restore on cold load (#1242)', () => {
     // No hash ⇒ no restore handle.
     await expect(app.mapCanvas).not.toHaveAttribute('data-hash-camera', /.*/);
   });
+
+  // #1289 — REGRESSION GUARD. The original viewbox-restore tests above assert
+  // ONLY the camera (`data-hash-camera`) + hash survival — never that the
+  // observations FETCH moved off the CONUS z3 seed to the restored viewport.
+  // That blind spot let the deep-link "zero markers at high zoom" bug ship: the
+  // camera restored but the fetch stayed pinned to the z3 seed (its settle idle
+  // swallowed by App's scope-move window), so the restored high-zoom rectangle —
+  // which shares no rows/cells with the z3 national aggregate — rendered empty.
+  //
+  // This test asserts the FETCH directly: a high-zoom `#map=` restore must fire
+  // an `/api/observations` request at zoom >= 6 with a bbox ENCLOSING the
+  // restored center. A marker-presence check follows, WebGL-guarded with
+  // `test.skip` (per the repo convention — full marker render needs a GPU the
+  // headless CI runner may lack; the fetch assertion is the GPU-free proof).
+  test('#1289: a high-zoom #map= restore fires the observations fetch at the restored viewport (not the z3 seed)', async ({
+    page,
+    apiStub,
+  }) => {
+    // A tiny high-zoom (z16) rectangle around Tucson, AZ — INSIDE the AZ scope
+    // envelope, so the restore validates + applies. The AZ_OBS fixture point
+    // (-110.974, 32.221) sits inside this rectangle, so the restored-viewport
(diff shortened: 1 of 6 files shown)


julianken:
@Mergifyio queue

